PR #21, opened 2016-07-22
Change some mercurial docs to git docs in the faq

kragniz:
This is an initial draft of moving the faq to git commands.

### DIFF
--- a/faq.rst
+++ b/faq.rst
@@ -137,193 +137,112 @@ For everyone
 
 The following FAQs are intended for both core developers and contributors.
 
-Where can I learn about the version control system used, Mercurial (hg)?
+Where can I learn about the version control system used, Git?
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
-Mercurial_'s (also known as ``hg``) official web site is at
-https://www.mercurial-scm.org/.  A book on Mercurial published by
-`O'Reilly Media`_, `Mercurial: The Definitive Guide`_, is available
-for free online.  Another resource is `Hg Init: a Mercurial tutorial`_
-by Joel Spolsky.
+Git_'s official web site is at https://git-scm.com/.  A book on Git published
+by Apress, `Pro Git`_, is available for free online.
 
-With Mercurial installed, you can run the help tool that comes with
-Mercurial to get help::
+With Git installed, you can run the help tool that comes with
+Git to get help::
 
-  hg help
+  git help
 
-The `man page`_ for ``hg`` provides a quick refresher on the details of
-various commands, but doesn't provide any guidance on overall
-workflow.
+The `man page`_ for ``git`` provides a quick refresher on the various commands,
+but doesn't provide any guidance on overall workflow. Each sub-command has a
+more detailed manpage. You can view these by running a command such as::
 
-.. _Mercurial: https://www.mercurial-scm.org/
-.. _O'Reilly Media: http://www.oreilly.com/
-.. _Mercurial\: The Definitive Guide: http://hgbook.red-bean.com/
-.. _man page: http://www.selenic.com/mercurial/hg.1.html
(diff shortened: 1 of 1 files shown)


faq.rst
```diff
@@ -137,193 +137,112 @@ For everyone
...
 
 Use::
 
-   $ hg branch
-   default
+   $ git branch
+   * master
 
-Or to get more information::
+The asterisk (\*)next to the branch name denotes the current branch.
```
berkerpeksag: `(\*)next` -> `(\*) next`

faq.rst
```diff
@@ -334,63 +253,37 @@ are removed.
...
-In your global ``.hgrc`` file add a section similar to the following::
-
-  [paths]
-  dg = ssh://hg@hg.python.org/devguide
-
-This example creates a ``dg`` alias for the ``devguide`` repository
-on ``hg.python.org``. This allows "dg" to be entered instead of the
-full URL for commands taking a repository argument (e.g. ``hg pull dg`` or
-``hg outgoing dg``).
+In ``.git/config`` file for the local repository, add
```
berkerpeksag: Can't we use `git branch --set-upstream-to` instead of editing `.git/config`?

faq.rst
```diff
@@ -399,62 +292,38 @@ How do I compare my local repository to a remote repository?
...
 
 How do I update my local repository to be in sync with a remote repository?
 '''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
 Run::
 
-   hg pull <remote repository>
+    $ git pull --ff-only <remote> <branch name>
+
+eg::
```
berkerpeksag: "Example" (or something similar) might be a better choice for non-native speakers.

berkerpeksag:
This looks great to me, thanks! :)

faq.rst
kragniz: Thanks for the feedback! Is what I just pushed what you were thinking? I will admit to fiddling with the git config file more than I should do...

berkerpeksag: LGTM. We probably need a separate FAQ entry for making `--ff-only` the default behavior but that can be done in a separeate pull request.

I will wait for @brettcannon's review and merge this weekend. Thanks again!

faq.rst
```diff
@@ -399,62 +290,38 @@ How do I compare my local repository to a remote repository?
 To display the list of changes that are in your local repository, but not
 in the remote, use::
```
vadmium: Maybe clarify that this only lists commits that were copied from the remote the last time the local copy was updated (via “git pull” or “git fetch”)

faq.rst
```diff
@@ -399,62 +290,38 @@ How do I compare my local repository to a remote repository?
...
- hg outgoing <remote repository>
+    $ git log origin/master..
 
 This is the list of changes that will be sent if you call
-``hg push <remote repository>``.  It does **not** include any :ref:`uncommitted
+``git push``.  It does **not** include any :ref:`uncommitted
 changes <hg-status>` in your working copy!
 
 Conversely, for the list of changes that are in the remote repository but
 not in the local, use::
```
vadmium: [Sorry, my previous comment was at the wrong place]

This only lists commits that were copied from the remote the last time the local copy was updated (via “git pull” or “git fetch”). I don’t think it will list new commits that will be fetched.

kragniz: Yeah, I guess say to run `git remote update` beforehand?

vadmium: Maybe. I’ve only used “git fetch”, but it seems to be similar.

kragniz: I've added a little note about this below.

brettcannon:
Looks like there's come conflicts. Could you update and fix them?

kragniz:
Sure, updated.

brettcannon:
Thanks!

PR #34, opened 2016-09-09
Move the search box from sidebar to header and footer.

ammaraskar:
Mirrors this change from cpython docs:

https://hg.python.org/cpython/rev/274b25cd501f
https://bugs.python.org/issue19489


zware:
Thanks @ammaraskar!

Issue #36, opened 2016-09-14
Dev Guide Get Started Guide 


Mariatta:
I found that the Get Started document here https://docs.python.org/devguide/ to be not very effective way for me to get started.

The page started by saying, "here are the basic steps":
1.  check out source code
2. compile
3. run tests
4. make patch
5. submit to issue tracker

I agree that these five steps correctly describe the full cycle of contributing to the code base.
**However**, to really get started for the *first time ever*, it's not gonna work. In realty, there is a step 0, and that is to Get Setup ( https://docs.python.org/devguide/setup.html#setup)

Can I suggest that Get Setup be clearly explicitly described as step 0 or step 1 of contributing? At the moment it is only subtly mentioned under the Get Started heading as a link. At first I missed this link completely, and it caused some frustration on my part.

As a first timer, I tried jumping on doing step 1 right away, (checkout out source code), but then I realized I don't have mercurial installed. I had to do some googling on how to install mercurial. Finally managed to check out the source code using the `hg clone`. Only much later that I came across the Get Setup page.

After I installed mercurial and checked out source code, I read the instruction on step 2, to compile python.
Foolishly, I tried to execute the command `/configure --with-pydebug && make -j2` 
Well that didn't work at all. I almost gave up with contributing because of failing to complete step 2. 
Again, I had to go back and read the Get Setup page more thorough explanation on how to compile, and it turns out for my machine (mac os) I needed a different command, whereas the get started guide says I could use that command above.

I discussed this with @gvanrossum earlier today, and he suggested that I file a bug or an issue about this matter.
Guido also pointed out that the --with-pydebug flag is for advanced debugger. and preferably it is separated into two lines

Getting local environment is the first step before anyone can begin contributing. It would really be great if this page could be improved and updated to be more beginner friendly. 

Thanks, and sorry if this is a long ticket. Let me know if anyone has further thoughts on how to improve this page. I'd be happy to help out however I can.

gvanrossum:
Thanks for persevering Mariatta! If you want to you can submit a PR with changes to the devguide that would have helped you. Any one of us will then be able to review and merge that change! (Hopefully getting started with the devguide is simpler than getting started with Python. :-)

willingc:
@mariatta Feel free to ask question about Sphinx or how to build the devguide here if needed too :-)

Mariatta:
Thanks, I made pull request for this. Would appreciate any feedback 😃

ncoghlan:
[...]

PR #37, opened 2016-09-21
Made "install dependencies" as Step number one.

Mariatta:
Attempt to address issue #36 

I'm open for suggestions to further improve this. Thanks.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -14,19 +14,36 @@ Quick Start
 Here are the basic steps needed to get :ref:`set up <setup>` and contribute a
 patch:
 
-1. :ref:`Get the source code <checkout>`::
+1. Set up and install dependencies.
+
+   Install :ref:`Mercurial <vcsetup>` and other dependencies.
+
+   The dependencies needed will depend on the platform you're on.
+   Go to :ref:`Get Setup <setup>` page for detailed information.
+
+2. :ref:`Get the source code <checkout>`::
 
       hg clone https://hg.python.org/cpython
 
-   See :ref:`version control setup <vcsetup>` for installing Mercurial.
 
-2. :ref:`Build Python <compiling>`.  On all platforms, install build
-   dependencies (such as compilers). On :ref:`UNIX <unix-compiling>`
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username at [bugs.python.org](http://bugs.python.org/) (b.p.o). If you don't already have an account at b.p.o, please [create one](http://bugs.python.org/user?@template=register) and make sure to add your GitHub username. If you do already have an account at b.p.o then please go there and under "Your Details" add your GitHub username.

And in case you haven't already, please make sure to sign the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA); we can't legally look at your contribution until you have signed the CLA.

Once you have done everything that's needed, please reply here and someone will verify everything is in order.

Mariatta:
Thanks. I updated my profile at bugs.python.org

This is my profile there.
http://bugs.python.org/user24617


Mariatta Wijaya

ncoghlan:
Heh, I must have just seen the CLA signature before the bot did :)

This looks good to me, so merging - thanks!

index.rst
```diff
@@ -14,36 +14,53 @@ Quick Start
...
+
+   The command to compile on UNIX is::
+
+      ./configure
+      make -j2
+
+   On some Mac OS:::
 
-      ./configure --with-pydebug && make -j2
+      CPPFLAGS="-I$(brew --prefix openssl)/include" \
```
berkerpeksag: This is already covered for both Homebrew and MacPorts in `setup.rst`. I don't think we need to duplicate this information here.

willingc: @Mariatta Thanks for the updates! Perhaps it would be better to mention that for OS X that the developer should refer to the section in `setup.rst` to set CPPFLAGS and LDFLAGS. This way, `setup.rst` can have details whether using brew or macports, but the user is notified that an additional step may be needed. As written and pointed out be @berkerpeksag, a user with macports would get an error if setting to these brew parameters. What do you think?

PR #57, opened 2016-10-07
Update cherry-picking for git flow

soltysh:
Partially addresses #44. 

@brettcannon ptal

PR #58, opened 2016-10-07
Update cherry-picking for git flow

soltysh:
Partially addresses #44. 

Replacement for #57, which was opened against wrong branch.

@brettcannon ptal

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -469,21 +469,21 @@ to master branch::
 
    git checkout master
    git apply --reject patch.diff
-   # Fix any conflicts (e.g. look for *.rej files); compile; run the test suite
+   # Fix any conflicts (e.g. look for *.rej files); compile; run the test suite.
    git add -A
    git commit -m '#12345: fix some issue.'
-   # Note the commit SHA (e.g. git log or git rev-parse --short HEAD)
+   # Note the commit SHA (e.g. git log or git rev-parse --short HEAD).
 
 Then you can switch to the ``3.6`` branch (or appropriate older version), cherry-pick
 the commit and run the test::
 
    git checkout 3.6
    # Instead of 3.6 use an appropriate branch reflecting the Python version you
-   # are backporting your change to
+   # are backporting your change to.
    git cherry-pick -x 123abc
-   # where 123abc is the SHA of the previous commit
-   # Fix any conflicts (add changes with git add -A and git cherry-pick --continue)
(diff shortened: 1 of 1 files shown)


brettcannon:
Generally looks good, but since we're updating the branch commit order (see inline comment) could you update that as well (basically commit to master first, then 3.6)?

committing.rst
```diff
@@ -460,28 +460,28 @@ Python version.
...
-   cd 3.6
-   hg import --no-commit patch.diff
-   # Compile; run the test suite
-   hg ci -m '#12345: fix some issue.'
+   git checkout 3.6
+   git apply --reject patch.diff
+   # Fix any conflicts (e.g. look for *.rej files); compile; run the test suite
+   git add -A
+   git commit -m '#12345: fix some issue.'
+   # note the commit SHA (e.g. git log)
```
berkerpeksag: We could also mention `git rev-parse --short HEAD` (or something else that does the same job) as an alternative here.

committing.rst
```diff
@@ -460,28 +460,28 @@ Python version.
...
 
-Then you can switch to the ``3.7`` clone, merge, run the tests and commit::
+Then you can switch to the ``3.7`` branch, cherry-pick the commit and run the test::
```
berkerpeksag: There won't be a 3.7 branch until it reaches to beta 1 in 2018. I think this will be replaced with 3.6 once you addressed Brett's comment above.

soltysh:
@brettcannon @berkerpeksag comments addressed. I've wrote it so that patch has to lands against master first, and then backport to appropriate older version (example is using 3.6). Do we have a list of branches in the devguide or at least a clear information that branch 2.7 is the latest 2.7 version? IMHO it might be valuable.

berkerpeksag:
> Do we have a list of branches in the devguide or at least a clear information that branch 2.7 is the latest 2.7 version? IMHO it might be valuable.

Yes, there's a table in the front page: https://docs.python.org/devguide/index.html#status-of-python-branches

Updated PR looks good to me, but it would be good to use a consistent style in comments (e.g. start every line with either an uppercase or a lowercase letter)

brettcannon:
Thanks to the new GitHub feature letting repo contributors edit PRs, I just tweaked the capitalization myself as well as ended all sentences with a period. Thanks, @soltysh !

soltysh:
Thank you!

PR #62, opened 2016-10-15
Updating committing.rst to reflect new git workflow

soltysh:
@brettcannon as promised I've started updating the `committing.rst` file to reflect git workflow. There are a few questions that popped-up while going through the doc. It's still WIP.

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -12,7 +12,8 @@ ensure your changes fulfill several mandatory criteria.
 
 .. note::
    If you want to share your work-in-progress code on a feature or bugfix,
-   either publish patches or create a public fork of the repository.
+   either open a WIP-prefixed PR, publish patches on the `issue tracker`_ or
+   create a public fork of the repository.
 
 
 Does the test suite still pass?
@@ -32,6 +33,7 @@ passes before pushing any code changes.
    build options).  But it will at least catch non-build specific,
    non-platform specific errors, therefore minimizing the chance for
    breakage.
+
 
 Patch checklist
 '''''''''''''''
@@ -115,7 +117,7 @@ original checkin message in that case).
 
 Finally, especially for larger patches, check if the submitter of the
 patch has a CLA in place (indicated by an asterisk following their username
-in the issue tracker). If the asterisk is missing and the patch is
+in the `issue tracker`_). If the asterisk is missing and the patch is
 non-trivial, direct them to the electronic `Contributor Licensing Agreement`_
 to ensure the PSF has the appropriate authorizations in place to relicense
 and redistribute their code.
@@ -224,11 +226,11 @@ a new paragraph(s) can be added to explain in proper depth what has happened
 understands the justification for the change).  Also, if a non-core developer
 contributed to the resolution, it is good practice to credit them.
 
-Mercurial hooks
+Hooks
 '''''''''''''''
 
-Special hooks have been added to the Mercurial repository to enable notifying
-the issue tracker of a commit related to an issue.
+Special hooks have been added to the repository to enable notifying the `issue tracker`_
+of a commit related to an issue.
 
 A commit message can mention one or several issues in one of the following
 ways::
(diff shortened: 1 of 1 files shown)


committing.rst
```diff
@@ -246,81 +247,70 @@ issue's page in the tracker, for each mentioned issue.
...
 Minimal Configuration
 ---------------------
 
-If you use Mercurial as a committer of patches (your own or others), you should
+If you use Git as a committer of patches (your own or others), you should
 set up some basic options in your `configuration file`_.
 
-* On *UNIX* (including Mac OS X), you must edit ``~/.hgrc``.
+* On *UNIX* (including Mac OS X), you must edit ``~/.gitconfig``.
+// TODO find out how windows handles git
```
soltysh: I need to find a windows machine and verify that.

committing.rst
```diff
@@ -246,81 +247,70 @@ issue's page in the tracker, for each mentioned issue.
...
-* *Under Windows*, you should also enable the `eol extension`_, which will
+* *Under Windows*, you should also enable the *autocrlf* option, which will
   fix any Windows-specific line endings your text editor might insert when you
   create or modify versioned files.  The public repository has a hook which
   will reject all changesets having the wrong line endings, so enabling this
   extension on your local computer is in your best interest.
 
+   [core]
+   autocrlf = true
+
```
soltysh: Are there any other settings that come to your mind or should I ask at core-workflow list?

committing.rst
```diff
@@ -246,81 +247,70 @@ issue's page in the tracker, for each mentioned issue.
...
 As a core developer, it can be very helpful to set up the same commit checks
 locally that the main repo enforces for incoming patch sets. This can save a
-great deal of frustration with the server rejecting ``hg push`` for changes
+great deal of frustration with the server rejecting ``git push`` for changes
 (which can be especially time consuming if you have already merged changes
 across branches).
 
+// TODO is that still true?
```
soltysh: How this is/will be handled wrt to github? I do believe setting (well obviously writing those) checks won't hurt, but will that be checked on the server?

committing.rst
```diff
@@ -347,6 +337,8 @@ will still be issued but the changeset will be applied anyway::
 Clones Setup
 ------------
 
+// TODO is that still needed?
```
soltysh: Do we still recommend having multiple clones? It is worth talking about forks and working with them, so that changes land in people's private forks rather than in the main repo. But given the fact that git branches are so easy do we recommend having multiple dirs, one per each branch. I do believe some of core-devs prefer that approach, but is it worth having it in here, still?

committing.rst
```diff
@@ -246,81 +247,70 @@ issue's page in the tracker, for each mentioned issue.
...
-* You should not push new named branches to the main repository.  You can
-  still use them in clones that you use for development of patches; you can
+* You should not push new branches to the main repository.  You can still use
+  them in your fork that you use for development of patches; you can
   also push these branches to a **separate** public repository that will be
   dedicated to maintenance of the work before the work gets integrated in the
   main repository.
 
-* You should collapse changesets of a single feature or bugfix before pushing
+* You should `squash commits`_ of a single feature or bugfix before pushing
```
soltysh: I'll add a new section about how to do it down the doc.

soltysh:
Forgot to link with #44

committing.rst
methane: In hg, only `PCbuild/readme.txt` is CR LF. All other files (including vcxproj) is LF.
So safe default config may be `autocrlf = input`.

committing.rst
brettcannon: I don't think it's worth keeping here. I haven't looked at the PR yet (started to read the comments first), but I think as long as we explain how to set up a different remote (e.g. `origin` and `upstream` for your fork and python/cpython, respectively) then that should cover it.

committing.rst
brettcannon: Don't know yet. We will (hopefully) have some CI set up quickly, but after that I don't know when we will get around to adding checks for whitespace, etc.

soltysh: I'll remove it for now, we'll revert what's needed from history once we have the necessary bits in place, in that case.

soltysh:
I've addressed all comments and hopefully finished changing this file, ptal.

brettcannon:
LGTM. When you're ready for this PR to be committed, @soltysh , just remove the `WIP` from the title.

soltysh:
Of course I forgot to do just that, updated. Further changes will be done as a followups :) Thank you!

brettcannon:
Thanks!

PR #71, opened 2016-10-28
Drop all hg-related questions from the FAQ.

### LINKED ISSUE #48: Update FAQ.rst for Git/GitHub
Only necessary if #27 doesn't happen.

brettcannon:
Questions already moved were left. Otherwise Git is covered well enough online
that we should avoid duplicating documenting anything that isn't specific to
our suggested workflow.

Closes #48


faq.rst
```diff
@@ -181,8 +183,8 @@ best to run a ``make distclean`` to ensure that all previously built files
 are removed.
 
 
-I want to keep a separate working copy per development branch, is it possible?
-''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+I want to keep a separate working copy per development branch, is it's possible?
```
Mariatta: is `it` possible?
instead of is `it's` possible

brettcannon: Heh, my brain read it as "_if_ its possible", hence the word change.

zware: We might also suggest `git worktree` here, I've used it some and it works pretty well.  Equivalent to `hg share`.

brettcannon: @Mariatta I fixed the typo. Thanks!
@zware If you want to add an entry for it then go for it!

Mariatta: Personally I'd prefer a more explicit form of question:

`Can I keep a separate working copy per development branch?`
`How do I keep a separate working copy per development branch?`

PR #88, opened 2016-12-24
Test results are not saved anywhere.

iselind:
Clarifying that the test results are not saved anywhere.

### DIFF
--- a/runtests.rst
+++ b/runtests.rst
@@ -59,6 +59,8 @@ using several Python processes so as to speed up things::
 If you are running a version of Python prior to 3.3 you must specify the number
 of processes to run simultaneously (e.g. ``-j2``).
 
+Keep in mind that the test results are not stored anywhere. If you want to keep the results for later reference, you'll have to redirect the output from the tests on your own.
+
 .. _strenuous_testing:
 
 Finally, if you want to run tests under a more strenuous set of settings, you
@@ -106,8 +108,8 @@ out and resolve the cause of the problem (for example, the default security
 settings on some platforms will disallow some tests)
 
 
-Writing
--------
+Writing Tests
+-------------
 
 Writing tests for Python is much like writing tests for your own code. Tests
 need to be thorough, fast, isolated, consistently repeatable, and as simple as


the-knights-who-say-ni:

Unfortunately our records indicate you have not signed a [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA). For legal reasons we need you to sign this before we can look at your contribution.

Once you have done everything that's needed, please reply here and someone will verify everything is in order.

iselind:
Linked to http://bugs.python.org/issue29047 but i cannot figure out how to add the pull request to the issue tracker correctly

iselind:
I have now signed the document.

Regards,

Patrik Iselind

berkerpeksag:
Thanks for the PR, but -1 from me. We can't and shouldn't add every small detail to the devguide. The all supported features of the test runner can already be seen by using the `-h` option and there is no need to document every unsupported feature.

zhangyangyu:
Agree with Berker. I don't understand why you emphasize the output. And it's obvious that you can redirect the output to store it.

iselind:
I for one expected the result to be stored somewhere.

Patrik

Den 24 dec 2016 13:46 skrev "Xiang Zhang" <notifications@github.com>:

iselind:
Other new developers might be helped by this as well.

Patrik

Den 24 dec 2016 16:41 skrev "mrx" <patrik.mrx@gmail.com>:

iselind:
@berkerpeksag , i'm not saying everything should be repeated. Just this one.

ammaraskar:
It's still a fairly weird thing to mention since most, if not all unit testing programs I've encountered simply print out their results to stdout and set the appropriate error code when they return. I guess it helps if you've never encountered unit testing in other projects before.

Some examples:
* tox
* unittest module
* rake (for ruby)
* mocha (for js)
* CUnit

brettcannon:
Thanks for the PR, @iselind , but the consensus is that this isn't necessary, so I'm closing this.

iselind:
I don't know if it's just me but i think that dev docs should cover enough
so that as many as possible can start to help out and not assume a certain
level of experience with this or that.

Patrik Iselind

brettcannon:
We're not trying to make Python contribution work for any person in the world who may want to help out (e.g. we are not here to teach people how to program). There is a certain level of prior knowledge necessary and we are saying that being able to read the documentation of `test` to know it doesn't save test results or to simply realize it doesn't like nearly all other test runners is a prerequisite to helping out with code.

PR #95, opened 2017-01-14
Clarify core developer responsibilities

ncoghlan:
We don't clearly document anywhere the additional expectations
that apply to core developers merging changes that don't apply
to folks contributing in other ways.

This new section is an initial step in addressing that omission.

### DIFF
--- a/coredev.rst
+++ b/coredev.rst
@@ -18,6 +18,44 @@ a while to make sure you understand the development process. If other core
 developers agree that you should gain commit privileges you are then extended
 an official offer.
 
+What it Means
+-------------
+
+As contributors to the CPython project, our shared responsibility is to
+collaborate constructively with other contributors, including core developers.
+This responsibility covers all forms of contribution, whether that's submitting
+patches to the implementation or documentation, reviewing other peoples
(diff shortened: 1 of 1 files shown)


coredev.rst
```diff
@@ -18,6 +18,44 @@ a while to make sure you understand the development process. If other core
 developers agree that you should gain commit privileges you are then extended
 an official offer.
 
+What it Means
+-------------
+
+As contributors to the CPython project, our shared responsibility is to
+collaborate constructively with other contributors, including core developers.
+This responsibility covers all forms of contribution, whether that's submitting
+patches to the implementation or documentation, reviewing other peoples
```
brettcannon: Is there a missing apostrophe for "peoples", e.g. "peoples'"?

brettcannon:
Only one grammar question, otherwise LGTM.

coredev.rst
willingc: Yep. Good catch @brettcannon. Should be `peoples'`

ncoghlan:
Thanks for the reviews folks - missing apostrophe has been added and the change pushed live :)

PR #123, opened 2017-02-18
include reference to create a pull request for the tests

DimitrisJim:
reword sentence to state how a pull request is required (and fix <patch> ref)

### DIFF
--- a/coverage.rst
+++ b/coverage.rst
@@ -214,8 +214,8 @@ times.
 
 Filing the Issue
 """"""""""""""""
-Once you have increased coverage, you need to
-:ref:`generate the patch <patch>` and submit it to the `issue tracker`_. On the
+Once you have increased coverage, you need to create an issue on the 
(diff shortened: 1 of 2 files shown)


DimitrisJim:
A very similar situation also exists in section [6.1. Helping with issues filed on the issue tracker](https://docs.python.org/devguide/docquality.html#helping-with-issues-filed-on-the-issue-tracker) while section [6.2. Proofreading](https://docs.python.org/devguide/docquality.html#proofreading

berkerpeksag:
> A very similar situation also exists in section 6.1. Helping with issues filed on the issue tracker, I'll add a PR for that too if needed.

Good catch, thanks! Could you add another commit to this one instead?

DimitrisJim:
Added the commit. Also took the liberty of mentioning how simple typos don't actually need an issue created for them.

berkerpeksag:
Thanks!

PR #135, opened 2017-02-25
Update last/latest versions: 3.2.6, 3.3.6, 3.4.6.

andreif:


andreif:
@berkerpeksag Ok, I just noticed inconsistency between versions in that respect. I can make two lines -- one for binary, one for security, so that's more clear. Should I do it?

andreif:
@berkerpeksag Updated according to your suggestions.

ned-deily:
@birkenfeld Any opinions about the 3.2 status comment above?

andreif:
@ned-deily I am not sure I understand why the table does not tell about last/latest security releases following the last binary release. I find it confusing. In addition to an exercise of mind to find the latest release of 3.N.x (apart of 3.6.x) on the main download page, I am not sure why it's made to be so difficult to see what the last version is. It's easier to check the Wikipedia page to get this info. But why?

andreif:
Oh, I was wrong -- there are no minor versions on the Wikipedia page.

ned-deily:
As the section header states, the point of this table is not to give information about releases but rather the status of the source tree branches to guide developers in choosing what branches to backport fixes to.  The primary source of information about releases is https://www.python.org/downloads/.   The last binary release info for security-only branches is sort of shorthand for "Last bugfix release before entering security-only status".   But the release information is redundant (and incomplete).  Perhaps the best solution is to remove all of the last release comments.  Also the branches listed reflected the status of the previous hg-based cpython repo.  With the move to GitHub, the older branches are being removed anyway, replaced by tags.  I think we should remove the info for the oldest branches here, that is, 2.6, 3.0, 3.1, and possibly 3.2 (depending on what @birkenfeld says), change the comments to  "Versions older than 2.7 and 3.2 have reached end-of-life.", and also change "default" to "master" for the feature branch name.

andreif:
@ned-deily Alright, thanks for the explanation! It will be more understandable this way.

ned-deily:
@andreif Thank you for bringing up the issue, it helped to clarify my thoughts on this section.  It would be great if you feel like updating your PR otherwise one of us will get around to updating this.

andreif:
@ned-deily Yes, sure, happy to help. I just thought we are waiting for some input from @birkenfeld ?

ezio-melotti:
To avoid similar discussions in the future, it might be better to add a sentence (or perhaps an rst comment) before the table to indicate its purpose.  A link pointing to https://www.python.org/downloads/ for whose who are looking for specific releases could also be added.

willingc:
Hi @andreif,

Thanks for your work on this PR. Do you want to make the updates suggested by @ned-deily or would you prefer someone else finish this PR? Just let us know. :sunny:

andreif:
Yeah, I am overloaded atm, so won't be able to do anything.

brettcannon:
If we provide a link to python.org/downloads as @ezio-melotti suggests then maybe we can drop the specific links all-together to avoid having to update them after a release? The less things to have to keep track of when cutting a release the better, right, @ned-deily 😉 .

matrixise:
With the last modifications on the `index.rst` file, is this PR needed ?

ned-deily:
Yes, I think we inadvertently bypassed this PR by a more recent update.  And, rather than add more info to the overloaded page of the devguide, I would like to provide a more comprehensive release summary page on the main website.  So I'm going to close this now.  Thanks everyone for your contributions!

PR #143, opened 2017-03-10
Remove Guido as an asyncio expert

brettcannon:
Request was made by email.


zware:
Also removed from auto-nosy on the tracker.

brettcannon:
@zware thanks! How does one do that on the tracker?

zware:
@brettcannon https://bugs.python.org/component to look up the component ID (or you can just edit it on that page), then https://bugs.python.org/component30 is asyncio.  Must have Coordinator role or greater to change it.

gvanrossum:
Thanks!

Issue #125, opened 2017-02-19
Add core-dev guidance on squash merge commit messages


zware:
See python/core-workflow#27: we should amend the devguide to specify how squash merge messages should be amended from GitHub's default message for a PR.

PR #144, opened 2017-03-13
Add Git Bootcamp page

Mariatta:
The no frills guide to git.

Closes https://github.com/python/devguide/issues/125, https://github.com/python/core-workflow/issues/27

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -20,7 +20,7 @@ You'll only need to do this once.
 
 3. When asked where to fork the repository, choose to fork it to your username.
 
-4. A fork will be created at https://github.com/UserName/cpython.
+4. A fork will be created at https://github.com/<username>/cpython.
 
 
 Cloning The Forked CPython Repository
@@ -28,7 +28,7 @@ Cloning The Forked CPython Repository
 
 From your command line::
 
-   $ git clone git@github.com:UserName/cpython.git
+   $ git clone git@github.com:<username>/cpython.git
    $ cd cpython
    $ git remote add upstream git@github.com:python/cpython.git
 
@@ -45,7 +45,7 @@ Creating and Switching Branches
 -------------------------------
 
 .. note::
-   Never commit directly to the `master` branch.
+   Never commit directly to the ``master`` branch.
 
 Create a new branch::
 
@@ -53,7 +53,7 @@ Create a new branch::
 
 which is equal to::
 
-   $ git branch some-branch # create 'some-branch', without checking it out
+   $ git branch some-branch master # create 'some-branch' off 'master, without checking it out
    $ git checkout some-branch # check out 'some-branch'
 
 To find out which branch you are in now::
@@ -142,7 +142,7 @@ Creating a Pull Request
 
 3. Select the base fork: ``python/cpython`` and base branch: ``master``
 
-4. Select the head fork: ``UserName/cpython`` and base branch: the branch
+4. Select the head fork: ``<username>/cpython`` and base branch: the branch
    containing your changes.
 
 5. Press ``Create Pull Request`` button.
@@ -238,8 +238,8 @@ For example, to fetch and checkout pull request #777::
    $ git pr 777
 
 
-Accepting and Merging Pull Request
-----------------------------------
+Accepting and Merging A Pull Request
+------------------------------------
 
 Pull requests can be accepted and merged by a Python Core Developer.
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -197,7 +197,7 @@ First, obtain the commit sha1 from the merged pull request:
 
 2. Scroll down and find the activity that says something like::
 
-     SomeUser merged commit <hash> into python:master ...
+     CoreDeveloper merged commit <hash> into python:master ...
 
 3. Follow the link to <hash>.
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -1,9 +1,11 @@
+.. highlight:: console
+
 .. _gitbootcamp:
 
 Git Bootcamp and Cheat Sheet
 ============================
 
-In this section, we'll go over some commonly used git commands that are
+In this section, we'll go over some commonly used Git commands that are
 relevant to CPython's workflow.
 
 .. contents::
@@ -16,17 +18,17 @@ You'll only need to do this once.
 
 1. Go to https://github.com/python/cpython.
 
-2. Press ``Fork``.
+2. Press ``Fork`` on the top right.
 
 3. When asked where to fork the repository, choose to fork it to your username.
 
-4. A fork will be created at https://github.com/<username>/cpython.
+4. Your fork will be created at https://github.com/<username>/cpython.
 
 
 Cloning The Forked CPython Repository
 -------------------------------------
 
-From your command line::
+You'll only need to do this once.  From your command line::
 
    $ git clone git@github.com:<username>/cpython.git
    $ cd cpython
@@ -36,7 +38,7 @@ From your command line::
 Listing the Remote Repositories
 -------------------------------
 
-::
+To list the remote repositories that are configured, along with their urls::
 
    $ git remote -v
 
@@ -47,14 +49,17 @@ Creating and Switching Branches
 .. note::
    Never commit directly to the ``master`` branch.
 
-Create a new branch::
+Create a new branch and switch to it::
 
-   $ git checkout -b some-branch master # creates a new branch off master
+   # creates a new branch off master and switch to it
+   $ git checkout -b <branch-name> master
 
-which is equal to::
+This is equivalent to::
 
-   $ git branch some-branch master # create 'some-branch' off 'master', without checking it out
-   $ git checkout some-branch # check out 'some-branch'
+   # create a new branch off 'master', without checking it out
+   $ git branch <branch-name> master
+   # check out the branch
+   $ git checkout <branch-name>
 
 To find out which branch you are in now::
 
@@ -69,7 +74,7 @@ To list all the branches, including the remote branches::
 
 To switch to a different branch::
 
-   $ git checkout another-branch-name
+   $ git checkout <another-branch-name>
 
 
 Delete Local Branch
@@ -77,7 +82,7 @@ Delete Local Branch
 
 To delete branch that you no longer need::
 
-   $ git branch -D branch-to-delete
+   $ git branch -D <branch-name>
 
 
 Staging and Committing Files
@@ -89,11 +94,11 @@ Staging and Committing Files
 
 2. To stage the files to be included in your commit::
 
-      $ git add /path/to/file1 path/to/file2 path/to/file3
+      $ git add path/to/file1 path/to/file2 path/to/file3
 
 3. To commit the files that have been staged (done in step 2)::
 
-      $ git commit -m "This is the commit message. Prefix it with bpo-XXXX."
+      $ git commit -m "bpo-XXXX: This is the commit message."
 
 
 Reverting Changes
@@ -101,7 +106,7 @@ Reverting Changes
 
 To revert changes to a file that has not been committed yet::
 
-    $ git checkout path/to/file
+   $ git checkout path/to/file
 
 If the change has been committed, and now you want to reset it to whatever
 the origin is at::
@@ -129,8 +134,8 @@ them to the remote repository.
 
 ::
 
-   $ git checkout some-branch
-   $ git push origin some-branch
+   $ git checkout <branch-name>
+   $ git push origin <branch-name>
 
 
 Creating a Pull Request
@@ -140,7 +145,7 @@ Creating a Pull Request
 
 2. Click ``compare across forks`` link.
 
-3. Select the base fork: ``python/cpython`` and base branch: ``master``
+3. Select the base fork: ``python/cpython`` and base branch: ``master``.
 
 4. Select the head fork: ``<username>/cpython`` and base branch: the branch
    containing your changes.
@@ -153,11 +158,11 @@ Syncing With Upstream
 
 Scenario:
 
-- You forked cpython repository some time ago.
+- You forked the CPython repository some time ago.
 - Time passes.
-- There have been new commits made in upstream cpython repository.
-- Your forked cpython repository is no longer up to date.
-- You now want to update your forked cpython repository to be the same as
+- There have been new commits made in upstream CPython repository.
+- Your forked CPython repository is no longer up to date.
+- You now want to update your forked CPython repository to be the same as
   upstream.
 
 Solution::
@@ -173,7 +178,7 @@ Another scenario:
 - You created ``some-branch`` some time ago.
 - Time passes.
 - You made some commits to ``some-branch``.
-- Meanwhile, there are recent changes from upstream cpython repository.
+- Meanwhile, there are recent changes from upstream CPython repository.
 - You want to incorporate the recent changes from upstream into ``some-branch``.
 
 Solution::
@@ -186,34 +191,15 @@ Solution::
 Backporting Merged Changes
 --------------------------
 
-When a pull request has been merged to master, and it needs to be backported
-into one of the maintenance branches.
+A pull request may need to be backported into one of the maintenance branches
+after it has been accepted and merged into ``master``.  It is usually indicated
+by the label ``needs backport to X.Y`` on the pull request itself.
 
-First, obtain the commit sha1 from the merged pull request:
+Use the utility script `cherry_picker.py <https://github.com/python/core-workflow/tree/master/cherry_picker>`_
+from the `core-workflow  <https://github.com/python/core-workflow>`_
+repository to backport the commit .
 
-1. Go to the merged pull request page, for example::
-
-     https://github.com/python/cpython/pull/PR-ID
-
-2. Scroll down and find the activity that says something like::
-
-     CoreDeveloper merged commit <hash> into python:master ...
-
-3. Follow the link to <hash>.
-
-4. Copy the complete hash value.
-
-The commit hash will be used below.
-
-To backport the commit to 3.6::
-
-   $ git fetch upstream
-   $ git checkout -b backport-someissue-3.6 upstream/3.6
-   $ git cherry-pick -x hashvalue
-   $ git push origin backport-someissue-3.6
-
-Go to https://github.com/python/cpython to create the pull request.  Select
-``3.6`` as the base branch, and ``backport-someissue-3.6`` as the head branch.
+The core developer who merged the pull request is expected to do the backport.
 
 
 Downloading Other's Patches
@@ -221,7 +207,7 @@ Downloading Other's Patches
 
 Scenario:
 
-- A contributor made a pull request to cpython.
+- A contributor made a pull request to CPython.
 - Before merging it, you want to be able to test their changes locally.
 
 Set up the following git alias::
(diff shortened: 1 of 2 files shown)


gitbootcamp.rst
```diff
@@ -0,0 +1,244 @@
...
+- There have been new commits made in upstream cpython repository.
+- Your forked cpython repository is no longer up to date.
+- You now want to update your forked cpython repository to be the same as
+  upstream.
+
+Solution::
+
+   $ git checkout master
+   $ git fetch upstream
+   $ git rebase upstream/master
```
brettcannon: Why do you have to rebase the master branch if you haven't committed to it? Is this a step that `git pull` does for you?

gitbootcamp.rst
```diff
@@ -0,0 +1,244 @@
...
+To backport the commit to 3.6::
+
+   $ git fetch upstream
+   $ git checkout -b backport-someissue-3.6 upstream/3.6
+   $ git cherry-pick -x hashvalue
+   $ git push origin backport-someissue-3.6
+
+Go to https://github.com/python/cpython to create the pull request.  Select
+``3.6`` as the base branch, and ``backport-someissue-3.6`` as the head branch.
+
```
brettcannon: You forgot to mention your script. 😄

brettcannon:
LGTM, but honestly I think having Ezio look over this would be great since I know he has asked for more of a git intro in the devguide.

gitbootcamp.rst
willingc: Difference in workflow approach. Some (myself included) use rebase/fetch instead of git pull.

brettcannon: Why is that? What's the pros/cons of the two approaches? (Basically why do you like the extra steps since this is meant for people not familiar with git and thus we might want to prefer the fewer-steps approach?)

willingc: `git pull --rebase` would be similar in approach and one could configure `git pull` to use rebase by default instead of merge.

pfmoore: I've only ever used ```git pull upstream master``` in any other projects I've used. To me it says more clearly what it's doing - "pull what's new on upstream onto my master branch". AFAIK, that's the normal advice (I don't recall where I learned the workflow from, but likely from various tutorials online). I think we should stick to the "normal" approach, as best as we can determine it, so we don't get newcomers being confused as to why we do things differently on Python.

I'd assume that fetch/rebase is intended more for cases where you're making changes on your local master branch. But my normal workflow is to only ever have master as a pristine copy of upstream's master, so ```pull upstream master``` says what I'd expect.

gitbootcamp.rst
Mariatta: Once it's moved to core-workflow, I'll update this :)

gitbootcamp.rst
Mariatta: I guess `git pull` works in this scenario (`master` branch).
But in the scenario below, people need to do a rebase. This provides a consistent workflow throughout I think.

bitdancer: I haven't read the rest of the doc, just noticed this discussion.  I agree with Brett, it seems much simpler to recommend 'git checkout master; git pull'.  Rebase is not needed here, and IMO will just set the wrong expectations in the mind of the beginner.  Now, if you were doing 'git checkout mybranch; git fetch upstream; git rebase upstream/master', *that* would make sense.

bitdancer: And in case it isn't clear, 'git pull' is *not* equivalent to 'git fetch; git rebase upstream/master'.  If you want that behavior you say 'git pull --rebase'.  But that only applies if you have changes committed to the branch your are pull --rebasing.

Unless, of course, my understanding of it is off :)

zware: `git pull upstream master` is roughly equivalent to `git fetch upstream && git merge --ff upstream/master`.  Adding the `--rebase` flag to the `pull` call turns it into `git fetch upstream && git rebase upstream/master`, which works out the same way when the history of your branch matches the history of `upstream/master`.  Recommending `git pull --rebase upstream master` covers the most common situations the most cleanly (including both scenarios listed here).

Personally, I hadn't previously thought through `git pull --rebase upstream master`, so I don't use it yet; I generally do `git pull upstream master` on master and do whatever other rebasing I need manually.  I expect that will change if I can remember it, though :)

gitbootcamp.rst
```diff
@@ -0,0 +1,244 @@
...
+      UserX wants to merge 1 commit into python:master from UserX:fixspamlib
+
+3. You'll need to first clone their cpython repo, and then checkout their
+   branch::
+
+      $ mkdir UserX
+      $ cd UserX
+      $ git clone https://github.com/UserX/cpython
+      $ cd cpython
+      $ git checkout fixspamlib
```
zware: This can be simplified significantly.  GitHub offers pull requests as remote branches that are not pulled automatically, but you can pull them explicitly by doing `git fetch upstream pull/${1}/head:pr/${1} && git checkout pr/${1}` where `${1}` is the PR number.  @berkerpeksag has this nicely packaged up as an alias:

```
[alias]
        pr = !sh -c \"git fetch upstream pull/${1}/head:pr/${1} && git checkout pr/${1}\" -
```

which allows one to do `git pr 615` to fetch and check out python/cpython#615.  The local branch name (`pr/${1}` or `pr/615` here) can be adjusted as you see fit; I might prefer `pr_${1}`.

gitbootcamp.rst
willingc: If you develop locally on a feature branch, then using:
```
git fetch upstream
git rebase upstream/master
```
puts all of your changes on top of the HEAD of master and results in less merge conflicts down the road. It's a very common development pattern with git.

gitbootcamp.rst
Mariatta: Very interesting.. I didn't know about this :)

gitbootcamp.rst
willingc: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/working-with-git/#after-upstream-has-changed

pfmoore: Understood but this section is about keeping your local master branch in sync with upstream. I don't see it mentioned anywhere, but I'd strongly recommend keeping local master as a pristine copy of upstream master, and only making local changes on branches. Assuming that's the approach, having a different command for "keeping my mirror of an upstream branch in sync" than "rebasing my local changes to reflect changes in master" seems entirely sensible to me.

I'd actually argue that you shouldn't (even) do fetch/rebase on a local feature branch, but should rather pull upstream master onto local master, then rebase your changes against local master. That's what I typically do (although in practice, most of the projects I work on are fine with not rebasing, so there's a lot less need).

willingc: @pfmoore We're saying the same thing. We're just approaching it from a different ordering of commands. As long as one keeps local, origin, and upstream remotes in mind, it really is personal preference. I never use `git pull` though it's completely fine to use it. FWIW, I rarely in the 3000 commits from the past year have a merge conflict of any kind.

pfmoore: OK, cool. I wasn't sure if this was intended as a recommendation, is all - to me, it's more confusing than a simple ```git pull```, and I'd tend to go for the simpler suggestion by default. But if we're saying the same thing, that's fine :-)

brettcannon: I think @pfmoore picked up on my initial confusion. As the instructions are just to keep master in sync then the rebase step seemed odd to me since you're not rebasing any previous commits you made locally. Maybe a comment at the start saying "there are alternative ways to perform the synchronization, but the approach taken to provide steps that are the same across scenarios"?

zware: That's why I suggest `git pull --rebase upstream master`.

willingc: I think @zware's suggestion is the best in terms of simplicity and flexibility.

pfmoore: I'd be inclined to prefer @zware's  ```git pull --rebase upstream master```, with a comment "you only need ```--rebase``` if you have local changes to the branch", That's simple enough for newcomers, explains when ```--rebase``` is needed, and remains consistent across scenarios.

gitbootcamp.rst
willingc: As an FYI (though we don't need to document), [hub](https://hub.github.com/) makes a maintainer's job easier (similar to @zware's mention of @berkerpeksag's alias).

Mariatta:
Made some adjustments:
- use the `git pull --rebase upstream master` approach
- explain how to create `git pr` alias
- explain how to merge and squash a PR
- provide example of good commit message

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+4. A fork will be created at https://github.com/UserName/cpython.
+
+
+Cloning The Forked CPython Repository
+-------------------------------------
+
+From your command line::
+
+   $ git clone git@github.com:UserName/cpython.git
```
brettcannon: Might want to add a note about substituting `UserName` or use something like `<username>` in case UserName on GitHub actually has CPython forked. 😉

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+-------------------------------
+
+.. note::
+   Never commit directly to the `master` branch.
+
+Create a new branch::
+
+   $ git checkout -b some-branch master # creates a new branch off master
+
+which is equal to::
```
brettcannon: Might want to mention it's equivalent *if* you are starting from the master branch, otherwise explicitly add the `master` part to the `git branch` step.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+local copy of a pull request as follows::
+
+   $ git pr <pr_number>
+
+For example, to fetch and checkout pull request #777::
+
+   $ git pr 777
+
+
+Accepting and Merging Pull Request
```
brettcannon: "Accepting and Merging a Pull Request"

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+Creating a Pull Request
+-----------------------
+
+1. Go to https://github.com/python/cpython.
+
+2. Click ``compare across forks`` link.
+
+3. Select the base fork: ``python/cpython`` and base branch: ``master``
+
+4. Select the head fork: ``UserName/cpython`` and base branch: the branch
```
brettcannon: `<username>` for `UserName`

gitbootcamp.rst
Mariatta: Thanks, I made the change in line 56.

gitbootcamp.rst
Mariatta: Done :)

gitbootcamp.rst
```diff
@@ -45,15 +45,15 @@ Creating and Switching Branches
...
+   Never commit directly to the ``master`` branch.
 
 Create a new branch::
 
    $ git checkout -b some-branch master # creates a new branch off master
 
 which is equal to::
 
-   $ git branch some-branch # create 'some-branch', without checking it out
+   $ git branch some-branch master # create 'some-branch' off 'master, without checking it out
```
brettcannon: I think you're missing an apostrophe at the end of "master".

Mariatta: Right 😅  Thanks. Fixed.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+The alias only needs to be done once.  After the alias is set up, you can get a
+local copy of a pull request as follows::
+
+   $ git pr <pr_number>
+
+For example, to fetch and checkout pull request #777::
+
+   $ git pr 777
+
```
ezio-melotti: Another scenario that is not covered here is: how can I update a PR that I pulled from a contributor (e.g. to add Misc/NEWS)?
This came up on #python-dev a few days ago, and afaiu:
* you can add your changes and push to the contributor repo using `git push -u git@github.com:<contributor>/cpython.git <branch>` (or add a remote with `git remote add <contributor> git@github.com:<contributor>/cpython.git` and then `git push -u <contributor> <branch>`).  In order to do this the contributor must have allowed the maintainers to edit the prs (should be allowed by default).
* you can add your changes, push the branch to your fork, and create a new and separate PR that replaces the original one.

It was not clear how the author is determined if multiple users (contributor and core-dev) worked on the same PR: author of the first commit? owner of the repo the PR comes from? author with most commits/loc in the PR?

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+-----------------
+
+To revert changes to a file that has not been committed yet::
+
+    $ git checkout path/to/file
+
+If the change has been committed, and now you want to reset it to whatever
+the origin is at::
+
+   $ git reset --hard HEAD
```
ezio-melotti: There are two more scenarios where I used `hg revert` that are not covered here:
1. how to revert all changes in the working copy (i.e. the equivalent of `hg revert -a` -- `git reset`? `git reset --hard`?).
2. how to revert the fix while leaving the test to verify if they fail without the fix.

The second case works differently since we are using PRs.  With HG I could apply a patch and do `hg revert Lib/` to revert all the changes in `Lib` while leaving the tests unchanged.  I think the git equivalent would be something like `git checkout master Lib/`, after I pulled a PR as described in a later section.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+For example, to fetch and checkout pull request #777::
+
+   $ git pr 777
+
+
+Accepting and Merging A Pull Request
+------------------------------------
+
+Pull requests can be accepted and merged by a Python Core Developer.
```
ezio-melotti: A question about "merging etiquette": if another core-dev submitted a PR, and it looks good to me, should I go ahead and merge it, or should I just leave a positive review and let the original core-dev merge it?
I think the latter is better, but I've seen the former happen.
Perhaps this could be clarified here.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+   $ git pr <pr_number>
+
+For example, to fetch and checkout pull request #777::
+
+   $ git pr 777
+
+
+Accepting and Merging A Pull Request
+------------------------------------
```
ezio-melotti: This should be marked as core-devs only, since contributors can't accept/merge.

willingc:
@ezio-melotti Wouldn't it make more sense since we are now on GitHub to merge this PR as is and then create a new PR for any other changes you wish to make?

ezio-melotti:
That would be fine with me, since there are no major problems with the patch.
I'll change my review to "approved" as not to block this, but let @Mariatta decide whether she wants to address the comments in this PR, create a separate PR, or leave it to me (in that case I still need some feedback for the comments I made).

willingc:
Thanks @ezio-melotti. Your flexibility is appreciated. I suspect this will get iterated over time as well. 👍

Mariatta:
Thanks for reviewing, @ezio-melotti 😃 
My intention with this page is not to provide a full tutorial of git, but to provide a quick reference page. 
For example, when a contributor forgot the command of how to rebase, they can go to this page and quickly find the command they need.
I also want this page to be useful at sprints where the goal is just to know enough git commands to start contributing to the *real thing*  😉   
Because of this, I think this page should be kept brief and concise, and not too wordy. 
There are various ways one can work with git to achieve the same results. Individuals can still choose to work with a workflow that they're comfortable with.

I will prepare another update to this PR later today to address some of your feedback.

There are definitely more topics that need to be covered here, but we can address them in future PRs :) 
for example
- how to resolve merge conflict
- `git log`
- `git show`

Thanks :)

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+Solution::
+
+   $ git checkout some-branch
+   $ git fetch upstream
+   $ git rebase upstream/master
+
+
+Backporting Merged Changes
+--------------------------
```
Mariatta: I think core dev who merged the PR is expected to do the backport, mainly because it involves removing ``needs backport to X.Y`` label, and applying ``cherry-pick for X.Y`` label.  Contributor don't have the right to apply and remove labels themselves.  It's fine to ask the contributor to do the cherry-pick themselves, but core devs should still apply/remove the appropriate labels.

For now I will just state here that the core dev is expected to do the backport. Once the cherry-picking bot is in place, this section will just be obsolete.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+To backport the commit to 3.6::
+
+   $ git fetch upstream
+   $ git checkout -b backport-someissue-3.6 upstream/3.6
+   $ git cherry-pick -x hashvalue
+   $ git push origin backport-someissue-3.6
+
+Go to https://github.com/python/cpython to create the pull request.  Select
+``3.6`` as the base branch, and ``backport-someissue-3.6`` as the head branch.
```
Mariatta: Compare & create PR by default will select the `python:master` as base branch, we'll need to manually select the appropriate base branch. :)

With cherry_picker.py, the correct maintenance branch will be pre-selected :)
So I'm replacing this whole section with a reference to cherry_picker.py, now that it's been migrated to the core-workflow repo.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+---------------------------
+
+Scenario:
+
+- A contributor made a pull request to cpython.
+- Before merging it, you want to be able to test their changes locally.
+
+Set up the following git alias::
+
+   $ git config --global alias.pr '!sh -c "git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}" -'
```
Mariatta: This alias was suggested by @zware. I don't actually know about the other variants. So I'll just leave this as is.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+The alias only needs to be done once.  After the alias is set up, you can get a
+local copy of a pull request as follows::
+
+   $ git pr <pr_number>
+
+For example, to fetch and checkout pull request #777::
+
+   $ git pr 777
```
Mariatta: I'll remove the example to make it consistent.

gitbootcamp.rst
Mariatta: I haven't had to do this myself, so I don't have the answer ...

gitbootcamp.rst
Mariatta: Mentioned in the next line.

gitbootcamp.rst
Mariatta: Perhaps this is to be asked at core-workflow or python-committers? :)

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+   $ git checkout another-branch-name
+
+
+Delete Local Branch
+-------------------
+
+To delete branch that you no longer need::
+
+   $ git branch -D branch-to-delete
```
Mariatta: The capital `-D` indicates 'force delete', that will delete branch that has not yet been merged.
I use this in cherry_picker.py and just got used to it.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+Staging and Committing Files
+----------------------------
+
+1. To show the current changes::
+
+      $ git status
+
+2. To stage the files to be included in your commit::
+
+      $ git add /path/to/file1 path/to/file2 path/to/file3
```
Mariatta: Will remove.

gitbootcamp.rst
```diff
@@ -0,0 +1,268 @@
...
+
+      $ git commit -m "This is the commit message. Prefix it with bpo-XXXX."
+
+
+Reverting Changes
+-----------------
+
+To revert changes to a file that has not been committed yet::
+
+    $ git checkout path/to/file
```
Mariatta: Fixed in the next commit.

gitbootcamp.rst
Mariatta: Not sure... I don't normally use the ``^``

ezio-melotti:
The latest changeset LGTM.

Mariatta:
Thanks for all the feedback and for reviewing 🙋 💯 :tada:

PR #147, opened 2017-03-14
Use https instead of http if supported

methane:
and fix some old links.

### DIFF
--- a/docquality.rst
+++ b/docquality.rst
@@ -16,16 +16,16 @@ will look along with validating that your new markup is correct).
 
 The documentation built from the :ref:`in-development <indevbranch>` and
 :ref:`maintenance <maintbranch>` branches can be viewed from
-http://docs.python.org/dev/.  The in-development and most recent 2.x and 3.x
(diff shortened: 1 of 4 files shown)


docquality.rst
```diff
@@ -98,4 +98,4 @@ that may be different from the main documentation.
 .. _separate repository:
 .. _devguide repo: https://github.com/python/devguide
 .. _the GitHub tracker: https://github.com/python/devguide/issues
-.. _Sphinx: http://sphinx.pocoo.org/
+.. _Sphinx: http://www.sphinx-doc.org/en/stable/
```
haypo: I suggest to drop /en/stable/

haypo:
LGTM except of a minor suggestion. I think that last one who wrote a similar change was Serhiy, so I added him as reviewer. I let him vote on your change :-)

docquality.rst
methane: Agree, URL used at https://pypi.python.org/pypi/Sphinx doesn't include /en/stable.

PR #148, opened 2017-03-15
remove references to spark.py

DimitrisJim:
In `compiler.rst` a reference to `spark.py`, used until Python `3.4`, is still present. [issue19655](https://bugs.python.org/issue19655) changed that for `Python >= 3.5` so, this PR simply removes the references to `spark.py` and changes the text under `asdl.py`.


brettcannon:
Thanks!

PR #153, opened 2017-03-27
Be a bit more strong with getting the CLA signed

### LINKED ISSUE #142: Explain how to fill out the CLA
Before contributing to Python, I searched the developer guide (https://docs.python.org/devguide/) for "CLA", to make sure I would not have to file one. The string "CLA" does not appear on that page, and there was no reference to having to sign something in the quickstart section, so I assumed there was no need.

When I actually sent my pull request, I received a github message claiming signing the PSF CLA is "necessary for legal reasons before we can look at your contribution".

After signing it, I checked the dev guide more thoroughly. On https://docs.python.org/devguide/committing.html#contributor-licensing-agreements, it says "It’s unlikely bug fixes will require a Contributor Licensing Agreement unless they touch a lot of code" (which my pull request doesn't).

There are two (and a half) problems here:

- The dev guide and the github message cannot both be right. One should be updated to match the other.

- If signing a CLA is required, I think it's a good idea to mention this early on in the dev guide. If for whatever reason a contributor cannot or will not sign the CLA, they should find out about it before they have produced and sent off a pull request.

- The abbreviation "CLA" might be common enough it's worth including in the dev guide, to make sure someone searching for that string (like I did) finds the right section.

brettcannon:
Closes #142

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -148,11 +148,10 @@ and redistribute their code.
 Contributor Licensing Agreements
 --------------------------------
 
-It's unlikely bug fixes will require a `Contributor Licensing Agreement`_
-unless they touch a *lot* of code. For new features, it is preferable to
-ask that the contributor submit a signed CLA to the PSF as the associated
-comments, docstrings and documentation are far more likely to reach a
-copyrightable standard.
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @brettcannon

Issue #161, opened 2017-04-10
Document when it's okay to dismiss another core dev's review


brettcannon:
If a core dev requests changes, when it is okay to dismiss that request? Discussion in python/core-workflow#36 and the reached conclusion that we should probably document.

Thtboi420:
Sent using Zoho Mail


---- On Mon, 10 Apr 2017 16:36:01 -0700 notifications@github.com wrote ----

If a core dev requests changes, when it is okay to dismiss that request? Discussion in python/core-workflow#36 and the reached conclusion that we should probably document.

PR #176, opened 2017-04-24
Document when a review can be dismissed.

Mariatta:
Closes https://github.com/python/devguide/issues/161


willingc:
Thanks @Mariatta 👍

Issue #173, opened 2017-04-21
Update the section about backporting changes


Mariatta:
The section: 
http://cpython-devguide.readthedocs.io/committing.html?highlight=backport#backporting-changes-to-python-3-6-or-older-version
has instructions on how to apply a patch to master, which is not desirable in the current workflow. We should be working on branches other than master.

I think this should be updated somehow. Maybe also consider merging it with the next section http://cpython-devguide.readthedocs.io/committing.html?highlight=backport#backport-and-cherry-pick-labels

ncoghlan:
I'd suggest updating it to show using `git log -10 --oneline` to get a list of the recent commit hashes and the first line of their commit messages.

PR #178, opened 2017-04-25
Update backport instructions

Mariatta:
Remove the part about merging into master.
Remove the part about prefixing the PR title with `X.Y`, cherry_picker.py already does this.

Closes https://github.com/python/devguide/issues/173

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -408,45 +408,25 @@ new features.  The other branches only receive bug fixes or security fixes.
 
 .. _branch-merge:
 
-Backporting changes to Python 3.6 (or older version)
-----------------------------------------------------
+Backporting Changes to Older Version
+------------------------------------
 
-The current in-development version of Python is in the ``master`` branch.  To properly
-port the patch to Python 3.6 (or older version), you should first apply the patch
-to master branch::
-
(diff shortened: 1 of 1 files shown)


committing.rst
```diff
@@ -408,45 +408,25 @@ new features.  The other branches only receive bug fixes or security fixes.
 
 .. _branch-merge:
 
-Backporting changes to Python 3.6 (or older version)
-----------------------------------------------------
+Backporting Changes to Older Version
```
willingc: @Mariatta: Should there be an `an` before `Older Version`?

Mariatta:
Thanks :)  Updated.

berkerpeksag:
I think the part about prefixing PR title should stay for now. Not every core dev use cherry_picker.py yet :)

willingc:
A reasonable comment during the transition @berkerpeksag. Perhaps adding a small note about prefixing the PR title and marking it "legacy process" and the cherry_picker.py as the "preferred process".

@Mariatta thoughts?

Mariatta:
Sounds reasonable :) I'm reviewing the other PR now, and will add back the PR prefix after.

PR #180, opened 2017-04-26
Minor tweaks in documenting.rst

DimitrisJim:
Two minor tweaks in wording and a fix of the `attribute` directive example that used `data` instead of `attribute`.


berkerpeksag:
Thanks!

PR #198, opened 2017-05-13
We don't need to apply `cherry-pick for *` label anymore.

Mariatta:
lol also just now noticed the typo 😛 

Background: https://mail.python.org/pipermail/core-workflow/2017-May/000991.html

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -429,8 +429,8 @@ Prefix the backport pull request with the branch, for example::
 Note that cherry_picker.py_ adds the branch prefix automatically.
    
 Once the backport pull request has been created, remove the
-``needs backport to X.Y`` label from the original pull request on ``master``.
-Only Core Developers can apply labels to GitHub pull requests.
+``needs backport to X.Y`` label from the original pull request.  (Only Core
+Developers can apply labels to GitHub pull requests).
 
 .. _cherry_picker.py: https://github.com/python/core-workflow/tree/master/cherry_picker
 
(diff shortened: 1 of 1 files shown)


committing.rst
```diff
@@ -428,10 +428,9 @@ Prefix the backport pull request with the branch, for example::
    
 Note that cherry_picker.py_ adds the branch prefix automatically.
    
-Apply the label ``cherry-pick or X.Y`` to the backport pull request.  Once the backport
-pull request has been created, remove the ``needs backport to X.Y`` label from the
-original pull request on ``master``.  Only Core Developers can apply labels to GitHub
-pull requests.  
+Once the backport pull request has been created, remove the
+``needs backport to X.Y`` label from the original pull request on ``master``.
```
brettcannon: It's actually possible that the cherry-pick is on master, e.g. if something is fixed in `master` and so initially was patched against `3.6` and then backported to `3.5`.

committing.rst
```diff
@@ -428,10 +428,9 @@ Prefix the backport pull request with the branch, for example::
...
+Only Core Developers can apply labels to GitHub pull requests.
```
brettcannon: Might as well put this in parentheses to de-emphasize it as people familiar with GH will know this.

committing.rst
Mariatta: I removed the phrase `on master` for now.
But it sounds like you're describing a new workflow?
e.g. PR against 3.6, and then cherry-pick into 3.5 and master. In this situation, we should advise people to prepare the PR against master first, and then backport into 3.6 and 3.5.
We also don't have `needs backport to master` label.

committing.rst
Mariatta: Done :) Thanks.

Issue #7, opened 2016-07-22
Add a Sprints organization/preparation section to devguide


brettcannon:
Originally from http://bugs.python.org/issue24016

matrixise:
@willingc I have started a new section with your comment from https://bugs.python.org/issue24016 

I have organised a sprint on 13rd and 14th october and I would like your feedback and maybe improve the text. https://github.com/matrixise/devguide/blob/add-sprint-section/sprints.rst

I can give you the access rights on this repo if you think you can help me.

matrixise:
I think we can add @bitdancer in the loop, because he has organised the sprints during PyCon US.

willingc:
Hi @matrixise,

Great start. :sunny: Feel free to open a WIP PR and I can comment (just one or two wording things) or access to your fork. The first may be simpler since it keeps all the discussion in one place.

Overall, I would call this a checklist (similar in spirit to what we created for OpenHatch https://github.com/openhatch/in-person-event-handbook/blob/master/checklists.pdf). OpenHatch's Open Source Event Handbook: http://opensource-events.com is another good resource.

Add to checklist under "during the event":

- Have a file/page that folks can add their contact info (name and GitHub userID) during the sprints (all contributors even without a PR are recognized as contributors i.e. user test of installation instructions is a contribution as valuable as code)

Suggest creating a static site that provides the sprint details:

- SciPy 2016 sprint (lektor): http://jupyter.org/scipy-sprint/
- Grace Hopper Open Source Day (nikola): https://willingc.github.io/ghc-openhatch-2016/welcome/

Perhaps creating a cookiecutter that takes the sprint details and autopopulates the static site would be a good first contributor project. (cc/ @Mariatta)

matrixise:
@willingc you have received the access rights for my repo, I would be really happy if you want to contribute to this part.

willingc:
Thanks @matrixise. Will do later this weekend. I'm actually helping run a hackathon today at a local university.

matrixise:
I am going to work on this part during this week-end. @willingc, did you have the opportunity to work/think about this section ?

Mariatta:
@matrixise @willingc Perhaps once you have the checklist ready, I can work on creating the cookie cutter project based on it. Thanks 😃

matrixise:
@Mariatta with pleasure, and of course, I can help you once you have a small proto.

willingc:
[...]

PR #201, opened 2017-05-18
PyCon Sprint Guide

Mariatta:
Provide several easy issues for first time contributors
Provide ideas for other long term contributions

Somewhat related to https://github.com/python/devguide/issues/7

### DIFF
--- a/sprintguide.rst
+++ b/sprintguide.rst
@@ -83,6 +83,8 @@ Dev Guide Issues
 
 #. Improve commit message `python/devguide#199 <https://github.com/python/devguide/issues/199>`_.
 
+#. Mention the core-workflow mailing list / repo `python/devguide#202 <https://github.com/python/devguide/issues/202>`_.
+
 
 Easy B.P.O Issues
 -----------------
--- a/sprintguide.rst
+++ b/sprintguide.rst
@@ -1,0 +1,141 @@
+.. highlight:: console
+
+.. _sprintguide:
+
+PyCon Sprint Guide
+==================
+
+This guide is written for PyCon US 2017 sprint.  For other sprints, we may need
+to come up with a different set of easy b.p.o. issues.
+
+Everyone at the sprint needs to follow the `PSF Code of Conduct <http://python.org/psf/codeofconduct/>`_.
+
+
+.. contents::
+
+Installations
+-------------
+
+The following resources take time to download.  You may not want to rely on the
+on the conference venue's wifi to download these.  Do them before you come to
+the sprint.
+
+#. Go through the :ref:`setup` guide.  Install the necessary dependencies
+   for your machine.
+
+#. Install sphinx v1.6.1::
+
+      pip install sphinx~=1.6.1
+
+#. Fork and clone the CPython repo.  Set your clone as the ``origin``, and
+   ``https://github.com/python/cpython`` as ``upstream``.
+
+#. Fork and clone the Dev Guide repo. Set your clone as the ``origin``, and
+   ``https://github.com/python/devguide`` as ``upstream``.
+
+Reading materials
+-----------------
+
+#. :ref:`Quick Start <index>` section of CPython Developer's Guide.
+
+#. :ref:`Documenting Python <documenting>`.
+
+#. :ref:`gitbootcamp`.
+
+#. :ref:`exploring`.
+
+#. :ref:`helptriage`.
+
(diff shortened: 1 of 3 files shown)


sprintguide.rst
```diff
@@ -0,0 +1,143 @@
...
+
+#. Core-mentorship `mailing list <https://mail.python.org/mailman/listinfo/core-mentorship/>`_.
+
+#. Comment on the issue, say that you're working on it. If an issue is assigned,
+   please ask first.
+
+#. #python-dev on `freenode <http://webchat.freenode.net/>`_.
+
+
+Dev Guide Issues
```
orsenthil: We could have a shared Google Docs/ Spreadsheet / Label these issues with pyconsprint2017 (etc), instead of listing them in a PR. And we could mention that link here.

Because we will likely encounter more issues that will get added and we may not need a new PR to add an issue.

Mariatta: I wasn't sure myself where to list these issues, I realize the devguide may not be the best place.
But at least it becomes community-maintained.
Whereas if it's a google docs, someone has to take ownership of it.
My thinking is people should not add tickets randomly just because they think it's an easy ticket.
We should still vet each ticket to make sure that it is a legitimate issue that needs fixing/improvement. When it is listed in the devguide, it's a sign that it's been approved by a core developer.

orsenthil: In the past, we have had a shared document (similar to paste bin). That was the reason for my comment. And yes, usually a core dev takes an ownership to keep the doc updated. At the same other others can edit too.  (If ownership is a problem, I can volunteer to be an owner of shared doc).

I am not -1, just sharing a point that it will be a cumbersome to keep list alive.

zware: What if we enabled the wiki on this repo, and added this list as a page there?

orsenthil: That's a fantastic idea. I am +1 on that.

Mariatta: Sounds good.

I don't have access to add  the wiki :)

zware: @brettcannon, what say you?

matrixise: @zware yep, add a wiki page instead of this PR, because a sprint code session can change from a pycon to an other one.

willingc:
Let's rework this PR to provide basic guidance for a sprint organizer at a meetup or another event by removing the PyCon 2017 content to the wiki/etherpad as suggested. @Mariatta I likely have some sprint guides from other communities (SciPy etc) if you want to take a look.

brettcannon:
There is also https://github.com/python/devinabox which is meant to act as a guide for  people hosting a sprint for CPython.

sprintguide.rst
mlouielu: Or maybe we could enable devguide's wiki, and use this wiki to maintain the list of the b.p.o issue?

mlouielu:
Is there any limited to used to sprint guide in PyCon (or outside PyCon)?

If I want to make a sprint in PyCon Taiwan, could it use this as a guide?

Mariatta:
@mlouielu If you'd like to use this PR as a template for your own sprint, please go ahead :)

I agree that the sprint guide is better suited elsewhere (eg wiki?) but I likely won't have time to continue work on this until next few weeks...

Mariatta:
Just saw @brettcannon 's comment about devinabox. Should this 'guide' be moved there then, instead of a wiki?

willingc:
Seems like a reasonable plan to move most of the "How to run a sprint" and any future templates to support running a sprint to devinabox (or new sprints repo).  The devguide can guide users, organizers, and contributors to the sprint resources.

willingc:
@mlouielu Since I'll be at PyCon Taiwan, I would be happy to participate in a CPython sprint.

Mariatta:
Alright I'll close this and open an issue in devinabox :) Thanks all

ncoghlan:
Just a note regarding the wiki: due to persistent spam problems, getting edit access to the wiki is a bit of a pain, and there's also no easy way to institute a collaborative review process for proposed changes.

So for most things, the GitHub+ReadTheDocs combination is going to provide a superior experience - you still get online editing support, anyone with a GitHub account can submit a PR, and the level of review required can vary based on the sensitivity of the content (e.g. devinabox could likely operate on the basis of "Accepted PR -> Merge privileges if you want them")

PR #203, opened 2017-05-21
PEP 545: Add mdk as French translation coordinator.

JulienPalard:
I'm following steps by step the [New Translation Procedure](https://www.python.org/dev/peps/pep-0545/#new-translation-procedure) according to [Migrate Github Repositories](https://www.python.org/dev/peps/pep-0545/#migrate-github-repositories) from [PEP 545](https://www.python.org/dev/peps/pep-0545).


ncoghlan:
Thanks, and congratulations on getting PEP 545 accepted :)

PR #206, opened 2017-05-22
Add help for building the devguide documentation

cjrh:


the-knights-who-say-ni:

I'm a bot set up to make sure that the project can legally accept your contribution by verifying you have signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA. This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://cpython-devguide.readthedocs.io/pullrequest.html#licensing) to rectify this issue.


Thanks again to your contribution and we look forward to looking at it!

matrixise:
Maybe we could add a `requirements.txt` file with your PR. @zware ?

matrixise:
Don't answer to my comment about the `requirements.txt` file, I see it in your PR.

BUILDING.rst
```diff
@@ -0,0 +1,19 @@
...
+Here is an example for building the HTML version of the devguide:
+
+.. code:: shell
+
+   $ python3 -m venv env
+   $ source env/bin/activate
+   (env) $ pip install -r requirements.txt
+   (env) $ make html
+
+After the processing completes, the new HTML documents will be available in the ``_build/html``
```
Mariatta: I would say "after the build is completed"

cjrh: Thanks! I'll move these notes to the page you mentioned, and change the wording.

willingc: Thanks @cjrh. I like the more explicit directions that you have for working with the devguide code. It will be a nice addition to the page Mariatta references. 👍

cjrh:
@Mariatta I've moved the instructions for virtualenv and using `requirements.txt` into `docquality.rst`.

zware:
To mirror cpython, we could have a `make venv` target to take care of actually making the venv (as a separate exercise, you could go so far as to add targets and dependencies such that `make html` creates the venv for you and uses it, if it doesn't already exist.  Both here and in cpython).

cjrh:
@zware Good idea!  I've found the venv entry in the cpython/Doc Makefile, and I'm learning how to set up dependencies inside makefiles.

cjrh:
@zware @Mariatta I've now set up the Makefile to automatically create and use a virtualenv for all the targets requiring `sphinx-build`.  The docs have now gone back to simply `$ make html`, but I point out that the builder is going to create a virtualenv behind the scenes.

Mariatta:
As part of local testing this PR, I created a new virtual environment called `venv_pr206`, and activated that environment. When I ran `pip install ..`, the dependencies were installed in `venv_pr206`. 
I'm getting an error when trying to do `make html` because it is now looking for `venv`. 
So maybe we still need to document that people should be using the environment name `venv` locally?

cjrh:
@Mariatta I think the idea from @zware suggestion was that people no longer need to create a venv at all: simply `$ make html` does it all automatically.  I'll go check to see if the docs still say that a user must create a venv...

zware:
@Mariatta What was the error you got?  Was it due to already having a venv activated?

Mariatta:
I think it was because I have an existing virtual environment called `venv` prior to this.
Is this suppose to update my existing `venv` with the new requirements?

zware:
Hmm, that shouldn't be; can you reproduce it?  As is, it should probably just use any venv named `venv` without modifying it at all.  We could make the `venv` target depend on `requirements.txt` to try to keep things up to date.

Mariatta:
Ok I cleaned up my venv and tried this again.
Everything works :)

Mariatta:
Thanks everyone!
Congrats @cjrh on your first contribution to the Dev Guide 🎉

PR #212, opened 2017-05-27
Document the news file format

brettcannon:
This should not be merged until larryhastings/blurb relocates to python/core-workflow, python/cpython is updated appropriately, and python/core-workflow#66 is closed.

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -23,7 +23,7 @@ later in this document):
 #. Was ``configure`` regenerated (if necessary)?
 #. Was ``pyconfig.h.in`` regenerated (if necessary)?
 #. Was the submitter added to ``Misc/ACKS`` (as appropriate)?
-#. Was an entry added to ``Misc/NEWS.d`` (as appropriate)?
+#. Was an entry added under ``Misc/NEWS.d/next`` (as appropriate)?
 #. Was "What's New" updated (as appropriate)?
 #. Were appropriate labels added to signify necessary backporting of the
    pull request?
@@ -79,7 +79,7 @@ The automated patch checklist runs through:
   (using ``Tools/scripts/reindent-rst.py``)
 * Has the documentation been updated?
 * Has the test suite been updated?
-* Has an entry in ``Misc/NEWS.d`` been added?
+* Has an entry under ``Misc/NEWS.d/next`` been added?
 * Has ``Misc/ACKS`` been updated?
 * Has ``configure`` been regenerated, if necessary?
 * Has ``pyconfig.h.in`` been regenerated, if necessary?
--- a/committing.rst
+++ b/committing.rst
@@ -23,7 +23,7 @@ later in this document):
 #. Was ``configure`` regenerated (if necessary)?
 #. Was ``pyconfig.h.in`` regenerated (if necessary)?
 #. Was the submitter added to ``Misc/ACKS`` (as appropriate)?
-#. Was ``Misc/NEWS`` updated (as appropriate)?
+#. Was an entry added to ``Misc/NEWS.d`` (as appropriate)?
 #. Was "What's New" updated (as appropriate)?
 #. Were appropriate labels added to signify necessary backporting of the
    pull request?
@@ -32,7 +32,7 @@ later in this document):
    If you want to share your work-in-progress code on a feature or bugfix,
    either open a ``WIP``-prefixed PR, publish patches on the
    `issue tracker`_ or create a public fork of the repository.
-   
+
 .. _issue tracker: https://bugs.python.org
 
 
@@ -79,7 +79,7 @@ The automated patch checklist runs through:
   (using ``Tools/scripts/reindent-rst.py``)
 * Has the documentation been updated?
 * Has the test suite been updated?
-* Has ``Misc/NEWS`` been updated?
+* Has an entry in ``Misc/NEWS.d`` been added?
 * Has ``Misc/ACKS`` been updated?
 * Has ``configure`` been regenerated, if necessary?
 * Has ``pyconfig.h.in`` been regenerated, if necessary?
@@ -115,7 +115,7 @@ someone else. When doing so you will want to make sure of some things.
 First, make sure the patch is in a good state. Both :ref:`patch` and
 :ref:`helptriage`
 explain what is to be expected of a patch. Typically patches that get cleared by
-triagers are good to go except maybe lacking ``Misc/ACKS`` and ``Misc/NEWS``
+triagers are good to go except maybe lacking ``Misc/ACKS`` and ``Misc/NEWS.d``
 entries (which a core developer should make sure are updated appropriately).
 
 Second, make sure the patch does not break backwards-compatibility without a
@@ -127,12 +127,12 @@ is worth it, ask on python-dev.
 
 Third, ensure the patch is attributed correctly with the contributor's
 name in ``Misc/ACKS`` if they aren't already there (and didn't add themselves
-in their patch) and by mentioning "Patch by <x>" in the ``Misc/NEWS`` entry
+in their patch) and by mentioning "Patch by <x>" in the ``Misc/NEWS.d`` entry
 and the checkin message. If the patch has been heavily modified then "Initial
 patch by <x>" is an appropriate alternate wording.
 
 If you omit correct attribution in the initial checkin, then update ``ACKS``
-and ``NEWS`` in a subsequent checkin (don't worry about trying to fix the
+and ``NEWS.d`` in a subsequent checkin (don't worry about trying to fix the
 original checkin message in that case).
 
 Finally, make sure that the submitter of the
@@ -171,17 +171,17 @@ for inclusion.
    http://www.python.org/psf/contrib/contrib-form/
 
 
-What's New and NEWS Entries
+What's New and News Entries
 ---------------------------
 
-Almost all changes made to the code base deserve an entry in ``Misc/NEWS``.
+Almost all changes made to the code base deserve an entry in ``Misc/NEWS.d``.
 If the change is particularly interesting for end users (e.g. new features,
 significant improvements, or backwards-incompatible changes), an entry in
 the ``What's New in Python`` document (in ``Doc/whatsnew/``) should be added
 as well.
 
 There are two notable exceptions to this general principle, and they
-both relate to changes that *already* have a NEWS entry, and have not yet
+both relate to changes that *already* have a news entry, and have not yet
 been included in any formal release (including alpha and beta releases).
 These exceptions are:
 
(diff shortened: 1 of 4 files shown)


committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+News entries go into the ``Misc/NEWS.d`` directory as individual files. The
+easiest way to create a news entry is to use the
+`blurb <https://pypi.org/project/blurb/>`_ tool and its ``blurb add`` command.
+
+If you are unable to use the tool you can create the news entry file manually.
+The ``Misc/NEWS.d`` directory contains a sub-directory named ``next`` which
+itself contains various sub-directories representing classifications for what
+was affected (e.g. ``Misc/NEWS.d/next/Library`` for changes relating to the
+standard library). The file name itself should be of the format
+``<date>.<issue-number>.<nonce>.rst``:
```
larryhastings: Format is ``<date>.bpo-<issue-number>.<nonce>.rst``.  The fields between <date> and <nonce> are theoretically tagged, so we can add more metadata later as needed.

committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+
+* ``<date>`` is today's date in ``YYYY-MM-DD`` format, e.g. ``2017-05-27``
```
larryhastings: I actually emit YYYY-MM-DD-HH-MM-SS, both to further prevent collisions, and to ensure that entries are added in strict chronological order.  It doesn't make a big difference, but I'd like it if you mentioned it.   Changing the line to

* ``<date>`` is today's date in ``YYYY-MM-DD`` or ``YYYY-MM-DD-HH-MM-DD`` format, e.g. ``2017-05-27``or  ``2017-05-27-08-22-35``

committing.rst
larryhastings: Oh, I guess below you suggest that "bpo-" is considered part of the issue number?  That kind of suggests we'd have other things there besides bpo-, which right now I don't think blurb would permit.  It certainly wouldn't emit it.

brettcannon: Clarified in the latest commit.

committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+* ``<issue-number>`` is the issue number the change is for, e.g. ``bpo-12345``
+* ``<nonce>`` is a unique, six-character string to guarantee the file name is
```
larryhastings: There's no length requirement for the nonce.  I think even a zero-character string would work.

committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+  `test.support.temp_cwd()`.  Patch by Chris Jerdonek.
 
 (In all other ``.rst`` files, the single backticks should not be used.  They are
 allowed here because NEWS is meant to be as readable as possible unprocessed.)
 
-A nice trick to make merging across branches work more smoothly is to
-put a new entry after the first or first two entries rather than at the very
-top.  This way if you commit, pull new changesets and merge, the merge will
-succeed automatically.
+**No editing should be  done to ``Misc/NEWS``**. While this was the traditional
```
larryhastings: Actually, we decided to remove Misc/NEWS from the CPython tree.  It'll be generated just-in-time as part of the release process so tarballs look the same.  But it should no longer be checked in.

committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
 
-The NEWS file is now read by Sphinx to produce the "Changelog" page; accordingly
-it should be valid reStructuredText.  The "default role" (single backticks) can
-be used to refer to objects in the documentation.  Example NEWS entry::
+So a file name may be
+``Misc/NEWS.d/next/Library/2017-05-27.bpo-12345.Yl4gI2.rst``.
 
-   - bpo-15304: Fix warning message when `os.chdir()` fails inside
-     `test.support.temp_cwd()`.  Patch by Chris Jerdonek.
+A news file contents should be valid reStructuredText. An 80 character column
```
larryhastings: Bad English.  Better: "The contents of a news file should be [...]"

committing.rst
```diff
@@ -201,25 +201,40 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+width should be used. The "default role"
+(single backticks) in reST can be used to refer to objects in the documentation.
+Example news entry::
```
larryhastings: I'd also like this to mention that, unlike in Misc/NEWS, news entries in Misc/NEWS.d *shouldn't* start with a `-`, shouldn't indent subsequent lines, and shouldn't start with `Issue #` or `bpo-`.

terryjreedy: I was wondering about that ;-)

committing.rst
```diff
@@ -201,25 +201,41 @@ when implemented, these changes *must* be noted in the "New Additions in
 Python 2.7 Maintenance Releases" section of the Python 2.7 What's New
 document.
 
-New NEWS entries are customarily added at or near the top of their
-respective sections, so that entries within a section appear in approximate
-order from newest to oldest.  However, this is customary and not a
-requirement.
+News entries go into the ``Misc/NEWS.d`` directory as individual files. The
+easiest way to create a news entry is to use the
+`blurb <https://pypi.org/project/blurb/>`_ tool and its ``blurb add`` command.
```
terryjreedy: Will blurb add open some editor or editing window?

committing.rst
```diff
@@ -201,25 +201,41 @@ when implemented, these changes *must* be noted in the "New Additions in
...
+itself contains various sub-directories representing classifications for what
+was affected (e.g. ``Misc/NEWS.d/next/Library`` for changes relating to the
+standard library). The file name itself should be of the format
+``<date>.bpo-<issue-number>.<nonce>.rst``:
+
+* ``<date>`` is today's date in ``YYYY-MM-DD`` format, e.g. ``2017-05-27``
+* ``<issue-number>`` is the issue number the change is for, e.g. ``12345``
+  for ``bpo-12345``
+* ``<nonce>`` is a unique, six-character string to guarantee the file name is
+  unique across branches, e.g. ``Yl4gI2``
```
terryjreedy: This seems like a problem for hand written files.  Will the nonce remain the same in backported files?  (I can take 'unique across branches' either way.)

committing.rst
brettcannon: Fixed.

terryjreedy: agreed

committing.rst
brettcannon: Clarified.

committing.rst
```diff
@@ -201,25 +201,41 @@ when implemented, these changes *must* be noted in the "New Additions in
...
-it should be valid reStructuredText.  The "default role" (single backticks) can
-be used to refer to objects in the documentation.  Example NEWS entry::
+So a file name may be
+``Misc/NEWS.d/next/Library/2017-05-27.bpo-12345.Yl4gI2.rst``.
 
-   - bpo-15304: Fix warning message when `os.chdir()` fails inside
-     `test.support.temp_cwd()`.  Patch by Chris Jerdonek.
+A news file contents should be valid reStructuredText. An 80 character column
+width should be used. The "default role"
+(single backticks) in reST can be used to refer to objects in the documentation.
```
terryjreedy: Will they be ignored if there is no corresponding object *in the docs*?

committing.rst
brettcannon: I dropped the mention.

committing.rst
```diff
@@ -235,10 +251,10 @@ it is common to simply paste the NEWS entry into the commit message.  Here
 is an example::
 
    bpo-42: the spam module is now more spammy.
```
terryjreedy: This contradicts Larry's comment above about not starting with 'bpo' or issue number.

committing.rst
```diff
@@ -235,10 +251,10 @@ it is common to simply paste the NEWS entry into the commit message.  Here
...
-   
+
    The spam module sporadically came up short on spam. This change
    raises the amount of spam in the module by making it more spammy.
-   
+
```
terryjreedy: I have never added spaces before acknowledgements before and I believe same was generally true.  Unless the resulting News/changelog is formatted differently from before, blank lines might make it harder, not easier to read.

committing.rst
brettcannon: I believe it will open an editor, yes.

committing.rst
brettcannon: How so? Just randomly bang on your keyboard and keep 6 letters. 😄

committing.rst
brettcannon: This is part of reST markup for a block literal so it actually doesn't when viewed as rendered HTML.

committing.rst
brettcannon: If you mean the indent, that's once again because it's reST formatting for a literal block. If you mean the newline then I personally have always used it.

Regardless, I didn't change this format and so if you want to discuss the formatting of commit messages that belongs in a separate issue.

committing.rst
larryhastings: It opens $EDITOR.

If $EDITOR isn't set, it will try a sane default on some platforms ("notepad" on Windows, /etc/alternatives/editor on Linux (if present)).  I think when Ned got it working on macOS he said it shouldn't try to guess on that platform.

committing.rst
larryhastings: Or any number of letters, even zero.  The nonce is really just there as a final arbiter to prevent filename collisions.

committing.rst
terryjreedy: Notepad is definitely lightweight, and for a typical news item, should be fine.

willingc: If the editor is not set on macOS, it seems reasonable to default to TextEdit which is what the operating system does.

committing.rst
```diff
@@ -127,12 +127,12 @@ is worth it, ask on python-dev.
...
 Third, ensure the patch is attributed correctly with the contributor's
 name in ``Misc/ACKS`` if they aren't already there (and didn't add themselves
-in their patch) and by mentioning "Patch by <x>" in the ``Misc/NEWS`` entry
+in their patch) and by mentioning "Patch by <x>" in the ``Misc/NEWS.d`` entry
 and the checkin message. If the patch has been heavily modified then "Initial
 patch by <x>" is an appropriate alternate wording.
 
 If you omit correct attribution in the initial checkin, then update ``ACKS``
-and ``NEWS`` in a subsequent checkin (don't worry about trying to fix the
+and ``NEWS.d`` in a subsequent checkin (don't worry about trying to fix the
```
willingc: Generally accepted usage is "check-in" for a noun or adjective or "check in" for a verb. No need to update for this PR as it's "checkin" throughout the document. Future revisions should probably correct this.

brettcannon: You can't point that out to the son of a former school teacher who almost had a minor in English and expect them not to fix it immediately. 😉

committing.rst
brettcannon: I've now clarified this.

committing.rst
brettcannon: This level of discussion is more appropriate in the core-workflow issue tracker.

brettcannon:
Thanks for all the feedback, everyone!

PR #213, opened 2017-05-29
PEP 545: Add inada.naoki as Japanese translation coordinator.

JulienPalard:


terryjreedy:
Julien, I merged this for you because you are not recorded as being able to do it.

PR #216, opened 2017-06-05
Add make serve to makefile

mlouielu:
Borrow from python/cpython documentation, add `make serve` to serve devgudie at localhost (port 8000).

If the doc didn't build up, it will first build up then serve the devguide.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  check      to run a check for frequent markup errors"
+	@echo "  serve      to serve devguide on the localhost (8000)"
 
 clean:
 	-rm -rf $(BUILDDIR)/*
@@ -138,3 +139,6 @@ doctest: venv
 
 check:
 	$(PYTHON) tools/rstlint.py -i tools -i venv
+
+serve: html
+	tools/serve.py _build/html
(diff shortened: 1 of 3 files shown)


brettcannon:
Why do you want this? Since the devguide is entirely static and uses relative paths you can simply open up the files locally in the browser and have things render properly.

willingc:
Hi @mlouielu, My typical workflow is to `make` the docs content build for Sphinx. Next, I run `python3 -m http.server` to start a web server that will render things locally. Was there an advantage that you felt that `make serve` would have?

mlouielu:
@brettcannon you are right, but some additional function like `search` will need someone to serve them.

@willingc `python -m http.server` will need to enter `localhost:8000/_build/html` in the navigation bar, but `make serve` will cover this, so you will only need to enter `localhost:8000` to it.

willingc:
Thanks for clarifying @mlouielu. I'm fine with being consistent with the CPython doc build which uses `serve`. A few thoughts on the PR:

- Let's add a `make.bat` file modification for Windows such as found in CPython docs
- I believe that you should add `serve` to the `Makefile`'s `.PHONY` lines https://github.com/python/devguide/blob/master/Makefile#L17

matrixise:
@mlouielu in 3.7, I have added the support of a directory for `python -m http.server` with `-d ` for example: `python -m http.server -d _build/html/`

matrixise:
@mlouielu will you support the `serve` parameter on unix-like systems ?

matrixise:
@willingc I have asked to @mlouielu to add the hostname `localhost` in the output.

Now for my part, I have tested the target `serve` of `Makefile`, and it's a success.

Do you think we could merge it ?

willingc:
Sure thing. Thanks @mlouielu and @matrixise.

PR #218, opened 2017-06-06
Document how to apply patch from mercurial to git

### LINKED ISSUE #193: Document how to apply patch from mercurial to git
How I do it: 

1. In the CPython mercurial repo, re-create the patch using --git parameter: 
    hg diff --git > issueNNNN-git.patch

2. In the CPython git repo, create and switch to a new branch. Apply the patch into CPython git repository:
    git apply /path/to/issueNNNN-git.patch

3. Commit, push, and open the PR.

csabella:
Closes #193

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -203,6 +203,33 @@ from the `core-workflow  <https://github.com/python/core-workflow>`_
 repository to backport the commit.
 
 
+.. _git_from_mercurial:
+
+Applying a Patch from Mercurial to Git
+--------------------------------------
+
+Scenario:
+
+- A Mercurial patch exists but there is no pull request for it.
+
+Solution:
+
+1. Download the patch locally.
+
+2. Create a new branch and switch to it.
+
+3. For Mac and Unix::
+
+   $ patch -p1 < /path/to/issueNNNN.patch
+
+   For Windows::
+
+   $ git apply /path/to/issueNNNN-git.patch
+
+4. Commit, push, and open the pull request.
+
+
+
 .. _git_pr:
 
 Downloading Other's Patches
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -216,17 +216,30 @@ Solution:
 
 1. Download the patch locally.
 
-2. Create a new branch and switch to it.
+2. Apply the patch::
 
-3. For Mac and Unix::
+       $ git apply /path/to/issueNNNN-git.patch
+       
+   If there are errors, update to a revision from when the patch was
+   created and then try the ``git apply`` again::
 
-   $ patch -p1 < /path/to/issueNNNN.patch
+       $ git checkout `git rev-list -n 1 --before="yyyy-mm-dd hh:mm:ss" master`
+       $ git apply /path/to/issueNNNN-git.patch
+       
+   If the patch still won't apply, then a patch tool will not be able to
+   apply the patch and it will need to be re-implemented manually.
+       
+3. If the apply was successful, create a new branch and switch to it.
 
-   For Windows::
+4. Stage and commit the changes.
 
-   $ git apply /path/to/issueNNNN-git.patch
+5. If the patch was applied to an old revision, it needs to be updated and
+   merge conflicts need to be resolved::
 
-4. Commit, push, and open the pull request.
+       $ git rebase master
+       $ git mergetool
+
+6. Push the changes and open a pull request.
 
 
 


mlouielu:
Should this add some text to describe when the apply patch have the conflict?

Mariatta:
Thanks for the PR, @csabella . Since issue #193 can be closed with this PR, I edited the PR description with `Closes #193`. The linked issue will then be automatically be closed when the PR is merged.

csabella:
@mlouielu   I searched the devguide for the term 'conflict' and also looked at the ToC for an appropriate page, but I couldn't find a reference to resolving merge conflicts.  I would think that the process would be the same (or at least similar) regardless if the conflict comes from this process or from a PR on github, so maybe here isn't the best place to address it, but rather it would be better to link to a page just for that?

gitbootcamp.rst
```diff
@@ -203,6 +203,33 @@ from the `core-workflow  <https://github.com/python/core-workflow>`_
...
+
+Solution:
+
+1. Download the patch locally.
+
+2. Create a new branch and switch to it.
+
+3. For Mac and Unix::
+
+   $ patch -p1 < /path/to/issueNNNN.patch
```
matrixise: `git apply` works fine on Linux, in this case, we could use the same command for all the operating systems.

Mariatta: patch command was recommended by another core developer, so I'm inclined to leave it here.

ned-deily: I vaguely remember the discussion.  It could be that the other person wasn't as familiar with `git apply`?  I agree with @matrixise that it makes sense to recommend the git supplied command since (1) it should work for all supported platforms and (2) it has the potential to work better than plain `patch`; read `git help apply` for all the details.

Mariatta:
Merge conflict can arise in other situations, such as syncing with upstream or backporting.  It's not necessary to specifically mention it here.

gitbootcamp.rst
matrixise: in the past, we used `hg import --no-commit patch_url` and not `patch -p1` maybe we could use the same convention and use the `git apply` command and not just `patch`.

matrixise: and the doc is really useful https://git-scm.com/docs/git-apply

Mariatta: Alright I'm swayed :) Let's simplify this and document only `git apply`. Thanks.

mlouielu:
@csabella @Mariatta Just want to mention, I'm not sure if other places have this problem, but IDLE has changed it's filename before and cause some merge conflict in old patch, for example: [bpo-17942](http://bugs.python.org/issue17942). Also I think some old patch (if the file which apply was edit frequently) will not able to simply apply to git.

gitbootcamp.rst
Mariatta: I'm trying out both `git apply` and `patch` to apply the patch from this issue: http://bugs.python.org/issue12815

I downloaded PatchProxy.diff, and then ran `patch -p1 < PatchProxy.diff`. There are some conflicts, but at least I could apply the patch as is.

I was not able to do `git apply PatchProxy.diff` because the patch was not created using `--git` option.

Maybe this is why the `patch` command was recommended in the first place? 🤔 
If it's availble in the OS, it's more straightforward to use.

With `git apply` we first need to make sure that the patch was made with `--git` option.
Or am I missing something?

zware: I just tried the same: parts of the patch applied with `patch`, but *not* all of it.  `git apply` refused to apply a partial patch, nothing to do with the `--git` option.  The patch is quite old (hence the conflicts), so I checked out a revision from that era via:

```
git checkout `git rev-list -n 1 --before="2011-08-22 03:25:00" master`
```

After that, `git apply` happily applied the patch, albeit while complaining about a lot of trailing whitespace.

At this point, I could get the patch updated with something like:

```
git checkout -b issue_12815
git commit -a -m "Cover smptd"
git rebase master
git mergetool
```

matrixise: the `patch` command was recommended because this command already existed before `git apply`, in the past, I used the `patch` command but you need to add `-p 1` to the command and if you try with `-p 0` or without this parameter, the patch is not applied.

with `git apply`, we have to give the path file and sure that the patch has been generated with `hg diff --git` if we have used `hg`. now we use `git` instead of `hg`. just apply the patch with `git apply`

csabella: Looking at the git apply documentation, there is an option to check the patch first.
```
--check
Instead of applying the patch, see if the patch is applicable to the current working tree and/or the index file and detects errors. Turns off "apply".
```
Do you think it would be worthwhile to add that first and then proceed with the suggestion from @zware if the patch won't apply?

matrixise: yep, for me it's ok, we could wait for the answer of @zware .

csabella:
@mlouielu   I agree with the differences in applying IDLE patches.  I don't know if that's something that should be addressed specifically in a "Contributing to IDLE"-type document or if it applies to other modules and should be mentioned in the devguide.  I also don't know if there's a specific set of steps to follow for converting those patches.

ned-deily:
I would suggest treating the issue of rebasing older, pre-rename IDLE patches as a separate issue, consulting with @terryjreedy first.  If necessary, one could follow a process similar to what @zware suggests above, that is, trying to apply a patch first to an older version of the repo and then using git's rebase tools to try to update it.  But those kinds of problems aren't limited to IDLE patches.

terryjreedy:
The devguide should not have anything IDLE specific.  Discussing IDLE-specific strategies for using existing IDLE patches would be an appropriate topic for idle-dev list.  When it was just me involved, I planned to hand edit either the patch until it applied or the code until it had the desired behavior change.  Docstring additions, internal renamings, and other code changes will increasingly make the second course necessary.

For module names,  I have a list of old name - new name pairs.  It can be reformatted as needed,  and needs to be updated to account for the addition of query.py and deletion of other files.  If the old name only refers to a module, the replacement can be done blindly (macosxSupport -> macosx).  If the old name also refers to a class, a person must just whether to change or skip each occurrence (EditorWindow -> editor or not).  The list should be split at least in two.  Further discussion should continue elsewhere.

gitbootcamp.rst
zware: I don't think there's much need to specify doing `git apply --check` first; if the patch won't apply cleanly, git will exit non-zero and not change anything.

And I'll note again, patches need not have been created with `hg diff --git` for `git apply` to be able to apply them.

I think `git apply` is the most foolproof way to make sure you get the full patch applied.  Try it from the HEAD of master first; if it fails, try updating to a revision from when the patch was created and apply again.  If it won't apply at that point, I'd say the patch is suspect and you're probably better off just reimplementing it rather than trying to get a patch tool to accept it.

csabella: @matrixise  Thank you.  I've made the changes.

Mariatta:
Thanks everyone

Issue #217, opened 2017-06-05
Quoting for config alias.pr


terryjreedy:
The following in Bootcamp does not work on windows because the quoting is wrong for Windows:
git config --global alias.pr '!sh -c "git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}" -'
git responds with a usage message.

Windows only recognizes "..." quotes on a command line.  Reversing the quotes creates the alias.
git config --global alias.pr "!sh -c 'git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}' -"
Subsequently, "git pr ####" works because git finds the 'sh' included with Git for Windows.

If the reversed form works on *nix, the current line can just be edited.  If not, then the new line should be added with a 'for Windows' note.

Mariatta:
hmm I tried reversing the quotes on my machine (MacOS) and it doesn't work

```
$ git config --global alias.pr "!sh -c 'git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}' -"
-bash: !sh: event not found
```

So we need to add a note that says "on windows, reverse the quotes" something like that?

terryjreedy:
"On Windows, reverse the single (') and double (") quotes." should do it.

PR #219, opened 2017-06-07
Update git pr instruction

Mariatta:
On Windows, the single and double quotes should be reversed
Closes https://github.com/python/devguide/issues/217


Mariatta:
Thanks for the review :)

PR #222, opened 2017-06-08
Fix urls with the right redirect in the documentation

matrixise:
Just use `make linkcheck` on the devguide


brettcannon:
Thanks!

PR #227, opened 2017-06-12
Adjust Editing Contributor's PR section of the bootcamp

Mariatta:
The original instruction is to merge with `origin/master`.
I think it's more accurate to merge with `upstream/master`, where `upstream` has been referenced throughout the document as cpython's repo, and `origin` refers to our own fork.

If our `origin/master` is not up to date, we might not see the right result.
So I think it's better to merge with `upstream/master`.


ncoghlan:
LGTM.

I reworded the commit message though, as this actually arose from the fact that I wrote the original example based on the origin/pr remote setup.

Mariatta:
Awesome! Thanks :)

PR #235, opened 2017-06-25
Trivial: Fix broken link for Peephole Optimizer Paper

banool:
The old link for this paper is broken (or at least doesn't link to the paper in question). A quick google search revealed a few other places where the paper can be found. I opted for the link on legacy.python.org but any of the other links could be used if they better suit the documentation style.

I've run `make html` and it seems to have worked fine 😄

### DIFF
--- a/compiler.rst
+++ b/compiler.rst
@@ -539,7 +539,7 @@ References
    http://www.cs.princeton.edu/research/techreps/TR-554-97
 
 .. [#skip-peephole] Skip Montanaro's Peephole Optimizer Paper
-   (http://legacy.python.org/workshops/1998-11/proceedings/papers/montanaro/montanaro.html)
+   (https://drive.google.com/open?id=0B2InO7qBBGRXQXlDM3FVdWZxQWc)
 
 .. [#Bytecodehacks] Bytecodehacks Project
    (http://bytecodehacks.sourceforge.net/bch-docs/bch/index.html)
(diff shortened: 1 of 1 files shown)


serhiy-storchaka:
I'm not sure that legacy.python.org will not disappear one time. Maybe @smontanaro have more stable link to his paper?

smontanaro:
I took down www.smontanaro.net. If you like (and want to avoid adding legacy.python.org references), I can put a copy up on Google Drive or Dropbox, or just send you a copy you can drop into place on the website itself.

brettcannon:
A link to Drive or Dropbox would also work.

serhiy-storchaka:
We could also just move the paper to non-legacy python.org.

smontanaro:
When I get home this evening, I'll toss it up on Google Drive. From there
you can copy it (or not) wherever you think best.

smontanaro:
Here it is. Let me know if you can't view it:

https://drive.google.com/open?id=0B2InO7qBBGRXQXlDM3FVdWZxQWc

Skip

brettcannon:
@smontanaro I can view it (it's raw HTML instead of rendered HTML, but since this is a historical document I think that's fine).

smontanaro:
I hope it's fine. I'm not going to go back and mess with it at this point.
:-)

brettcannon:
Thanks everyone for helping out!

PR #245, opened 2017-07-25
Add the release managers for the current branches of Python

matrixise:


matrixise:
@ned-deily I understand,

but I am going to explain the reason of this PR.

I wanted to know the RM for two releases and from the index page, this info is not visible, that's why I have added the RM column in the table.

I close it.

Issue #238, opened 2017-07-21
Clarify that backports should start from the squashed-and-merged commit


ncoghlan:
After pointing a new contributor to https://docs.python.org/devguide/gitbootcamp.html#backporting-merged-changes to handle their own backport PRs, I discovered that it's currently taking a step for granted: finding the relevant commit to backport from the CPython master branch, rather than squashing your original PR branch locally.

Since we specifically want to use the squashed-and-merged commit so that the backports include the corresponding commit message, we should make that explicit here.

ncoghlan:
Some initial thoughts on how we might improve this:

1. Move https://docs.python.org/devguide/gitbootcamp.html#backporting-merged-changes to be directly after https://docs.python.org/devguide/gitbootcamp.html#accepting-and-merging-a-pull-request
2. Add a note suggesting the following pair of commands:

    git fetch upstream
    git rev-parse ":/bpo-12345"

3. Mention that the relevant commit hash can also be found by looking at the merge notification on the original PR (e.g. https://github.com/python/cpython/pull/2786#event-1173237073 )

Mariatta:
There's a write up about which commit hash to use in [cherry_picker](https://github.com/python/core-workflow/blob/master/cherry_picker/readme.rst#id4) It can be copied over here.

terryjreedy:
I presume 'bpo-1235' is a stand-in for the fork branch the PR is based on.  Once the merge is done, it disappears from the PR (as the PR can no longer be added to using the branch name ;-).  Copying the commit hash available after the PR is merged seems best.  I try to remember to grab it before I start typing the cherry_picker command.  (Which is working quite well for simple backports.)

ncoghlan:
@terryjreedy `git rev-parse ":/bpo-12345"` searches for a commit *message* containing the given string (rather than a branch name) and returns the commit hash(es).

NicholasKobald:
I found the link Mariatta mentioned to be very useful. The reason I ended up doing it the way I did despite seeing that reference, was because I got errors trying to use the final merge commit hash, since I had no correctly set python/cpython to be upstream of my forked repository (and so the hash gave me a `fatal: bad object ..` error.) 

Because of running into that error and not being sure at first how to fix it, I suggest adding the following as well:

Insure the cpython repository is upstream of your forked repository.
```
git remote add upstream https://github.com/python/cpython.git
```
Then, pull the most recent commits to your local machine.
```
git fetch upstream
git rev-parse ":/bpo-12345"
```

[...]

PR #254, opened 2017-08-28
Clarify which commit hash is used for backporting

Mariatta:
Move ``Backporting Merged Changes`` section to right after the ``Accepting and Merging a Pull Request``

Closes https://github.com/python/devguide/issues/238

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -312,6 +312,14 @@ page.  Find the event that says something like::
 
 By following the link to ``<commit_sha1>``, you will get the full commit hash.
 
+Alternatively, the commit hash can also be obtained by the following git commands::
+
+   $ git fetch upstream
+   $ git rev-parse ":/bpo-12345"
+
+The above commands will print out the hash of the commit containing ``"bpo-12345"``
+as part of the commit message.
+
 
 Editing a Pull Request Prior to Merging
 ---------------------------------------
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -199,18 +199,6 @@ Solution::
    $ git push --force origin some-branch
 
 
-Backporting Merged Changes
---------------------------
-
-A pull request may need to be backported into one of the maintenance branches
-after it has been accepted and merged into ``master``.  It is usually indicated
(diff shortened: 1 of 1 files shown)


gitbootcamp.rst
```diff
@@ -305,6 +293,26 @@ Pull requests can be accepted and merged by a Python Core Developer.
...
+after it has been accepted and merged into ``master``.  It is usually indicated
+by the label ``needs backport to X.Y`` on the pull request itself.
+
+Use the utility script `cherry_picker.py <https://github.com/python/core-workflow/tree/master/cherry_picker>`_
+from the `core-workflow  <https://github.com/python/core-workflow>`_
+repository to backport the commit.
+
+The commit hash for backporting is the squashed commit that was merged to
+the ``master`` branch.  On the merged pull request, scroll to the bottom of the
+page.  Find the event that says something like::
```
methane: When I resolved conflict while cherry-picking to 3.6 branch, I want to use squash commit of 3.6 branch
for backporting to 3.5.

Mariatta: When backporting the squashed commit from `master`, we get to keep the reference to the original PR on `master`, so bedevere-bot can remove the `needs backport to ..` labels.

Does this happen when backporting from 3.6 to 3.5?

ncoghlan: It should - the trick in this case is that the "needs backport to 3.5" label should go on the *backport* PR against the 3.6 branch, rather than the original PR against master.

ncoghlan: Right, the only time it makes sense if when the backport from master *fails*, and even then, it only applies for security fixes, or the times when we have two full maintenance branches open (i.e. from beta 1 of the upcoming stable through to the release of that branch and the subsequent final full maintenance release of the now previous stable branch).

However, now that we branch at beta, we're likely to spend around 1/3rd of the time in that state.

ncoghlan: @Mariatta Clarifying: I don't think the above needs to be covered in this PR, but there *are* cases where we want to do X.Y -> X.Y-1 backports, rather than backporting directly from master.

Mariatta: Thanks :) I'm going to merge this.

Issue #267, opened 2017-09-27
Cannot build devguide in virtual environment on Windows


jeff5:
On Windows `make html` does nothing.

There are several problems here:
1. There's a use of `goto` inside parenthes. Yeah, it ought to work but [it doesn't reliably](https://ss64.com/nt/syntax-brackets.html). In fact, it stops the action dead.
2. A virtual environment is not set up for you as claimed in the guide.
3. The python command used is py -3 which ignored the active venv I created. (Not sure why, but it's not how the Unix Makefile works anyway.) 

I think that with problem 1 fixed, one could build by installing Sphinx site-wide, but I'd like to follow the sensible advice in [section 6.3](https://devguide.python.org/docquality/#helping-with-the-developer-s-guide), and use a virtual environment.

I'll propose you a fix in due course, as much like the Unix Makefile as I can.

zware:
Have a look at what [Doc/make.bat](https://github.com/python/cpython/blob/master/Doc/make.bat) does, I think it's in better shape than this make.bat is at this point.

jeff5:
@zware Thanks. There are some interesting things there, but not the use of a virtual environment.

I'm aiming for as much like the Makefile as possible and otherwise minimum change. (`.bat`s are so fragile.)

I take the hint from [this conditional](https://github.com/python/cpython/blob/db50ba7c72bd5f0617c50df3f1a1bb8b26955882/Doc/make.bat#L9) however that if you've defined your own command in SPHINXBUILD we should butt out with our venv.

zware:
Ah, you're right; we haven't added venv support to Doc/make.bat.  Sorry for the red herring :(

Yes, if SPHINXBUILD is set, we should assume the user knows what they're doing and just go with it.  Otherwise, I like your plan of trying to match Makefile's behavior as closely as possible.

jeff5:
Resolved in #268.

PR #268, opened 2017-09-28
Build in a virtual environment on Windows (#267)

jeff5:
Add to the Windows build script a clause to create and use a virtual
environment and install Sphinx there, comparable with the Unix Makefile.
The user may suppress this behaviour by defining a her preferred command
in the environment variable SPHINXBUILD, as before.

### DIFF
--- a/docquality.rst
+++ b/docquality.rst
@@ -86,12 +86,31 @@ To build the devguide, some additional dependencies are required (most
 importantly, `Sphinx`_), and the standard way to install dependencies in
 Python projects is to create a virtualenv, and then install dependencies from
 a ``requirements.txt`` file. For your convenience, this is all *automated for
-you* and all you have to do to build the devguide is run::
+you*.
+
+On a Unix-like system all you have to do to build the devguide is run::
 
     $ make html
 
 in the checkout directory, which will write the files to the ``_build/html``
-directory.  Note that ``make check`` is automatically run when
+directory.
+
+On Windows the command is:
+
+..  code-block:: doscon
+
+    > .\make html
+
+in the checkout directory, and again this will write the files to the
+``_build\html`` directory.
+If you are running PowerShell, you may have to enable unsigned local scripts to
+run, with the command (used once only):
+
+..  code-block:: ps1con
+
+    PS> Set-ExecutionPolicy -Scope CurrentUser RemoteSigned
+
+Note that ``make check`` is automatically run when
 you submit a :doc:`pull request <pullrequest>`, so you should make
 sure that it runs without errors.
 
--- a/docquality.rst
+++ b/docquality.rst
@@ -86,30 +86,17 @@ To build the devguide, some additional dependencies are required (most
 importantly, `Sphinx`_), and the standard way to install dependencies in
 Python projects is to create a virtualenv, and then install dependencies from
 a ``requirements.txt`` file. For your convenience, this is all *automated for
-you*.
-
-On a Unix-like system all you have to do to build the devguide is run::
+you*. To build the devguide on a Unix-like system use::
 
     $ make html
 
-in the checkout directory, which will write the files to the ``_build/html``
-directory.
-
-On Windows the command is:
+in the checkout directory. On Windows use:
 
 ..  code-block:: doscon
 
     > .\make html
 
-in the checkout directory, and again this will write the files to the
-``_build\html`` directory.
-If you are running PowerShell, you may have to enable unsigned local scripts to
-run, with the command (used once only):
-
-..  code-block:: ps1con
-
-    PS> Set-ExecutionPolicy -Scope CurrentUser RemoteSigned
-
+You will find the generated files in ``_build/html``.
 Note that ``make check`` is automatically run when
 you submit a :doc:`pull request <pullrequest>`, so you should make
 sure that it runs without errors.
--- a/make.bat
+++ b/make.bat
@@ -47,7 +47,9 @@ if "%1" == "clean" (
 	goto end
 )
 
-rem If we get this far, we're going to use the SPHINXBUILD command
+rem Targets other than "clean", "check", "serve", "help", or "" need the
+rem Sphinx build command, which the user may define via SPHINXBUILD.
+
 if not defined SPHINXBUILD (
 	rem If it is not defined, we build in a virtual environment
 	if not exist venv (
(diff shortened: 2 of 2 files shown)


make.bat
```diff
@@ -39,14 +39,27 @@ if "%1" == "help" (
...
-)
+:help0
 
 if "%1" == "clean" (
 	for /d %%i in (%BUILDDIR%\*) do rmdir /q /s %%i
 	del /q /s %BUILDDIR%\*
 	goto end
 )
 
+rem If we get this far, we're going to use the SPHINXBUILD command
```
ezio-melotti: Unless we got a "clean" "check", "serve", "help", or "" we should get here, right?

make.bat
```diff
@@ -183,3 +196,4 @@ cmd /C %PYTHON% tools\serve.py %BUILDDIR%\html
 goto end
 
 :end
+endlocal
```
ezio-melotti: I think you are missing a newline at the end here.

docquality.rst
```diff
@@ -86,12 +86,31 @@ To build the devguide, some additional dependencies are required (most
...
+
+in the checkout directory, and again this will write the files to the
+``_build\html`` directory.
+If you are running PowerShell, you may have to enable unsigned local scripts to
+run, with the command (used once only):
+
+..  code-block:: ps1con
+
+    PS> Set-ExecutionPolicy -Scope CurrentUser RemoteSigned
+
```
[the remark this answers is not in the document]
jeff5: Ok.

In the "Creator's Update" to Windows 10, a few months back, Microsoft took away convenient access to cmd from the File explorer (which is how I'd normally start my day), leaving only PS. I took the hint and rewrote my scripts. It _is_ better IMO and should be considered the new default. I can't speak for uptake either, but yes, people have probably figured out the permissions thing.

make.bat
jeff5: Yes. The venv bit is here rather than up top for that reason. Perhaps it was someone's logic to order the targets so in anticipation. Happy to assume so.

make.bat
```diff
@@ -39,14 +39,27 @@ if "%1" == "help" (
...
 if "%1" == "clean" (
 	for /d %%i in (%BUILDDIR%\*) do rmdir /q /s %%i
 	del /q /s %BUILDDIR%\*
 	goto end
 )
 
+rem If we get this far, we're going to use the SPHINXBUILD command
+if not defined SPHINXBUILD (
+	rem If it is not defined, we build in a virtual environment
+	if not exist venv (
```
zware: This needs some protection against running the script from somewhere other than the dir holding the script. I think `if not exist %~dp0venv` should do it, but make sure I remembered the correct flags.

docquality.rst
```diff
@@ -86,12 +86,31 @@ To build the devguide, some additional dependencies are required (most
...
+directory.
+
+On Windows the command is:
+
+..  code-block:: doscon
+
+    > .\make html
+
+in the checkout directory, and again this will write the files to the
+``_build\html`` directory.
```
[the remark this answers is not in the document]
zware: The `.\` shouldn't even be necessary unless that's a PowerShell-ism; the same command should work on both platforms.

docquality.rst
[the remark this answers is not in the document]
zware: I agree that this probably doesn't need to be here.

By the way, right click the start button for an easy shortcut to Command Prompt (in regular and admin flavors), unless they also removed that in which case I'm never updating my VM again :)

make.bat
jeff5: Sorry, missed this comment the first look. I'll think about this. First thought is that it didn't have it before.

Second is that it might be useful and another thing I could take from the CPython docs build. I guess it would pushd to the right place ... https://github.com/python/cpython/blob/8d59aca4a953b097a9b02b0ecafef840e4ac5855/Doc/make.bat#L4

jeff5: To clarify, I think the idiom in that example would make it run correctly whatever the current directory, but I'll check that (and figure out whether `popd` is needed).

jeff5:
@zware The `.\make` is indeed a concession to PShell, which requires it. It works with CMD too. I think CMD users will know it is unnecessary for them to type. Without this I'd be reverting `docquality.rst` to where I started (except for Ezio's side-improvements)!

jeff5:
Although still showing "changes requested" I believe the review comments are now answered, mostly by change. (Not sure if I've subverted the workflow somehow.) Anyhow, is this good enough to land this now?

make.bat
```diff
@@ -39,14 +41,29 @@ if "%1" == "help" (
 	echo.  check
 	echo.  serve      to serve devguide on the localhost (8000)
 	goto end
-)
+:help0
```
ezio-melotti: I'm not particularly familiar with bat files, and I'm not sure I understand this change.
ISTM that you just avoid printing the help if the arg is not "help", so the end result should be the same as before.
FWIW all the other commands follow the structure:
```
if "%1" == "something" (
    ...
    goto end
)
```
except "check" and "serve".
If possible, to maintain symmetry and improve readability, I would keep/adopt the same structure for all the different targets, and combine with an OR the two targets that result in an help ("help" and "").  This could be done in a separate PR, but unless there's a reason to break this structure, I would either revert or refactor this change.

(I made this comment on the first review, but apparently forgot to save/add it, so I'm adding it back now.)

jeff5: I'm not great with `.bat` files either, but as briefly stated in the related issue #267, the `.\make html` currently builds nothing because `goto end` is executed unconditionally. I attributed this to known issues with parenthetical `if` bodies and `goto`, anr the legacy-style work-around made it go away.

On closer inspection, the problem is the un-escaped `(8000)`. I can fix that.

There is no OR operator, but I can make it a bit clearer (closer to what `sphinx-quickstart` generated).

ezio-melotti: If there's a valid reason to do it that way, then go ahead, but perhaps add a comment to document it :)

jeff5: Thanks to your challenge I have found the real problem and the right answer.

ezio-melotti:
Merged, thanks for the PR!

jeff5:
Thanks for the helpful reviews.

You may have spotted that my contributions are a side-effect of trying to re-use your work, or follow its model, in a devguide for Jython. I'll offer further things as I spot them.

PR #269, opened 2017-09-30
Fix whitespace and quoting issues identified by "make check"

jeff5:
This fixes 11 of 12 complaints by make check, but there is still one I don't know what to do with:
```
PS devguide> .\make check
[2] .\gitbootcamp.rst:282: default role used
1 problem with severity 2 found.
```
I think this last is a false alarm, but there doesn't seem to be any escaping it short of change to rstlint.py to recognise literal blocks. The offending item is the command nested with backticks in:
```
$ git checkout `git rev-list -n 1 --before="yyyy-mm-dd hh:mm:ss" master`
```

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -240,7 +240,7 @@ to the pull request.
 
 After the pull request has been merged, it can be backported using cherry_picker.py_.
 
-The commit hash can be obtained from the original pull request, or by using `git log`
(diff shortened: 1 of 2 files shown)


zware:
I agree that the last one is a false-positive; we'll fix that in a separate PR (feel free to write that PR if you'd like :))

PR #274, opened 2017-10-08
Document an example to show how to use git worktrees

SurenNihalani:
First attempt for #105


setup.rst
```diff
@@ -92,6 +92,9 @@ i.e., a version in :ref:`maintenance mode <maintbranch>`, you can checkout
 a release branch. For instance, to checkout a working copy of Python 3.5,
 do ``git checkout 3.5``.
 
+If you want the ability to work on multiple checkouts of python, you can use git 
+worktrees. See :ref:`committing`'s worktree section for an example.
```
Mariatta: Two ways to reference the `git_worktrees` section:

```
:ref:`git_worktrees`
```
or
```
:ref: Git worktree section <git_worktrees>
```

Build the doc to see the difference between the two.

SurenNihalani: Perfect. Thanks for the help and the review! Updated the code review and verified the build: 
image

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
 
 .. _cherry_picker.py: https://github.com/python/core-workflow/tree/master/cherry_picker
 
+.. _git_worktrees:
+
+Speeding up development workflow using git worktrees
+----------------------------------------------------
+
+When backporting changes to multiple versions, it's often useful to have more than one checkouts
```
JelleZijlstra: more than one *checkout*

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
+of the codebase. This way you can do things like running tests on one version
```
JelleZijlstra: comma after "This way"

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
+    Preparing /home/user/cpython_2.7 (identifier cpython_2.7)
+    HEAD is now at e37fefba8b Revert the last change:
+
+    $ cd ~/cpython_2.7
+
+    # make changes 
+    
+    $ git commit 
+
+It's important to know that a branch can only be checked out in one of the worktrees unless you
```
JelleZijlstra: Remove "It's important to know that". If it's not important, you wouldn't have included it, would you?

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
+    HEAD is now at e37fefba8b Revert the last change:
+
+    $ cd ~/cpython_2.7
+
+    # make changes 
+    
+    $ git commit 
+
+It's important to know that a branch can only be checked out in one of the worktrees unless you 
+use ``--force`` option with git checkout or git worktree add.
```
JelleZijlstra: Add "the" before `--force` and quote the other two git commands in backticks.

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
 .. _cherry_picker.py: https://github.com/python/core-workflow/tree/master/cherry_picker
 
+.. _git_worktrees:
+
+Speeding up development workflow using git worktrees
+----------------------------------------------------
+
+When backporting changes to multiple versions, it's often useful to have more than one checkouts
+of the codebase. This way you can do things like running tests on one version
+while committing a patch on another version. For example: ::
```
ezio-melotti: This paragraph seems to have lines that are too long.

I'd prefer to add two spaces after the full stops.

You can just use ``For example::`` -- it will automatically add a ``:`` and format the next block of code.

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
+
+    $ git worktree add ~/cpython_2.7 origin/2.7                                 
+    Preparing /home/user/cpython_2.7 (identifier cpython_2.7)
+    HEAD is now at e37fefba8b Revert the last change:
+
+    $ cd ~/cpython_2.7
+
+    # make changes 
+    
+    $ git commit
```
ezio-melotti: Unless the default highlighting for the devguide is ``shell``, you should specify it.  (You should be able to tell by checking the highlighting of the built docs).

I would remove the empty lines and add more comments: for example what does ``git worktree add ~/cpython_3.6 origin/3.6``?  Something as simple as ``$ # create a new clone of origin/3.6 in the ~/cpython_3.6 dir`` is enough.

A minor nit, but I would prefer if the dir was simply called ``3.6``, IIRC that's what we used in the old devguide and what I've been personally used, since it's shorter and matches the branch name.

committing.rst
```diff
@@ -258,6 +258,41 @@ Developers can apply labels to GitHub pull requests).
...
+    HEAD is now at e37fefba8b Revert the last change:
+
+    $ cd ~/cpython_2.7
+
+    # make changes 
+    
+    $ git commit 
+
+A branch can only be checked out in one of the worktrees unless you 
+use the ``--force`` option with ``git checkout`` or ``git worktree add``.
```
ezio-melotti: This could be turned around:  "If you want to checkout a branch in multiple worktrees, you have to use the ``--force`` option with ``git checkout`` or ``git worktree add``."

setup.rst
```diff
@@ -92,6 +92,9 @@ i.e., a version in :ref:`maintenance mode <maintbranch>`, you can checkout
 a release branch. For instance, to checkout a working copy of Python 3.5,
 do ``git checkout 3.5``.
 
+If you want the ability to work on multiple checkouts of python, you can use git
```
willingc: Python

Mariatta:
@SurenNihalani Could you make the requested changes?
Thanks.

SurenNihalani:
Ack. will address this by end of the day.

committing.rst
SurenNihalani: couldn't really do shell because the lexer couldn't parse all the standard output. lmk if you think we should still do it.

SurenNihalani:
I have changed the example to use upstream instead. lmk if i need to make other changes.

committing.rst
```diff
@@ -258,6 +258,35 @@ Developers can apply labels to GitHub pull requests).
...
 
+.. _git_worktrees:
+
+Speeding up development workflow using git worktrees
+----------------------------------------------------
+It's often useful to have more than one checkout of the codebase (for testing on one version + 
+developing on another concurrently). For example::
+
+    # create a new clone of upstream/3.6 in ~/cpython_3.6 dir
+    $ git worktree add ~/3.6 upstream/3.6
```
merwok: Can you make the current working directory explicit?

terryjreedy: The '~' here and elsewhere presumes that one put the original clone in ones home directory, which I suspect is unusual.  Just use '../3.6' to put the worktree beside the clone, as one must be in the original clone for the git command to work.

I just checked, and '~/whatever' does not ever work on windows.  I am still adamantly opposed to making the devguide linux specific, including the $ crap.  We really did agree last spring to not do that.

Mariatta:
So I don't really know how git worktree works, so I can't comment on that.

If the `$` signs were to be removed from the code examples, is there a way to distinguish between a command, and the output?

SurenNihalani:
I don't think so. I am open to other solutions.

merwok:
If we stop using the `$ ` prompt convention, the only way would be to have a code block for every command and one for every output.  This could break the flow of reading and writing :(

I agree that the devguide and docs in general should give instructions for unix systems and windows, but I don’t really get the hostility toward `$ `.  It’s a notation convention that matches Python’s own `>>> ` that’s used in the docs too.

methane:
And many people used bash on Windows when using git CLI.

committing.rst
```diff
@@ -258,6 +258,35 @@ Developers can apply labels to GitHub pull requests).
...
+    $ git worktree add ../3.6 upstream/3.6   
+    Preparing ../3.6 (identifier 3.6)
+    HEAD is now at 49731ef8d2 dict: add iterkeys aliases in 3.5
+    $ cd ../3.6;  # go to the checkout
+    $ git cherry-pick 686c146b1ca1eb5c7ef058bb28772ac612a8ec1d  # add a patch and commit the result
+    [bpo-31620 c48137b107] bpo-31620: have asyncio/queues not leak memory when you've exceptions during waiting
+    Date: Fri Sep 29 01:03:19 2017 +0000
+    files changed, 25 insertions(+)
+    create mode 100644 Misc/NEWS.d/next/Library/2017-10-06-04-35-31.bpo-31620.gksLA1.rst
+    $ ./python -m test -j3  # run tests on the new patch
```
terryjreedy: This is *nix specific and does not work on Windows.

`> .\python  -m test -j3` will work, but as far as I know, the `.\` is unneeded as `python` in the repository directory overrides `python` on the path.  So either drop `./` or add
`> python -m test -j3` for the Windows version of the command.

terryjreedy:
My hostility is more toward only presenting *nix-specific versions of commands that do not work on Windows (and the sometimes dismissal of Windows users).  I found another such command in the current patch to fix.

As I remember, the decision last spring was to not use the system-specific prompts except when giving two system-specific versions of a command.  That convention could be used in the case I flagged.

Eric, what 'docs' are you referring to?   *Python Setup and Usage* puts all commands in code boxes and only adds '$' in one example in the Unix section.   If '$' is used generically elsewhere, it might date from when Python was mostly unix-only.

merwok:
If your quote of «docs» is from «It’s a notation convention that matches Python’s own `>>>` that’s used in the docs too.», I meant that we have a lot of code blocks in the `pycon` format (python console) like this:

```python
>>> entry
output
```

which can help readers not already familiar with the unix prompts / output code blocks understand them.

committing.rst
JulienPalard: Dropping "./" on *nix won't work (will hit the python from the PATH, not from the repository).

JulienPalard: The documentation [about running tests](https://devguide.python.org/runtests/) specifies 

> You may need to change this command as follows throughout this section. On most Mac OS X systems, replace ./python with ./python.exe. On Windows, use python.bat. If using Python 2.7, replace test with test.regrtest.

and then use a lot of `./python`.

I we do not want to create two example blocs maybe use a paragraphe like it?

willingc:
Hi @SurenNihalani, Thanks for taking the time and putting in the effort to create this PR. :cake:

I'm wondering if at this point it might be best to close this PR or scale back the PR so that the `setup.rst` file mentions reading the official git documentation (https://git-scm.com/docs/git-worktree) and remove changes to `committing.rst`.

Please let me know which way you would like to proceed. Thanks.

willingc:
Thanks @SurenNihalani. I appreciate your responsiveness.

Issue #191, opened 2017-05-07
Clarify policy of personal branches


brettcannon:
https://cpython-devguide.readthedocs.io/committing.html#working-with-git says you shouldn't push personal branches to the cpython repo, which is accurate. But an exception should be mentioned for when you make a quick edit through the web UI, in which case the branch can exist for less than 24 hours.

Mariatta:
Done in https://github.com/python/devguide/pull/276

PR #276, opened 2017-10-08
Add exception regarding personal branches

pablogsal:
This Pull Request implements the changes proposed on #191. A very simple statement was added to the dev guide on my understanding of the proposed changes. If I need to add a more detailed explanation on when this is allowed or how to delete the branches (or anything else) I will be more than happy to change it. 😄


ezio-melotti:
Merged, thanks for the PR!

PR #278, opened 2017-10-12
Don't recommend single backticks in news files

berkerpeksag:
* It's not a best practice
* It doesn't make news files more readable
* It's easy to confuse its syntax with Markdown
* And it adds unnecessary steps to contributing process.
  Contributors shouldn't have to remember which style should be
  used in news files and the rest of the documentation.
* Sphinx's default_role setting defaults to None
  (http://www.sphinx-doc.org/en/stable/config.html#confval-default_role)
  and neither Doc/conf.py nor blurb set it explicitly so the meaning of
  default_role can change in future releases of Sphinx.

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -178,18 +178,13 @@ standard library). The file name itself should be of the format
 So a file name may be
 ``Misc/NEWS.d/next/Library/2017-05-27.bpo-12345.Yl4gI2.rst``.
 
-The contents of a news file should be valid reStructuredText. The "default role"
-(single backticks) in reST can be used to refer to objects in the documentation.
-An 80 character column width should be used. There is no indentation or leading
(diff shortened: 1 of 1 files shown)


berkerpeksag:
I found `default-role` directive at https://github.com/python/cpython/blob/277c84067ff5dfa271725ee9da1a9d75a7c0bcd8/Doc/tools/extensions/pyspecific.py#L267

ezio-melotti:
What happens to older news entries that use single backticks?
Perhaps blurb can detect/convert single to double backticks, so that the end result will be consistent even if people use single backticks?

brettcannon:
I don't think we need to change how things work now, just stop documenting that backticks are usable.

willingc:
Thanks @berkerpeksag

PR #282, opened 2017-10-12
Limit line length to 80 & clean-up nearby formatting.

jeff5:
Primary aim is to conform to the 80 character rule (where possible). At the same time I've cleaned up other formatting faults in the same area as the long lines.

I don't personally like to re-flow paragraphs further than end of the affected sentence when I'm writing, so I mostly haven't. It makes it easier to see what changed.

### DIFF
--- a/documenting.rst
+++ b/documenting.rst
@@ -457,7 +457,9 @@ the extension mechanisms of reST, and Sphinx makes heavy use of it.
 
 Basically, a directive consists of a name, arguments, options and content. (Keep
 this terminology in mind, it is used in the next chapter describing custom
-directives.)  Looking at this example,::
+directives.)  Looking at this example,
+
+::
 
    .. function:: foo(x)
                  foo(y, z)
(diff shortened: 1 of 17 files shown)


documenting.rst
```diff
@@ -456,7 +457,7 @@ the extension mechanisms of reST, and Sphinx makes heavy use of it.
 
 Basically, a directive consists of a name, arguments, options and content. (Keep
 this terminology in mind, it is used in the next chapter describing custom
-directives.)  Looking at this example, ::
+directives.)  Looking at this example,::
```
ezio-melotti: Isn't the space necessary here?
You could remove the comma and use ``example::``, but otherwise I would leave it as is.

jeff5: That's a bit of Eclipse formatting (good for some things), not intentional. I see that in the original, the comma shows and the colon is invisible. I'll replace with the long form.

PR #284, opened 2017-10-14
Add entry to developer log.

jeff5:
Corrects ancient oversight.


jeff5:
This feels a little self serving, but it keeps the record straight. The text is from @fwierzbicki , and I've chosen to insert it where it belongs chronologically, rather than at the top as specified, as that seems less confusing to readers.

I've been in https://hg.python.org/committers.txt already from the start.

ncoghlan:
We're usually going to know our own history better than anyone else, so submitting corrections for errors and omissions makes sense to me! :)

PR #285, opened 2017-10-14
Demote "Working with Git" within committing.rst

jeff5:
Makes committing.rst consistent with the rest of the guide in having
just one top-level section. This eliminates a surprise navigation where
18 comes next after 16.

In the file history, this section (as "Working with Mercuruial") grew too large and was promoted in-place (commit e8b299e9), I infer, preparatory to a move that never came. Then it shrank again as sub-sections went to many other places in the guide.

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -193,7 +193,7 @@ unprocessed.)
 
 
 Working with Git_
-=================
+-----------------
 
 .. seealso::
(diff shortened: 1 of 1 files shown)


ncoghlan:
Makes sense to me, thanks!

Issue #292, opened 2017-11-01
Add linkcheck to CI


levlaz:
I noticed a TODO in the [.travis.yml](https://github.com/python/devguide/blob/master/.travis.yml#L8) file to add linkcheck to CI. 

This is simple enough to do, is there a historical reason why it has not been done?

zware:
I suspect it's just the fact that it would be failing, or take significant time to run.  It would be worth a PR to turn it on and see what it does, but there's no guarantee that we'd merge it yet.

levlaz:
It might be useful to run the report on each build (and not cause the build to "fail") and save it as an artifact for handy reference. 

It seems that in order to use artifacts on travis you need to [upload them to S3 manually](https://docs.travis-ci.com/user/uploading-artifacts/). Which seems not ideal. 

Would you be opposed to switching to CircleCI? They host artifacts for free. 

Disclaimer: I work at CircleCI. :)

willingc:
Next action: Enable linkcheck on travis as a test. Mark this test as Failure OK so that it does not prevent merges yet provides information to contributors.

hugovk:
Please see PR #369.

PR #293, opened 2017-11-02
PoC for using CircleCI

levlaz:
This is what running the build on CircleCI might look like. If you look at my fork there are a couple cool things. 

https://circleci.com/gh/levlaz/devguide/3

1. The entire build (including linkcheck) ran in under 2 minutes. 
2. We are able to store the build output as artifacts. 

screen shot 2017-11-01 at 7 23 02 pm

This allows us to make a nice report for "broken" things in linkcheck here: https://3-108794746-gh.circle-artifacts.com/0/home/circleci/project/_build/linkcheck/broken.txt

And we can see the entire build HTML which could be useful for previewing the state of the project at a specific commit: https://3-108794746-gh.circle-artifacts.com/0/home/circleci/project/_build/html/index.html

*note: linkcheck does not fail the build even though it has failures because the command that is run is `make linkcheck || true` which forces an exit code of 0. 

fix https://github.com/python/devguide/issues/292

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,43 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.6
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: dependency-cache-{{ checksum "requirements.txt" }}
+
+      - run:
+          name: install python dependencies
+          command: |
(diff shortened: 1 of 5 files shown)


levlaz:
Sorry, closing because I screwed up my git history and made the branch from another branch.

PR #294, opened 2017-11-02
CircleCI PoC

levlaz:
This is what running the build on CircleCI might look like. If you look at my fork there are a couple cool things. 

https://circleci.com/gh/levlaz/devguide/5

1. The entire build (including linkcheck) ran in under 1 minute. 
2. We are able to store the build output as artifacts. 

screen shot 2017-11-01 at 7 23 02 pm

This allows us to make a nice report for "broken" things in linkcheck here: https://3-108794746-gh.circle-artifacts.com/0/home/circleci/project/_build/linkcheck/broken.txt

And we can see the entire build HTML which could be useful for previewing the state of the project at a specific commit: https://3-108794746-gh.circle-artifacts.com/0/home/circleci/project/_build/html/index.html

*note: linkcheck does not fail the build even though it has failures because the command that is run is `make linkcheck || true` which forces an exit code of 0. 

fix https://github.com/python/devguide/issues/292

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,45 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.6
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: dependency-cache-{{ checksum "requirements.txt" }}
+
+      - run:
+          name: install python dependencies
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r requirements.txt
+
+      - save_cache:
+          key: dependency-cache-{{ checksum "requirements.txt" }}
+          paths:
+            - "venv"
+
+      - run:
+          name: build sphinx
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r requirements.txt
+            sphinx-build -n -W -q -b html -d _build/doctrees . _build/html
+
+      - run:
+          name: run linkcheck
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r requirements.txt
+            # don't fail the build if linkcheck fails
+            make linkcheck || true
+
+            # make "broken" report
+            cat _build/linkcheck/output.txt | grep broken > _build/linkcheck/broken.txt
+
+      - store_artifacts:
+          path: _build


brettcannon:
The issue with Circle CI is we get a single builder for the entire Python organization. That means your 1 minute build time won't start in 1 minute if another project starts using it. While with Travis we have I think 20 builders across the organization.

brettcannon:
I noticed at https://github.com/python/devguide/issues/292#issuecomment-341294512 you mention how you work at Circle, so obviously I just explained something to you which you already knew. 😄

brettcannon:
If you want to discuss opportunities for the Python organization to use Circle then feel free to email core-workflow@python.org to start up a conversation (or you can email me personally if you want me to guess at what initial hurdles there may be).

levlaz:
Hi Brett,

You get 4 builders total for any open source project.

(Where open source is any public project).

levlaz:
Hey Brett, 

Cool! My previous comment was just to say that by "default" any FOSS project will get 4 builders for free. Which is cool. 

I am 99% certain that we could make something work with the python org though so I'll be in touch over that email. 

Thank you!

levlaz:
I reached out to the core-workflow team via the email that Brett provided. I suppose we can put this "on hold" while we continue the conversation over email.

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.6
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: dependency-cache-{{ checksum "requirements.txt" }}
```
merwok: This means we lose the whole virtualenv when a dependency is added or changed.  In my projects I have keys like `project-v1` and I bump the number when I change cache paths, base image or CI tools.

levlaz: Hey, 

> This means we lose the whole virtualenv when a dependency is added or changed.

I am not quite sure I understand what you mean. If you add a dependency, `pip` is still being ran so it will always install all dependencies. If it is found in the cache it will be "instant", if not it will be downloaded. 

>In my projects I have keys like project-v1 and I bump the number when I change cache paths

I am not quite sure I understand the difference between this and your example. The key can be any arbitrary thing. I just named it `dependency-cache`. In addition you can have [multiple cache keys](https://circleci.com/docs/2.0/caching/#using-keys-and-templates) which increases the chance of a cache being hit.

merwok: I mean the previous cache won’t be restored, so pip will have to install all dependencies fresh.  In contrast, a key without requirements file checksum will cause the cache to be restored, and then pip will install only the one new dependency.

willingc:
Hi all, 

Thanks to @levlaz for the Proof of Concept PR.

The conversation about CircleCI and workflow has moved to the [core-workflow mailing list](https://mail.python.org/mm3/archives/list/core-workflow@python.org/thread/D46UCLZQWJ4Y3RKT43PUDNM6XL6UGDOB/).

As an FYI, I have been happily using CircleCI alongside Travis on JupyterHub development for the past 2 years. 

Currently, I'm going to close this Proof of Concept PR, and after additional discussion on the core-workflow mailing list, we can submit a formal PR, if needed.

Thanks. - Carol :sunny:

PR #297, opened 2017-11-04
Remove $ sign from gitbootcamp.

Mariatta:


Mariatta:
Thanks @ezio-melotti. Good point about the highlighting. I rendered the page to check.
I've added `code-block` directives in places where the highlight was affected.

PR #303, opened 2017-11-08
Show the 'Last Updated' info in the footer.

### LINKED ISSUE #300: devguide was last updated at None.
At the very bottom of this page: https://devguide.python.org/, it says `Last updated on None.`.

Mariatta:
Fixes #300 .

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -139,7 +139,7 @@ html_static_path = ['tools/static']
 
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.
-#html_last_updated_fmt = '%b %d, %Y'
+html_last_updated_fmt = '%b %d, %Y'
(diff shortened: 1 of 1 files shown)


conf.py
```diff
@@ -139,7 +139,7 @@
 
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.
-#html_last_updated_fmt = '%b %d, %Y'
+html_last_updated_fmt = '%b %d, %Y'
```
ezio-melotti: What do you think about using ISO format?

Mariatta: I like the format `'%b %d, %Y'`, it is the same one we have in docs.python.org 😃

Mariatta:
Thanks @ezio-melotti :)

PR #307, opened 2017-12-07
Removes Firefox search-engine link that throws 404

### LINKED ISSUE #306: Firefox Search Engine plugin link returns 404 error
Under the key resources section: 
https://devguide.python.org/#key-resources

Link to `Firefox search engine plug-in` pointing to https://www.python.org/dev/searchplugin/ appears to be broken.

CuriousLearner:
Resolves #306 

cc @brettcannon

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -265,7 +265,6 @@ Key Resources
     * `Meta tracker <http://psf.upfronthosting.co.za/roundup/meta>`_ (issue
       tracker for the issue tracker)
     * :doc:`experts`
-    * Firefox search engine plug-in
 * `Buildbot status`_
 * Source code
     * `Browse online <https://github.com/python/cpython/>`_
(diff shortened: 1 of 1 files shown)


index.rst
```diff
@@ -265,7 +265,7 @@ Key Resources
     * `Meta tracker <http://psf.upfronthosting.co.za/roundup/meta>`_ (issue
       tracker for the issue tracker)
     * :doc:`experts`
-    * `Firefox search engine plug-in`_
+    * Firefox search engine plug-in
```
ezio-melotti: If the plug-in is gone, then there's no reason to keep this line around.

CuriousLearner: Oh, @brettcannon suggested on the issue that the reference should be removed. 

So, what should I do?

ezio-melotti: Both this line and the link should be removed.

CuriousLearner: @ezio-melotti fixed :)

CuriousLearner:
Thanks @ezio-melotti 👍

PR #309, opened 2017-12-13
Make News.d/next filename description match blurb

bennorth:
The filenames for news snippets in subdirectories of `NEWS.d/`, as generated by `blurb`, include both the date and time ([source](https://github.com/python/core-workflow/blob/master/blurb/blurb.py#L186)).  The current description in `committing.rst` only mentions the date.  This PR updates the description to match the behaviour of `blurb`.


brettcannon:
Thanks!

PR #312, opened 2017-12-20
Update 3.1.5 EOL date

hugovk:
[PEP 375](https://www.python.org/dev/peps/pep-0375/) says:

> 3.1 received security fixes until June 2012.

But in practice the last release was 3.1.5, which PEP 375 gives as 2012-04-06 but the [release page](https://www.python.org/downloads/release/python-315/) gives as 2012-04-09 and confirms in:

> Python 3.1.5 was released on April 9, 2012.


JulienPalard:
PEP 375 initially told [3.1 will recieve security fixes until June 2014](https://github.com/python/peps/commit/c179cfa268516f971aa5b848249dec65826d909b), don't know what's the real date but Python 3.1 is dead since a long time, I'm not sure this matter.

Issue #313, opened 2018-01-22
'Show source' links to text pages and not Github


csabella:
In `docs.python.org`, 'Show source' in the sidebar links to the Github source page, but in the dev guide, it links to a static text page.  Should they function the same?

brettcannon:
I don't see any reason not to have the devguide point to the equivalent repo page.

PR #314, opened 2018-01-23
Link Show Sources to Github

csabella:
For issue #313.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -113,7 +113,7 @@ templates_path = ['tools/templates']
 html_sidebars = {
     # Defaults taken from http://www.sphinx-doc.org/en/stable/config.html#confval-html_sidebars
     # Removes the quick search block
-    '**': ['localtoc.html', 'relations.html', 'sourcelink.html']
+    '**': ['localtoc.html', 'relations.html', 'customsourcelink.html']
 }
 
 # Additional static files.
--- a/tools/templates/customsourcelink.html
+++ b/tools/templates/customsourcelink.html
@@ -1,0 +1,13 @@
+{%- if show_source and has_source and sourcename %}
+  <div role="note" aria-label="source link">
+    <h3>{{ _('This Page') }}</h3>
+    <ul class="this-page-menu">
+      <li><a href="{{ pathto('bugs') }}">{% trans %}Report a Bug{% endtrans %}</a></li>
+      <li>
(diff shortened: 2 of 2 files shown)


conf.py
```diff
@@ -113,7 +113,7 @@
 html_sidebars = {
     # Defaults taken from http://www.sphinx-doc.org/en/stable/config.html#confval-html_sidebars
     # Removes the quick search block
-    '**': ['localtoc.html', 'relations.html', 'sourcelink.html']
```
ezio-melotti: Is the sourcelink.html page used elsewhere after this change?
If not, maybe it should be removed, or perhaps you can edit the page directly instead of creating a new one.

csabella: I actually copied this code from the Docs folder under cpython (Issues 19489 and 28929).  I don't think sourcelink.html is used anywhere else, based on the Sphinx docs and code.  I know you know way more about this than I do, but here's a link to the Sphinx sourcelink.html (https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/sourcelink.html).  Since the cpython code created a custom version and stored it locally, I just assumed that that was a best practice.  I couldn't find a sourcelink.html file locally.

As far as the `_sources` folder, CPython still creates that for Docs, so I didn't know if there was a way to get rid of it.  Under Sphinx, `searchtools.js` has the following code:
```
          $.ajax({url: DOCUMENTATION_OPTIONS.URL_ROOT + '_sources/' + item[5] + (item[5].slice(-suffix.length) === suffix ? '' : suffix),
                  dataType: "text",
                  complete: function(jqxhr, textstatus) {
                    var data = jqxhr.responseText;
                    if (data !== '' && data !== undefined) {
                      listItem.append(Search.makeSearchSummary(data, searchterms, hlterms));
                    }
                    Search.output.append(listItem);
                    listItem.slideDown(5, function() {
                      displayNextItem();
                    });
                  }});
```

If that URL in that JQuery is changed to also use the Github link, then `_sources` may not be necessary.  It's autogenerated and I would need to find out where to suppress it.  But, `_sources` also isn't very large, so I didn't see any harm in keeping it.

Mariatta: `sourcelink.html` is builtin from sphinx, so it's not something we can remove. It doesn't exist within devguide's repo.
So I think the approach of using customsourcelink works.

PR #318, opened 2018-01-29
Mention how to add co-author info.

Mariatta:
When converting other people's patch, ensure to add the "Co-authored-by:" message.

### DIFF
--- a/pullrequest.rst
+++ b/pullrequest.rst
@@ -323,8 +323,10 @@ their own pull request. If the author does not respond after a week, it is
 acceptable for another contributor to prepare the pull request based on the
 existing patch. In this case, both parties should sign the :ref:`CLA <cla>`.
 When creating a pull request based on another person's patch, provide
-attribution to the original patch author by adding "Original patch by
-<author name>." to the pull request description and commit message.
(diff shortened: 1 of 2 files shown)


pullrequest.rst
```diff
@@ -323,8 +323,10 @@ their own pull request. If the author does not respond after a week, it is
 acceptable for another contributor to prepare the pull request based on the
 existing patch. In this case, both parties should sign the :ref:`CLA <cla>`.
 When creating a pull request based on another person's patch, provide
-attribution to the original patch author by adding "Original patch by
-<author name>." to the pull request description and commit message.
+attribution to the original patch author by adding "Co-authored-by: 
+Author Name <email_address> ." to the pull request description and commit message.
```
matrixise: why a Co-author ? If you are the author, and just this author, we don't have to use 'Co-authored-by'.

Mariatta: This section is about creating a pull request based on another person's patch. In this case, I think co-author is appropriate.
If there's only one author, and the patch author and PR author is the same, then there is no need to add `Co-authored-by`.

matrixise: ok thanks

Issue #324, opened 2018-02-04
Building docs without make doesn't seem to work


Mariatta:
I followed https://devguide.python.org/documenting/#without-make to build the docs without make:
```
sphinx-build -bhtml . build/html
```

It finished without error but nothing showed up under `build/`

Wonder if it's just me. Perhaps I miss something? I've followed the previous steps and `make html` works for me (after blurb got installed).

terryjreedy:
On 2/4/2018 9:49 AM, Mariatta wrote:
> I followed https://devguide.python.org/documenting/#without-make to 
> build the docs without make:
> 
> |sphinx-build -bhtml . build/html |
> 
> It finished without error but nothing showed up under |build/|

Where are you running sphinx-build?  My local repository has a build/ 
directory that consists of nothing but leftover test artifacts (which 
should probably be removed by the tests).  That should *not* have html/.

I just ran the above (with a full path to sphinx-build) in /Doc, and 
/Doc/build/html was (still) populated with updated files, as expected. 
My sphinx install is probably a couple of years old, so that might 
possibly make a difference.

matrixise:
Hi @Mariatta 

In a docker image based on the last ubuntu:rolling

```
docker run --rm -it ubuntu:rolling /bin/bash
apt-get update && apt-get install -y vim && vim /etc/apt/sources.list
# modify the repository to include the sources
apt-get update
apt-get install -y git-core python3-venv
apt-get build-dep -y python3
cd /tmp
git clone https://github.com/python/cpython
python3 -m venv /tmp/venv
source /tmp/venv/bin/activate
# we recommend to install sphinx and the dependencies
pip install wheel sphinx==1.6.1
cd cpython/Doc
# just build the documentation
sphinx-build -bhtml . build/html
# result is a build/html directory with the compiled documentation
```

so in my case, I have no issue with the documentation.

matrixise:
@Mariatta here is a small Dockerfile, you can read it https://gist.github.com/matrixise/a8d27af4c874da5fdc1d0b8cae5afccf maybe we could use it for the build of the documentation for the newcomers, but this one is not yet perfect.

it's just for the fun, I won't use it in production because I have no issue with the virtualenv system and Python in general. but normally with this Dockerfile, I just explain how I compile the documentation.

ned-deily:
[...]

PR #327, opened 2018-02-07
Remove the 'Without make' section

matrixise:
Firstly, 'Without make' makes no sense because we can pass a parameter to the
``make`` command, for example, ``make html``, or ``make latexpdf`` and
we will get the same result.

Secondly, We have to provide an easy way for the developers which want
to generate the documentation.


matrixise:
Hi all,

I just created this PR, but I prefer that we discuss before an eventual merge.

@Mariatta and @ned-deily

JulienPalard:
Why removing it? I can imagine some distro without `make` where it make sense to be able to simply build the doc this way.

ezio-melotti:
Are you sure that using `make` is always possible?

I suggest 3 different ways to deal with this (depending on the answer of the previous question):
1) leave it as is;
2) Merge the two sections and convert the latter to a single line/paragraph at the end (``It is also possible to build the documentation by running `sphinx-build -b<builder> . build/<builder>` in the `Doc` dir, where `<builder>` is one of html, text, latex, or htmlhelp.``);
3) Remove the section and the header of the previous one (`using make`);

terryjreedy:
Ned is correct that this section is indeed a copy.  It can be criticized for repeating 'install sphinx' (already said above) and being too general (the topic is building the html version).  I suggest condensing it to just two lines before the 'See also' note, with the specific command.

To build html without make, run the following in the Doc directory:
    sphinx-build -bhelp . build/html

I suggest putting this before 'see also' because 'see also' applies to this as well as the make stuff.

brettcannon:
But people can always read the `Makefile`. If people feel that strongly about this then they can simply state the fact that people can always inspect the `Makefile` manually to see what it's doing. Otherwise we are needlessly duplicating what is in the `Makefile` in the docs which becomes annoying when maintaining the `Makefile`.

terryjreedy:
Reading the windows make.bat is a real pain and extracting the single command needed for developers to check their docs patches is not completely trivial.  Make.bat also merges blurbs into the news log, which is irrelevant unless one uses rst makeup.

brettcannon:
OK, so it sounds like we are not going to reach consensus beyond "keep it as small as possible", so let's aim for that.

willingc:
Hi @matrixise,

Let's get the PR completed finished off. I have two suggestions below if you would like to add them. If you are too busy post-PyCon to do so, I will happily add them. Just let me know... Thanks!

I left a message on the open issue #324:

Next actions:

- Rename section "not make" to "using sphinx-build"
- Add a sentence or two on why using sphinx-build is useful in some cases e.g. using command line options such as verbosity, nit-picky, etc. http://www.sphinx-doc.org/en/master/man/sphinx-build.html#options

matrixise:
@willingc when you have time, just check my new commit. 

PS: I am really sorry for this very late answer :/

documenting.rst
```diff
@@ -1565,11 +1565,12 @@ You can also use ``make help`` to see a list of targets supported by
...
 See also :file:`Doc/README.rst` for more information.
 
+Using sphinx-build
+------------------
 
-Without make
-------------
-
-Install the Sphinx, blurb, and python-docs-theme packages from PyPI.
+Sometimes we directly want to execute the sphinx-build tool, in this case, you can
```
brettcannon: "tool instead of through ``make`` (although the latter is still the preferred way). In this case,"""

documenting.rst
```diff
@@ -1565,11 +1565,12 @@ You can also use ``make help`` to see a list of targets supported by
...
+use the following command line but before you have to install the Sphinx, blurb,
```
brettcannon: "command line from the ``Doc`` directory (make sure to install Sphinx, blurb, and python-docs-theme from PyPI)::" and link to their PyPI pages. This then means you should remove line 1575.

matrixise:
@brettcannon just updated, thank you for your review.

PR #330, opened 2018-02-27
[bpo-32967]Fix "make check" in devguide

neerajbadlani:
[devguide] make check
python3 tools/rstlint.py -i tools -i venv
No problems found.
[devguide]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ venv/
 
 # Rope project settings
 .ropeproject
+
+# Auto/Temp Generated Files by Tools
+*.swp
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -90,3 +90,6 @@ venv/
 
 # Rope project settings
 .ropeproject
+
+# Auto/Temp Generated Files by Tools
+*.swp
```
Mariatta: I don't think we should include .swp here?

neerajbadlani: I can take it out .. but any reason ?

Mariatta:
Thanks!

PR #333, opened 2018-03-01
[don't merge yet] Reference python-docs-theme when building docs without make

Mariatta:
Only merge once https://github.com/python/cpython/pull/2017/ has been merged.


ned-deily:
As I mentioned elsewhere, I really think we should remove the "Without make" section in the devguide and just refer users to the Doc README.

Mariatta:
Thanks @ned-deily. I've completely forgotten about that discussion. But I agree, removing the without make section makes sense. I'll close this.

PR #339, opened 2018-03-11
Update triage page with tables

willingc:
The goal of this PR is to provide people new to triaging a cheatsheet-like view of the fields on bugs.python.org.

This is a first step in possibly creating a "triaging workgroup".

### DIFF
--- a/triaging.rst
+++ b/triaging.rst
@@ -278,7 +278,8 @@ Status
 |               | information to reproduce the bug reported.  Requesting     |
 |               | additional information and setting status to *pending*     |
 |               | indicates that the issue should be closed if the necessary |
-|               | information in a timely manner (i.e. one month).           |
+|               | information is not provided in a timely manner (i.e. one   |
+|               | month).                                                    |
 +---------------+------------------------------------------------------------+
 | closed        | The issue has been resolved (somehow).                     |
 +---------------+------------------------------------------------------------+
@@ -298,7 +299,7 @@ with the "open" status.
 +---------------+------------------------------------------------------------+
 | fixed         | A fix for the issue was committed.                         |
 +---------------+------------------------------------------------------------+
-| later         | Issue is to be worked on at a later date.                  |
+| later         | Issue is to be worked on in a later release cycle.         |
 +---------------+------------------------------------------------------------+
 | not a bug     | For some reason the issue is invalid (e.g. the perceived   |
 |               | problem is not a bug in Python).                           |
@@ -306,7 +307,8 @@ with the "open" status.
 | out of date   | The issue has already been fixed, or the problem doesn't   |
 |               | exist anymore for other reasons.                           |
 +---------------+------------------------------------------------------------+
-| postponed     | Issue will not be worked on at the moment.                 |
+| postponed     | Issue will not be worked on at the moment but in a future  |
+|               | minor release version.                                     |
 +---------------+------------------------------------------------------------+
 | rejected      | Issue was rejected (especially for feature requests).      |
 +---------------+------------------------------------------------------------+
--- a/triaging.rst
+++ b/triaging.rst
@@ -10,12 +10,13 @@ Fields in the Issue Tracker
 ---------------------------
 
 The major elements found in an issue report include:
-* Classification (including Title) - These fields categorize the issue
-  The fields include Title, Type, Stage, Components, and Version.
+
+* Classification (including *Title*) - These fields categorize the issue
+  The fields include *Title*, *Type*, *Stage*, *Components*, and *Version*.
 * Process - These fields indicate the state of the issue and its progress
-  toward resolution. The fields include: Status, Resolution, Dependencies,
-  Superseder, Assigned To, Nosy List, Priority, Keywords, Comment, File,
-  File Description, Remote hg repo, GitHub PR.
+  toward resolution. The fields are *Status*, *Resolution*, *Dependencies*,
+  *Superseder*, *Assigned To*, *Nosy List*, *Priority*, *Keywords*, *Comment*,
+  *File*, *File Description*, *Remote hg repo*, *GitHub PR*.
 * Messages
 * History
 
@@ -372,8 +373,8 @@ Checklist for Triaging
 
 * Read the issue comment(s).
 * Review and set classification fields
-    - Title: should be concise with specifics helpful to someone scanning a
-      list of issue titles
+    - Title: should be concise with specifics (such as Module name) which are
+      helpful to someone scanning a list of issue titles
     - Type
     - Stage
     - Components: multiple items may be set
--- a/triaging.rst
+++ b/triaging.rst
@@ -14,7 +14,7 @@ Fields in the Issue Tracker
 
 The major elements found in an issue report include:
 
-* Classification (including *Title*) - These fields categorize the issue
+* Classification (including *Title*) - These fields categorize the issue.
   The fields include *Title*, *Type*, *Stage*, *Components*, and *Version*.
 * Process - These fields indicate the state of the issue and its progress
   toward resolution. The fields are *Status*, *Resolution*, *Dependencies*,
@@ -27,6 +27,9 @@ Title
 '''''
 A brief description of the issue. Review whether the title is too generic or
 specifies an incorrect term or library.
+
+Adding a prefix at the start of the title to indicate the module, such as 
+IDLE, DOC, or ASYNCIO, is optional, but can be helpful to others.
 
 Type
 ''''
@@ -49,7 +52,7 @@ specific type, please do not set a type.
 |                | arguments for existing functions. Also used for          |
 |                | improvements in the documentation, test suite and        |
 |                | other refactorings. A good place to discuss enhancements |
-|                | prior to filing an issue is **python-ideas** mailing     |
+|                | prior to filing an issue is `python-ideas`_ mailing      |
 |                | list.                                                    |
 +----------------+----------------------------------------------------------+
 | performance    | Situations where too much time is necessary to complete  |
@@ -60,8 +63,9 @@ specific type, please do not set a type.
 |                | used.                                                    |
 +----------------+----------------------------------------------------------+
 | security       | Issues that might have security implications. Report     |
-|                | security vulnerabilities which should not be made public |
-|                | to security@python.org.                                  |
+|                | security vulnerabilities using the procedure found in    |
+|                | the `Reporting security issues in Python`_ page on the   |
+|                | python.org website.                                      |
 +----------------+----------------------------------------------------------+
 
 Stage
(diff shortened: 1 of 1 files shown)


csabella:
Wow, what a great idea.  This is much clearer - thanks for all the work!  :-)  

One thing that I wanted to suggest, related to the bug tracker, but unrelated to this issue, is to put the relevant module as the first word in the title.  I know that adding a new dropdown for module has been turned down several times, but I find that issues with the module as the first word are much easier to navigate (and to sort on).  I only bring it up here because you changed the description for 'Title'.

willingc:
Thank you @csabella. I was going to ping you later to see if you had time to review this WIP PR. Feel free to leave suggestions, clarifications, etc. directly in the code.

triaging.rst
```diff
@@ -206,49 +242,63 @@ The issue is a duplicate of the listed issue(s).
...
+| pending       | The issue is blocked until someone (often the              |
+|               | :abbr:`OP (original poster)`) provides some critical       |
+|               | information; the issue will be closed after a set amount   |
+|               | time if no reply comes in.                                 |
+|               |                                                            |
+|               | Useful when someone opens an issue that lacks enough       |
+|               | information to reproduce the bug reported.  Requesting     |
+|               | additional information and setting status to *pending*     |
+|               | indicates that the issue should be closed if the necessary |
+|               | information in a timely manner (i.e. one month).           |
```
csabella: 'if the necessary information' is missing the verb.

triaging.rst
```diff
@@ -206,49 +242,63 @@ The issue is a duplicate of the listed issue(s).
...
++---------------+------------------------------------------------------------+
+| later         | Issue is to be worked on at a later date.                  |
++---------------+------------------------------------------------------------+
+| not a bug     | For some reason the issue is invalid (e.g. the perceived   |
+|               | problem is not a bug in Python).                           |
++---------------+------------------------------------------------------------+
+| out of date   | The issue has already been fixed, or the problem doesn't   |
+|               | exist anymore for other reasons.                           |
++---------------+------------------------------------------------------------+
+| postponed     | Issue will not be worked on at the moment.                 |
```
csabella: What is the difference between `postponed` and `later`?

triaging.rst
```diff
@@ -262,34 +312,64 @@ indicate a remote branch by adding ``#BRANCH`` to the end of the URL.
...
+| ``msg<number>``                                             | Links to the tracker message ``<number>``.            |
++-------------------------------------------------------------+-------------------------------------------------------+
+| ``PR <number>``,                                            | Links to `GitHub pull requests`_.                     |
+| ``PR<number>``, or                                          |                                                       |
+| ``pull request <number>``                                   |                                                       |
++-------------------------------------------------------------+-------------------------------------------------------+
+| a 10-, 11-, 12-, or 40-digit hex ``<number>``               | Indicates a Git or Mercurial changeset identifier and |
+|                                                             | generates a link to changeset ``<number>``on GitHub   | 
+|                                                             | or https://hg.python.org/. The ``git`` and ``hg``     |
```
csabella: The markup is off on the description for commit number.

triaging.rst
```diff
@@ -3,193 +3,243 @@
...
++----------------+----------------------------------------------------------+
+| crash          | Hard crashes of the Python interpreter -- possibly with  |
+|                | a core dump or a Windows error box.                      |
++----------------+----------------------------------------------------------+
+| enhancement    | Issues that propose the addition of new functionality,   |
+|                | such as new functions, classes, modules, or even new     |
+|                | arguments for existing functions. Also used for          |
+|                | improvements in the documentation, test suite and        |
+|                | other refactorings. A good place to discuss enhancements |
+|                | prior to filing an issue is **python-ideas** mailing     |
```
csabella: Can python-ideas be a link?

triaging.rst
```diff
@@ -3,193 +3,243 @@
...
+* History
 
 Title
 '''''
-Should be properly descriptive of what the issue is about. Occasionally
-people file an issue that either has too generic of a title or end up thinking
-they filed about X but in fact it turns out to be about Y and thus the
-title is now wrong.
+A brief description of the issue. Review whether the title is too generic or
+specifies an incorrect term or library.
```
csabella: I mentioned this in my other comment, but mentioning it here too.  This may need to be a new issue for further discussion.  I think it would be helpful for the title to begin with the module that the issue is for.  A large portion of the tickets have this now, but not all of them.  It makes search (and sorting) easier to find tickets related to a module.

willingc: Let's recommend that folks do as you suggest. Do you have a format that has worked well?

csabella: Thanks for asking!  :-)  Well, for IDLE, we use `IDLE:` and I think that let's everyone know they can skip it.  :-)  Seriously though, if you go into the bug tracker and then search, then click 'Sort on' next to title, you'll see (with sort descending), that the `zlib` and `zip` modules come up first.  That's really just the format I'd suggest, putting the module first instead of somewhere else in the title.  I don't think a colon after it is necessary, unless it reads better.  I've downloaded all the bugs using the .csv download and realized this could make it a lot easier for people to find related tickets.  The old 'I think there's a ticket for this, but I can't find it' becomes easier when you can narrow the search.  :-)  It also gives a sense for the modules that may be neglected (for example, argparse has a ton of open tickets).

csabella:
One question that I had about triaging, especially old tickets, that you may or may not want to address on this page, is when is it helpful and when is it noise?  I've been going through old documentation issues and marking some as 'easy' so that new contributors can find them more easily.  I try to only do a few a week as to not make a lot of old tickets prominent.  But, for example, there have been many issues that I have seen which could get the 'languishing' tag or have the title changed as per my previous suggestion.  But, doing a lot of that at once would just be noise.  Maybe you can suggest guidelines around that?

brettcannon:
@csabella honestly, if you're motivated to update something manually, then just do it. 😄 Making sure something actually occurs rather than have it be forgotten is more important than worrying about people getting another email.

triaging.rst
```diff
@@ -3,193 +3,243 @@
...
+| performance    | Situations where too much time is necessary to complete  |
+|                | the task. For example, a common task now takes           |
+|                | significantly longer to complete.                        |
++----------------+----------------------------------------------------------+
+| resource usage | Situations where too many resources (e.g. memory) are    |
+|                | used.                                                    |
++----------------+----------------------------------------------------------+
+| security       | Issues that might have security implications. Report     |
+|                | security vulnerabilities which should not be made public |
+|                | to security@python.org.                                  |
```
brettcannon: Should probably link to https://www.python.org/news/security/ as it has e.g. PGP key to use.

triaging.rst
willingc: @csabella Added below and in the checklist.

willingc:
@csabella @brettcannon I think that I've covered your review comments. I've removed the WIP.

csabella:
LGTM.

triaging.rst
```diff
@@ -3,193 +3,247 @@
...
 '''''
-Should be properly descriptive of what the issue is about. Occasionally
-people file an issue that either has too generic of a title or end up thinking
-they filed about X but in fact it turns out to be about Y and thus the
-title is now wrong.
+A brief description of the issue. Review whether the title is too generic or
+specifies an incorrect term or library.
+
+Adding a prefix at the start of the title to indicate the module, such as 
+IDLE, DOC, or ASYNCIO, is optional, but can be helpful to others.
```
ncoghlan: For most modules, and even `doc` as a pseudo-module, I'd suggest a lower case prefix - caps are kinda shouty. (`IDLE` is an exception, since the IDE itself actually is called `IDLE`)

triaging.rst
```diff
@@ -3,193 +3,247 @@
...
+will be needed.
+
++---------------+----------------------------------------------------------+
+|     Stage     |                       Description                        |
++===============+==========================================================+
+| test needed   | The steps which are needed to reproduce the issue. The   |
+|               | bug reporter should post a script, instructions, or      |
+|               | example to help someone test or reproduce the issue.     |
++---------------+----------------------------------------------------------+
+| needs patch   | A patch is needed to solve the problem (i.e. fixing the  |
```
ncoghlan: Should we use "patch or pull request" in this table, with a short note below about their functional equivalence? They're pretty much interchangeable as far as the tracker is concerned, but readers that are only familiar with GitHub may not have encountered the "patch" term.

triaging.rst
```diff
@@ -3,193 +3,247 @@
...
+|                  | an alpha release.                                      |
++------------------+--------------------------------------------------------+
 
 As a guideline, *critical* and above are usually reserved for crashes,
 serious regressions or breakage of very important APIs.  Whether a bug
-is a *release blocker* is a decision better left to the release manager so,
-in any doubt, add him or her to the *nosy list*.
+is a *release blocker* is decided by the release manager. Triagers may
+recommend this priority and should add the release manager to the 
+*nosy list*.
```
ncoghlan: Should we link to https://devguide.python.org/#status-of-python-branches here, and perhaps add the RM's tracker ID's to that table?

triaging.rst
```diff
@@ -262,34 +332,74 @@ indicate a remote branch by adding ``#BRANCH`` to the end of the URL.
...
-* ``Dir/file.ext`` and ``Dir/file.ext:NNN`` generate links to files in the
-  `Python source code repositories <https://github.com/python/cpython/>`_,
-  possibly linking to the line number specified after the ``:``.
-  ``3.6/Dir/file.ext`` will generate a link with ``3.6`` as branch.
-* ``PEP <number>`` and ``PEP<number>`` link to the
-  :abbr:`PEP (Python Enhancement Proposal)` ``<number>``.
-* ``devguide`` (lowercase), ``devguide/triaging``, and
-  ``devguide/triaging#generating-special-links-in-a-comment`` generate links to
-  the Devguide, this page, and this section respectively.
+Comment abbreviation can automatically generate a link to various web pages.
```
ncoghlan: Perhaps "Using the following abbreviations in a comment will automatically generate a link to relevant web pages."

triaging.rst
willingc: Great point. Will add pull request.

triaging.rst
willingc: Good suggestion

triaging.rst
willingc: Good suggestion. Will do.

triaging.rst
willingc: Going to fix this as a I see a typo as well.

willingc:
I've gone ahead and merged. Further changes can be handled in future PRs. Thanks @csabella @brettcannon and @ncoghlan for the reviews 💐

ezio-melotti:
Thanks for working on this @willingc, the new version looks great!

PR #342, opened 2018-03-13
Add Carol's motivations

willingc:


willingc:
Thanks @Mariatta for the review. 🌴

Issue #189, opened 2017-05-06
Simplify the index page by audience


brettcannon:
The index page has become rather long and is starting to feel like a wall of text. It should probably be clearly separated into 3 sections:

1. New contributors
   a. New to open source (e.g. needs help to build Python)
   b. Just new to the project (i.e. just needs to know how the project does things
2. Core devs
3. Appendix

Once we do this I think it will help show redundancies that have accumulated over time (e.g. all the git information that we have spread out across the devguide).

ncoghlan:
I think the core dev section may also need to be split, although I'm not sure what the two sub-audiences should be called. The way I think of them:

* Professional software developer: your day job specifically involves writing software to meet the needs of someone other than yourself. As part of that work, you're already familiar with automated testing, reviewing code changes, discussing design trade-offs, and deciding the appropriate scope of particular projects.
* Domain expert: your primary interest & expertise lies in an area other than professional software development (e.g. computer science education, general education, journalism, legal practice, economics, scientific research, data analysis, system operations & administration, systems engineering). Contributing to CPython accounts for the bulk of your experience with the practices & processes involved in developing software as part of a larger team working on a code base that's too large for any one person to comprehensively understand.

For the first group, the key things to highlight are those points that relate to collaborating with self-motivated volunteers rather than working solely with full-time colleagues as part of a paid development team. Most importantly: don't assume everyone else working on the project is a professional software developer.

For the second group, we may need to be more explicit about things we'd consider "foundational knowledge you should already have learned at work" for a professional software dev (such as how to use a code review tool as a reviewer, or the right terms to use to search for things on Google).

terryjreedy:
Thank you for identifying the two groups and suggesting that the second group, which includes me, should not necessarily by empty.

pfmoore:
Agreed - I'd put myself in the second group too. Even though I'm technically a developer, in practical terms the processes used in my line of work are so different from those used in open source that it may as well be a different profession. (Although I've worked enough in open source that the tools are familiar, the processes are most decidedly not).

I'd also point out that for group 2 we should also include reminders that many core devs are in group 1 - which means that they may not even understand the difficulties you're having with the tools and processes, so work with them accordingly.

ncoghlan:
Attempting a more value-neutral description of the two categories:

[...]

PR #349, opened 2018-03-24
add navigation table by contributor type to landing page

willingc:
Addresses #189 

This PR adds navigation by type of user to the top of the devguide index page. This partially solves 2 issues: 
- guide folks by roles to relevant docs in a space efficient manner
- the sidebar toctree navigation requires multiple clicks to get to a particular page in the devguide. This small table gives one-click access to frequently used pages.

screenshot 2018-03-24 15 21 16

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -5,9 +5,27 @@ Python Developer's Guide
 .. highlight:: bash
 
 This guide is a comprehensive resource for :ref:`contributing <contributing>`
-to Python_ -- for both new and experienced contributors.  It is
-:ref:`maintained <helping-with-the-developers-guide>` by the same community
-that maintains Python.  We welcome your contributions to Python!
+to Python_ -- for both new and experienced contributors. It is 
+:ref:`maintained <helping-with-the-developers-guide>` by the same
+community that maintains Python.  We welcome your contributions to Python!
+
++------------------------+---------------------+-----------------------+---------------------+
+| New Contributors       | Documentarians      | Triagers              | Core Developers     |
++========================+=====================+=======================+=====================+
+| :doc:`setup`           | :doc:`docquality`   | :doc:`tracker`        | :doc:`coredev`      |
++------------------------+---------------------+-----------------------+---------------------+
+| :doc:`help`            | :doc:`documenting`  | :doc:`triaging`       | :doc:`developers`   |
++------------------------+---------------------+-----------------------+---------------------+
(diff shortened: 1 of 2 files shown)


index.rst
```diff
@@ -5,9 +5,27 @@ Python Developer's Guide
...
++========================+=====================+=======================+=====================+
+| :doc:`setup`           | :doc:`docquality`   | :doc:`tracker`        | :doc:`coredev`      |
++------------------------+---------------------+-----------------------+---------------------+
+| :doc:`help`            | :doc:`documenting`  | :doc:`triaging`       | :doc:`developers`   |
++------------------------+---------------------+-----------------------+---------------------+
+| :doc:`pullrequest`     | :ref:`style-guide`  | :ref:`helptriage`     | :doc:`committing`   |
++------------------------+---------------------+-----------------------+---------------------+
+| :doc:`runtests`        | :ref:`rst-primer`   |                       | :doc:`devcycle`     |
++------------------------+---------------------+-----------------------+---------------------+
+| :doc:`fixingissues`    |                     |                       | :doc:`experts`      |
```
ncoghlan: Perhaps migrate "experts" over to the Triagers column? Core devs update that list, but it's triagers that rely on it to update the nosy list appropriately.

And since we have room in the column, perhaps point core devs at the Motivations & Affiliations page?

csabella:
Very nice!

ezio-melotti:
I kind of like the idea, but I'm not sure about the current implementation:
1. this moves the quick reference further down. After the quick ref there are already a couple other sections with more or less detailed lists of links that might be replaced by this table.
2. While the table stands out and is space-efficient horizontally, it's probably difficult to navigate on mobile and still has lot of empty space.  Perhaps merging the triagers/core devs column or using a different format might be better.
3. I think some of those links could be left out, and the table could be made shorter.
4. While this helps addressing #189, I feel it doesn't quite convey the separation I described in the messages there.

willingc:
@ezio-melotti I'm reworking this PR based on your comments.

#352 will improve the UX and navigation by making all devguide sections accessible from the landing page and others.

I plan to rework this PR to put the table a bit lower on the page in the Contributing section. This should satisfy your concern about the top of the page. The table IMHO is better than the wall of text (whether on mobile or not). I do believe at some point refactoring to a responsive docs theme would be a good move.

willingc:
Rendering after editing PR for Ezio's feedback.

screenshot 2018-03-25 17 05 57

Issue #120, opened 2017-02-17
Post-migration Devguide reorganization


ezio-melotti:
I think the Devguide is due for some reorganization after the migration.

In particular:
* [The version control setup section](https://github.com/python/devguide/blob/master/setup.rst#version-control-setup) should link to a separate page "git" page.
* [The lifecycle of a PR page](https://github.com/python/devguide/blob/master/pullrequest.rst) seems to expand on the quick start, but doesn't include detailed setup instructions and git commands.  This page seem to be aimed at contributors.  Also the Generation section should be under Preparation, alongside another subsection about running tests.
* [The runtests page](https://github.com/python/devguide/blob/master/runtests.rst) expands on a paragraph [pullrequest/preparation](https://github.com/python/devguide/blob/master/pullrequest.rst#preparation).
* The [silence warnings](https://github.com/python/devguide/blob/master/silencewarnings.rst) and [fixing easy issues](https://github.com/python/devguide/blob/master/fixingissues.rst) are both 2-paragraphs pages that might be moved as sections of other pages.
* The [tracker](https://github.com/python/devguide/blob/master/tracker.rst) and [triaging](https://github.com/python/devguide/blob/master/triaging.rst) page should be updated and perhaps merged.
* The  [where to get help](https://github.com/python/devguide/blob/master/help.rst) and [following Python's development](https://github.com/python/devguide/blob/master/communication.rst) could be merged.
* [The commiting page](https://github.com/python/devguide/blob/master/committing.rst) seems to be aimed to core-dev, but since now the workflow for both core devs and contributors is quite similar, it should perhaps be merged with the [pullrequest page](https://github.com/python/devguide/blob/master/pullrequest.rst).  The page also contains setup info and other git commands that are useful to contributors too.
 This could be handled in different ways:
  - 1 page, additional sub-sections for core-devs only
  - 1 page, with a single section that includes all the core-devs info
  - 2 pages, one for contributors only, the other for the core-devs info
* [The continuous integration page](https://github.com/python/devguide/blob/master/buildbots.html) should be updated.
* The [experts page](https://github.com/python/devguide/blob/master/experts.html) should be moved closer to the [developers page](https://github.com/python/devguide/blob/master/developers.html).
* [The mercurial for git developers page](https://github.com/python/devguide/blob/master/gitdevs.html) is now unnecessary, and it should either be removed, or better reversed and become "git for mercurial developers".

The other pages don't have any major issue :)

Mariatta:
[...]

Issue #40, opened 2016-09-23
Add an appendix to the index page


brettcannon:
Stuff that is not part of what a normal person will care about can go in there, e.g. details about the compiler or how to update the grammar.

willingc:
@brettcannon Do you wish to close this or perhaps clarify further? I'm not sure what you mean by appendix and what content it would hold.

willingc:
@brettcannon #354 seems to have taken a reasonable first step on addressing this. I'm going to close this issue. We can always add new issues with specific action items. 👍

PR #354, opened 2018-03-26
add appendix with topic pathways

willingc:
Related to #40, #120, #189 

This PR adds an Appendix that groups the individual document pages into logical subtopics. The goal of this PR is to gather topics together to help:

- consolidation and refactoring of sections that only have one paragraph
- deduplicate redundant content

My hope is that this can be a first pass and that we can iterate on this in future PRs.

screenshot 2018-03-26 07 47 19

### DIFF
--- a/appendix.rst
+++ b/appendix.rst
@@ -1,18 +1,27 @@
 Appendix: Topics 
 ================
 
-Basics
-------
+Basics for contributors
+-----------------------
 
-* :doc:`setup`
+.. note:: **Recommended reading**
+
+   - :doc:`setup`
+   - :doc:`pullrequest`
+
 * :doc:`help`
 * :doc:`communication`
 
 Core developers
 ---------------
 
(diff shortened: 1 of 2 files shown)


ezio-melotti:
This looks great, and might even be a first step in reordering the pages in the devguide (something I wanted to do as part as #120), eventually making this appendix obsolete.

One distinction I want to make that is missing from this, is that contributors should read setup.rst and pullrequest.rst; core devs should read setup.rst, pullrequest.rst, and committing.rst.
All other pages don't necessarily have to be read, but consulted if and when the need arises.

willingc:
Hi @ezio-melotti. Glad you like it. My goal was to make this a starting point for refactoring/deduplicating content, and I agree it may be unneeded after that.

I've gone ahead and added the sections you mention and put them in a note box. I'm going to go ahead and merge. Feel free to iterate on another PR. Thanks.

PR #364, opened 2018-05-14
Added additional instructions to Mac build section for building python >= 3.7

lcuio:
The mac instructions are a little unclear for python versions >= 3.7 and results in openssl not used in build when using homebrew for deps. This adds instructions on what to do.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -216,12 +216,16 @@ with **Homebrew**::
 
     $ brew install openssl xz
 
-and configure::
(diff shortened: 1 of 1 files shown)


lcuio:
@ned-deily

lcuio:
@vstinner

vstinner:
Lance Ure (@lcuio): Congratulations for your change merged into the devguide 🎉!

PR #368, opened 2018-05-15
Update details of the compiler implementation.

serhiy-storchaka:


serhiy-storchaka:
I have added core devs who worked with the bytecode generation in recent years. @markshannon, please take a look too.

compiler.rst
```diff
@@ -369,25 +378,34 @@ bytecode step of the compiler.  Several pieces of code throughout Python depend
...
+changed, the ranges in the ``magic_values`` array in :file:`PC/launcher.c`
+must also be updated.  Changes to :file:`Lib/importlib/_bootstrap_external.py`
+will take effect only after running ``make regen-importlib``.
 
 Finally, you need to introduce the use of the new bytecode.  Altering
 :file:`Python/compile.c` and :file:`Python/ceval.c` will be the primary places
-to change. But you will also need to change the 'compiler' package.
-The key files to do that are :file:`Lib/compiler/pyassem.py` and
-:file:`Lib/compiler/pycodegen.py`.
+to change. You must to add the case for a new opcode into the 'switch'
```
ericvsmith: Extra word "to": "You must add the case for a new opcode ...".

compiler.rst
```diff
@@ -396,20 +414,22 @@ the magic number if you change the bytecode, while you are debugging your work
 you will be changing the bytecode output without constantly bumping up the
 magic number.  This means you end up with stale .pyc files that will not be
 recreated.
-Running ``find . -name '*.py[co]' -exec rm -f {} ';'`` should delete all .pyc
+Running ``find . -name '*.py[co]' -exec rm -f '{}' +`` should delete all .pyc
```
ericvsmith: Things I learned today: Usage of `+` in `find exec`. Is this widely supported? It does works on a Mac. I can check Cygwin if I can ever get out of Cleveland.

serhiy-storchaka: It is the part of POSIX and I'm sure it is widely supported on systems used for developing future CPython versions.

It was not supported on some ancient commercial Unixes like SunOS 5.8 15 years ago: http://www.manpages.info/sunos/find.1.html.

compiler.rst
serhiy-storchaka: Thank you! Fixed.

willingc:
Thanks @serhiy-storchaka. Definitely an improvement. :sunny:

PR #370, opened 2018-05-15
coredev.rst: Clarify "Developer role" and "is commiter" bit are separate

encukou:
I went through the list of things to do and didn't realize these are separate bits.
Add info on what they do (== how you can check), and explicitly mention the orthogonality.


willingc:
Thanks for the contribution @encukou :cake:

PR #376, opened 2018-05-26
Adding adrianliaw as the coordinator of the zh-tw translation.

JulienPalard:
See: https://mail.python.org/pipermail/doc-sig/2018-May/004030.html

### DIFF
--- a/experts.rst
+++ b/experts.rst
@@ -364,4 +364,5 @@ Japanese       inada.naoki
 Bengali India  kushal.das
 Hungarian      gbtami
 Portuguese     rougeth
+Chinese (TW)   adrianliaw
 =============  ============


terryjreedy:
I presume that the long phrase 'Chinese as used in Taiwan' will expand the drop-down box considerably.  If so, I think something narrower would be better.  It also seems a bit awkward as a phrase rather than a title.   Would 'Chinese (Taiwan)' be acceptible?   What is there were a separate 'Portuguese as used in Brazil'?  Would not 'Portuguese (Brazil)' by okay?

ncoghlan:
Given that we're likely to gain an entry for Mainland China as well, perhaps it would make sense to use "Chinese (CN)" and "Chinese (TW)", to directly align with the `zh-cn` and `zh-tw` language subtags?

That also keeps us firmly in the space of describing regional languages, without inadvertently making any political comments through our particular choice of phrasing when describing those regions.

willingc:
I like @ncoghlan's suggested text. Thanks @JulienPalard.

terryjreedy:
I agree with using the 'official' regional subtags.

JulienPalard:
I agree with you all, commit updated.

willingc:
Thanks @JulienPalard 👍

PR #380, opened 2018-05-28
Improve make instructions on Windows

andresdelfino:


documenting.rst
```diff
@@ -1494,8 +1494,16 @@ You can also use ``make help`` to see a list of targets supported by
...
-emulate :command:`make` as closely as possible.
+**On Windows**, run the following from the root of your :ref:`repository clone
+<checkout>` to build the output as HTML::
+
+   cd Doc
+   make html
+
+While the :file:`make.bat` batchfile tries to emulate :command:`make` as
+closely as possible, the venv target is not implemented, so the required
+dependencies will be installed on your system if you do not setup a virtual
```
brettcannon: "will be" -> "will need to be"?

andresdelfino: Just a comment in case you don't know how make.bat works: it actually installs all dependencies before building the docs, so the user doesn't have to install anything.

Do you still think it's better to change this?

andresdelfino: If you already knew that and think your wording is better, I'll change it :)

brettcannon: Then I would say something like "will be automatically installed" and mention people probably want to make sure they are working in a virtual environment first.

andresdelfino: Yeap, I agree.

PR #384, opened 2018-06-09
Clarify GH web UI usage by core devs

### LINKED ISSUE #382: GitHub quick edit through web UI not working?
https://devguide.python.org/committing/#working-with-git
```quote
An exception to this rule: you can make a quick edit through the web UI of
GitHub, in which case the branch you create can exist for less than 24 hours.
This exception should not be abused and be left only for very simple changes.
```

I tried to use this for a simple wording change in the docs.  It doesn't seem to work as 'master' is a protected branch.  I think the devguide should provide a more clear explanation of the steps (in case you can actually do these trivial edits) or we should remove that section.

willingc:
Closes #382

### DIFF
--- a/committing.rst
+++ b/committing.rst
@@ -212,10 +212,6 @@ repositories means you have to be more careful with your workflow:
   maintenance branches (currently ``2.7`` or ``3.6``).
   You should commit against your own feature branch, and create a pull request.
 
-* For a small change, you can make a quick edit through the web UI of
-  GitHub. You will still create a branch on your fork, and the change will
-  go through the PR process.
-
 It is recommended to keep a fork of the main repository around, as it allows
 simple reversion of all local changes (even "committed" ones) if your local
 clone gets into a state you aren't happy with.
--- a/committing.rst
+++ b/committing.rst
@@ -208,13 +208,13 @@ repositories means you have to be more careful with your workflow:
   dedicated to maintenance of the work before the work gets integrated in the
   main repository.
 
-  An exception to this rule: you can make a quick edit through the web UI of
-  GitHub, in which case the branch you create can exist for less than 24 hours.
-  This exception should not be abused and be left only for very simple changes.
-
 * You should not commit directly into the ``master`` branch, or any of the
   maintenance branches (currently ``2.7`` or ``3.6``).
   You should commit against your own feature branch, and create a pull request.
+
+* For a small change, you can make a quick edit through the web UI of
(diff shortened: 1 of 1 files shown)


committing.rst
```diff
@@ -208,14 +208,14 @@ repositories means you have to be more careful with your workflow:
...
-  GitHub, in which case the branch you create can exist for less than 24 hours.
-  This exception should not be abused and be left only for very simple changes.
-
 * You should not commit directly into the ``master`` branch, or any of the
   maintenance branches (currently ``2.7`` or ``3.6``).
   You should commit against your own feature branch, and create a pull request.
 
+* For a small change, you can make a quick edit through the web UI of
+  GitHub. You will still create a branch on your fork, and the change will
+  go through the PR process.
```
ncoghlan: Unfortunately, if you have merge access on the target repo, GH doesn't provide a way to request that the PR branch be created on your fork (if it did, we'd just say to use that option, and wouldn't need the policy exception).

Instead, the "Create a PR for this change" option only lets you create the branch directly in the repo you're editing.

willingc: Good point. I never realized that is what it did. I only use it for quick typos and delete the branch right after.

I'm inclined to remove the entire paragraph as most changes can go through the PR process even if the dev self merges the PR for typos and such.

ncoghlan:
The rationale for having that paragraph was to make an explicit exception to the "no personal branches in the main repo" policy so that we *can* fix docs typos through the web UI, without having to have a local checkout.

I know for me personally, I'm not going to go through the full `git checkout -b ...`->edit in text editor->`git add .`->`git commit -m ...`->`git push pr`->create PR dance just to remove a doubled word or similar trivial edit.

The other aspect to consider is that the web editing UI works normally for folks that *don't* have merge privileges (i.e. the branch gets created in their personal fork, not the target repo), so folks that *become* core developers would be surprised to find out that they can't use that workflow any more.

brettcannon:
I agree with @ncoghlan . I think the main issue in #382 was that @nascheme didn't realize he had to make a temp branch to make sure that the docs didn't accidentally break.

willingc:
@ncoghlan @brettcannon I'll rework the wording to, hopefully, better reflect what's happening. Thanks for the reviews.

ncoghlan:
Thanks @willingc! That's definitely better than what we have now, so I'm going to go ahead and merge it.

PR #385, opened 2018-06-10
Repository devguide -> cpython

andresdelfino:


willingc:
Good catch @andresdelfino. Happy to merge. Thanks!

PR #386, opened 2018-06-11
Use black for python codestyle

willingc:
There are only a few Python files in the devguide repo. I've run them through the Black codestyle formatter for easier readability.


conf.py
```diff
@@ -18,14 +18,14 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('.'))
+# sys.path.insert(0, os.path.abspath('.'))
```
serhiy-storchaka: I'm not sure this is an enhancement.

This is not a human-readable comment as above lines, but a commented out sample. Currently you need to press DEL only once for switching it on. With your changes you will need to press DEL twice, and it will be harder to distinguish it from the above comment.

conf.py
```diff
@@ -111,61 +111,61 @@
 html_sidebars = {
     # Defaults taken from http://www.sphinx-doc.org/en/stable/config.html#confval-html_sidebars
     # Removes the quick search block
-    '**': ['localtoc.html', 'globaltoc.html', 'relations.html', 'customsourcelink.html'],
+    '**': ['localtoc.html', 'globaltoc.html', 'relations.html', 'customsourcelink.html']
```
serhiy-storchaka: Trailing comma makes the dict easier for modification.

conf.py
```diff
@@ -174,57 +174,67 @@
...
 latex_documents = [
-  ('index', 'PythonDevelopersGuide.tex', u'Python Developer\'s Guide Documentation',
-   u'Brett Cannon', 'manual'),
+    (
+        'index',
+        'PythonDevelopersGuide.tex',
+        u'Python Developer\'s Guide Documentation',
+        u'Brett Cannon',
+        'manual',
+    )
```
serhiy-storchaka: Trailing comma makes the list easier for modification.

And I'm not sure that writing 5-tuple vertically adds readability (especially if add yet few 5-tuples) .

tools/rstlint.py
```diff
@@ -111,12 +197,14 @@ def check_line_length(fn, lines):
...
-               'http://' not in line and \
-               not line.lstrip().startswith(('.. function',
-                                             '.. method',
-                                             '.. cfunction')):
-                yield lno+1, "line too long"
+            if (
+                line.lstrip()[0] not in '+|'
+                and 'http://' not in line
+                and not line.lstrip().startswith(
+                    ('.. function', '.. method', '.. cfunction')
```
serhiy-storchaka: The current formatting of the `startswith()` argument looks more readable to me.

tools/rstlint.py
```diff
@@ -126,18 +214,21 @@ def check_leaked_markup(fn, lines):
     """
     for lno, line in enumerate(lines):
         if leaked_markup_re.search(line):
-            yield lno+1, 'possibly leaked markup: %r' % line
+            yield lno + 1, 'possibly leaked markup: %r' % line
 
 
 def main(argv):
-    usage = '''\
+    usage = (
```
serhiy-storchaka: What is the purpose of this change?

ncoghlan:
I think I'm with Serhiy here: the utility of code formatters like black is to handle cases where the formatting really doesn't matter, and there's money to be saved by getting your developers to spend less time arguing about trivialities in a code review.

I don't think "the formatting doesn't really matter" holds in the case of documentation examples - you'll often want to format things in particular ways in order to highlight specific lines and constructs, and that formatting may not match what you might do in "real" code.

conf.py
terryjreedy: I agree.

conf.py
terryjreedy: and hence is considered good practice.

tools/rstlint.py
```diff
@@ -221,8 +312,10 @@ def main(argv):
     else:
         for severity in sorted(count):
             number = count[severity]
-            print('%d problem%s with severity %d found.' %
-                  (number, number > 1 and 's' or '', severity))
```
terryjreedy: This is a standard way to write this.

tools/rstlint.py
terryjreedy: I presume to have a closing ')' that lines up with 'usage' and 'try', as in some C styles with '}'.  But unneeded (...) around Python expressions is not typical style.

tools/rstlint.py
terryjreedy: For this one, I prefer the revision.

willingc:
Hi folks,

Thanks for the review and the comments. I expected there to be opinions on these changes since it is style ;-)

A few general comments:

- I wanted to see responses to Black's default style (with the exception of changing string format for single or double quotes). This seems like a low risk repo to test the code style as well as getting feedback from all of you. cc/ @ambv 
- As this is one of the repos that new contributors get started on, I erred on the side of scanning readability of a file over conveniences of one or two keystrokes.
- I am happy to go back and edit these files by hand for changes that folks feel strongly about.
- I'll address individual comments later today.

Thanks all :smile:

tools/rstlint.py
```diff
@@ -221,8 +312,10 @@ def main(argv):
...
+            print(
+                '%d problem%s with severity %d found.'
+                % (number, number > 1 and 's' or '', severity)
```
serhiy-storchaka: I would rewrite `number > 1 and 's' or ''` as `'s' if number > 1 else ''`.

willingc: I agree 👍

tools/rstlint.py
willingc: This one feels like it could go either way. If it's cool with folks, I'm going to err on the side of a new contributor and keep the revision which puts all of the options for starting lines in one tuple.

tools/rstlint.py
willingc: Going to change this back to the original which I agree is cleaner.

ambv: Multiline strings are one of the two places in the Python grammar where significant indentation is broken. Black tries to avoid unnecessary parentheses but is easily fooled with multiline strings that aren't aligned with the current indentation level. This is a technical limitation more than conscious design.

tools/rstlint.py
ambv: We follow PEP 8's "operators go in the front" revision from 2016. All code everywhere goes against this style because even PEP 8 used to advise the other way round. pycodestyle even has a warning about this (W503). I don't feel string interpolation is special enough to break the rules.

ambv:
FWIW removing trailing commas from indented single-element collection literals is a bug: https://github.com/ambv/black/issues/274 I will have it fixed for the next release.

willingc:
Thanks folks for the review. I've gone back and made most of the suggested changes by @terryjreedy and @serhiy-storchaka.

willingc:
@terryjreedy and @serhiy-storchaka I've made all recommended changes from your review with the exception of one place where we were in split agreement. If I don't hear any objection in a couple of days, I'm going to go ahead and merge. Thanks for the reviews:sunny:

serhiy-storchaka:
My comment was related actually to all newly added trailing commas.

conf.py
```diff
@@ -215,8 +220,13 @@
...
-    ('index', 'pythondevelopersguide', u"Python Developer's Guide Documentation",
-     [u'Brett Cannon'], 1)
+    (
+        'index',
+        'pythondevelopersguide',
+        u"Python Developer's Guide Documentation",
+        [u'Brett Cannon'],
+        1,
+    ),
 ]
```
ezio-melotti: Any reason to keep the ``u`` prefix around, especially since it's only used on some of the strings?
I would also put the square brackets and parentheses on the same line (i.e. ``[(`` and ``)]``) and remove a level of indentation.

tools/rstlint.py
```diff
@@ -20,26 +20,110 @@
...
+    'staticmethod',
+    'tabularcolumns',
+    'testcode',
+    'testoutput',
+    'testsetup',
+    'toctree',
+    'todo',
+    'todolist',
+    'versionadded',
+    'versionchanged',
```
ezio-melotti: I like the original version better.

tools/rstlint.py
ezio-melotti: For ``startswith`` both styles are ok with me (with a slight preference about the new one.
However I wouldn't go on a newline after the first ``(`` and before the last ``):``.

conf.py
ambv: Black doesn't hug multiple levels of brackets to make it trivial to scan logical nesting levels visually.

tools/rstlint.py
ambv: You need to say why for this to be actionable.

tools/rstlint.py
ambv: This style of bracket indentation is a core tenet of Black. It indents contents of all brackets the same. That makes it symmetrical and provides some usability features:
- for `def func(`, indenting after the opening bracket decouples contents from the length of the function name
- putting `):` on its own dedented line provides a visual delimiter between the function signature and body, or between the `if` statement test and body, or `while` condition and body, and so on.

tools/rstlint.py
merwok: It doesn’t take two or three screens for a human to read the whole list :)

ezio-melotti: I generally use two different styles depending on the situation:
```python
result = function(with, a, few, args,
                  with, short, names)
```
or
```python
result = a_long_named_function(
    with_a_long_named_argument,
    and_another_long_argument,
    and_maybe_an_expression_like,
    'foo' if condition else 'bar',
)
```
In the first example, putting all args on their own line, would be a waste of vertical space, and (IMHO) it's perfectly readable as is (and as the original list we are commenting was).
Readability is not only determined by the readability of the individual lines, it also depends on how easy it is to navigate vertically among functions and classes.

conf.py
ezio-melotti: The main issue is the fact that 3 lines are "wasted" for parentheses/brackets.  The argument is also similar to the one I made below.  I would either use:
```python
man_pages = [('index', 'pythondevelopersguide',
              u"Python Developer's Guide Documentation",
              [u'Brett Cannon'], 1)]
```
This takes 1/3 of vertical space compared to the "expanded" version (but I would only use it in case of a single tuple) or
```python
man_pages = [
    ('index', 'pythondevelopersguide',
     u"Python Developer's Guide Documentation",
     [u'Brett Cannon'], 1),
]
```
This takes roughly half of the vertical space, and I would use it if I want/need more horizontal space or have (or planning to add) more tuples.

It also depends on the context: if it's a module-level constant I can splurge a bit and waste more vertical space, if it's a variable or a function call in the middle of a 10-lines function, I would avoid making the function twice as long if possible.  The fact that the style is context-dependent is also the reason why I generally avoid using tools that alter the formatting I used.

tools/rstlint.py
willingc: Thanks for the reviews and input. I can see your points on length. So, it looks like it comes down to ease of scanning or extra scrolling.

I'm going to go ahead and merge as in the PR. This is the first time that the file has been touched in 2+ years.

Issue #389, opened 2018-06-16
Add on new section about Translations  to 7. Documenting Python


willingc:
> There are now several official translations: see current doc pages, the bottom of https://devguide.python.org/experts/, and https://www.python.org/dev/peps/pep-0545/.  According to the pep, discussion is on doc-sig.  https://mail.python.org/mailman/listinfo/doc-sig

> 'third party' is not quite right, but there is no 'separate repository and issues tracker' selection.

> Carol, perhaps devguide 7. Documenting Python should have a short new section 7.6 Translations with at least the above info, better formatted.

See original message from bpo https://bugs.python.org/issue24978#msg319774

devarshigoswami:
If this hasn't been handled, Can I work on it?

za:
Maybe we can also close this issue cc: @ezio-melotti .

PR #391, opened 2018-06-21
Added 'Translations' section to '7. Documenting Python'

blossomica:
This is a fix for issue https://github.com/python/devguide/issues/389. Please feel free to let me know if any additional edits are needed.


documenting.rst
```diff
@@ -1529,3 +1529,9 @@ see the make targets above).
 .. _Sphinx: http://sphinx-doc.org/
 .. _virtualenv: https://virtualenv.pypa.io/
 .. _blurb: https://pypi.org/project/blurb/
+
+
+Translations
+============
+
+There are now several official documentation translations. See section :ref:`21.5. Documentation Translations <experts>` and :PEP:`545`.
```
terryjreedy: This line needs to be split into two lines.

The purpose of this addition is to properly direct people interested in helping with translations, and to give the rest of us a place to refer such people to.   I think the above should be followed by a link to the doc sig instead of leaving it buried in the PEP.  But Julien should confirm that the PEP is up to date or not and otherwise decide what this section should contain.

blossomica: Thanks for the review, @terryjreedy. I made some edits, based on your comments, feel free to let me know if you think there is anything else I should fix.

blossomica:
You're welcome!

PR #392, opened 2018-06-24
Clarify that backport PR needs to include GH- from master

Mariatta:
Backport PR needs to be in the form of `[X.Y] title (GH-NNNN)`, this is the pattern that bedevere uses:

https://github.com/python/bedevere/blob/master/bedevere/backport.py#L11

Adding a bookmark link, so I can reference it from a new status check proposed in https://github.com/python/core-workflow/issues/211#issuecomment-399727455


Mariatta:
Thanks!

PR #399, opened 2018-07-21
Updated setup guide to mention a dependency on .NET 3.5.

vsajip:


Mariatta:
Please keep the size-XS label to indicate that the changeset in this PR is small enough.
The first size-XS was added by a bot.

vsajip:
> Please keep the size-XS label

Sorry, I removed it because I thought it was added by accident. When I looked at the labels, `size-XS` had no description and I couldn't be sure what it meant. Of course I thought of the obvious interpretation from clothing labels, but didn't quite see how it applied to the PR :smile: If it had said e.g. `small-patch` I would probably have left it alone.

zooba:
Thanks, @vsajip!

Issue #390, opened 2018-06-19
Simplify developer log


brettcannon:
Instead of acting like a ledger, make it a list of active and emeritus members. Active members can be listed by the year they got commit privileges while emeritus is alphabetical and lists years active.

E.g.

## Active committers
- 2018
  1. Pablo
- 2003
  1. Brett Cannon
- 1989
  1. Guido van Rossum

## Emeritus committers
- Paul Du Bois (1998 - 2005)

Mariatta:
Great idea! I suggest "core developer" instead of "committer".

Stewori:
This proposal seems to wipe out a lot of information currently contained in the developer log. This is usually information about who recommended the person, but can also be scope information like "for work on the xy module", "for documentation work". Most relevantly it can indicate the person is working on an alternative PSF-owned Python implementation, e.g. "compatibility work for PyPy".

How would this information be preserved?

Mariatta:
So I realize I don't even know what is considered as "alternative PSF-owned Python implementation". Is there such list somewhere?

Stewori:
I don't know of an official list, but I think the PSFshould know what software is under their license. E.g. looking at Jython's license, the first paragraph reads

```
1. This LICENSE AGREEMENT is between the Python Software Foundation
("PSF"), and the Individual or Organization ("Licensee") accessing and
otherwise using this software ("Jython") in source or binary form and
its associated documentation.
```
If the license agreement is made between users and the PSF that indicates kind of ownership, no? (IANAL) It's also PSF for which the contributor agreement is signed.
In this sense, there is a [list in the wikipedia](https://en.wikipedia.org/wiki/Category:Software_using_the_PSF_license) but that's certainly not official.
Apparently that list does not contain PyPy and IronPython, so things might not be that easy. I always assumed they had the same license model as Jython, given that committers were usually included into dev log in the same fashion. However that's a question someone at the PSF or from PyPy or IronPython is better suited to answer.

berkerpeksag:
I agree with @Stewori.

Looking at PR #404, I think we are losing a lot of information and I don't think the new format is entirely accurate (because the section is now renamed to "Active core developers") For example, see "2014" section at https://github.com/python/devguide/pull/404/files#diff-fe3fd1dac75eba5fb6edcd98b012b444R55 Four of the listed names are not CPython core developers and one of them was added as a PEP editor (i.e. not a CPython or Jython core dev)

Mariatta:
[...]

PR #404, opened 2018-08-02
WIP - GH-390: Simplify the developer log

matrixise:


Mariatta:
I would prefer that Active Core Developer's section only list CPython core developers.

matrixise:
?? it's the case. 
if one person receives the commit flag, this person is a core dev.
I don't understand your sentence "Active Core Developer's section only list CPython core developers."

Could you explain?

Thank you

Mariatta:
Currently this list contains core developers of alternate Python implementation (IronPython, PyPy). I think we should keep it as a list of CPython core developers.

jimbaker:
I'm one of the Jython core devs. Although I have not been so active on Jython recently, except for occasional emails, I see three things that are objectionable about this proposal:

1. Jython has been historically involved in the development of the Python language, which of course is not the same as CPython. (CPython is the reference implementation of the Python language.) This makes sense, given that it is one of the two oldest projects in the Python community, the other being Mailman. Such aspects of the Python language as generators, coroutines, and optional static typing were expressly considered in the context of whether they could be readily implemented in Jython.

2. The CPython reference implementation contains the standard library for Python implementations.  Certain aspects of the standard library, such as requiring Python implementations where possible, are expressly designed for alternative implementations.

3. Recent proposed changes in governance, due to Guido van Rossum stepping down from his BDFL role, heighten the need for the recognized core developers of Python, regardless of implementation to continue to contribute this institutional knowledge as part of the further development of the Python language. We have only raised objections occasionally, and sometimes we have also helped prodded things along, but we believe our voice has been valuable in Python’s progress.

Work on Jython continues, albeit slowly with the small number of contributors we have. More importantly Jython remains useful to many users. There's also active work to bring Jython up to par with Jython 3, as well as very interesting work on JyNI to support the CPython extension API.

matrixise:
@Mariatta IronPython and PyPy, Armin Rigo (PyPy) was a core dev of CPython and Michael Foord is also a Core Dev.
What are the other ones?

brettcannon:
@Mariatta If you're talking about the folks who MvL gave commit rights to like @jimbaker  then they are core devs by the same process as the rest of us so there's nothing to tweak. We purposefully did that because (a) we knew all of them, and (b) they were contributing fixes upstream to the stdlib.

@jimbaker this list is for people who have been given commit privileges at some point on CPython. I'm sure Jython has its own list. But I don't think either team is going to cross-list the other implicitly as being a member of their development team, e.g. I'm not about to claim I'm a Jython team member just like I'm sure folks on the Jython team who don't have commit privileges on CPython are going to say they are a Python core dev.

ncoghlan:
I think most of my comments are more at the "What is the desired outcome?" level, so I posted those as a comment on https://github.com/python/devguide/issues/390#issuecomment-410424947

At a PR development level, the names for the "Active core developers" list need to come from https://github.com/orgs/python/teams/python-core/members

Everyone else in the developer log was either never a core developer (their CVS/SVN/hg access was for some other purpose and they never went beyond that original role), or else has since moved to emeritus status (either explicitly, by requesting that their commit access be removed, or implicitly, by failing to take the required steps to reacquire them following the move to GitHub)

jimbaker:
@brettcannon Fair enough, **explicit vs implicit** is an important -- and Pythonic -- distinction. So let’s keep it explicit, as we have done to date.

matrixise:
@ncoghlan I can't read https://github.com/orgs/python/teams/python-core/members (HTTP 404), maybe because I am not a core-dev myself. but with this list, we don't have the date when the developed has received the commit flag.

For this new list, should I only take the github user accounts? it seems weird because they have acquired the commit flag before Github. Ok, we use Github since 1y but on b.p.o, they have the flag.

brettcannon:
I might need to write a script that pulls the committer data from both b.p.o and GH and then correlates it in order to get a somewhat accurate list (plus there has been talk of storing all of this in TOML).

willingc:
@matrixise Do you want to update this PR?

aeros167:
@willingc Since it's been a while since this PR was updated, I created a new one that uses the same formatting, but includes recently added core developers and removes anyone not on the GitHub python-core team: https://github.com/python/devguide/pull/529.

Since I was recently added to the organization, I'm able to see the full listing of all members of the team.

PR #408, opened 2018-08-24
Update the example on how to checkout maintenance branch

Mariatta:
Clarify how to checkout from the remote.
Provide example using `upstream` instead of `origin`.


Mariatta:
Thanks!

PR #413, opened 2018-09-13
Restore table of recently end-of-lifed versions

jdufresne:
When discussing dropping EOL Python support in projects, referencing this table and its dates is useful. Gives perspective by easily identifying how long a Python version hasn't been supported.

The table was removed in commit 4eb2f5ce515db2edbf7720adac2fc247b4cdfd2e.

Refs #411


willingc:
Thanks for the PR @jdufresne and sharing that you find the table useful for motivating people to move to more recent Python 3 releases.

@ezio-melotti Perhaps we should add this table back, but remove it from the index page to another page. Thoughts?

ezio-melotti:
The reasons why I removed it are:
1. it is another table that needs to be updated manually and might end up containing outdated information;
2. it is in a different place that the other table (for good reasons, since it's not important enough to be on the index), so it's easier to miss and becoming outdated;
3. the index page already states that all the branches that are not listed reached end-of-life (even though it doesn't specify the dates);
4. I didn't expect many people to find it particularly useful :);

If it's added back, it should go to the devcycle.rst page, in particular at the bottom of the "End-of-life branches" section that I added in #411.  Under the other table in the index page, an rst comment should be added with a note about the other table (e.g. ".. remember to update the end-of-life table in devcycle.rst"), the status column should be removed, the comment column should be removed (as you did), a single link to the download page could be added under the table (as I did in the index page), I would rename the "Tag" column to "Branch", the Release Manager column could be added to mirror the other table.

To prevent the(se) table(s) to becoming outdated, an entry should be also added to PEP 101, stating that in case of a new major release or a minor release that marks the end-of-life of a branch, the devguide should be updated (or at least an issue about the update should be opened here).

willingc:
Thanks @ezio-melotti. I agree moving it off the index page is the best move. I do think there is value to have it as you suggest on the devcycle page.

@jdufresne Would you be willing to amend the PR to add @ezio-melotti's suggestion:
> It should go to the devcycle.rst page, in particular at the bottom of the "End-of-life branches" section that I added in #411. Under the other table in the index page, an rst comment should be added with a note about the other table (e.g. ".. remember to update the end-of-life table in devcycle.rst"), the status column should be removed, the comment column should be removed (as you did), a single link to the download page could be added under the table (as I did in the index page), I would rename the "Tag" column to "Branch", the Release Manager column could be added to mirror the other table.

Once the amended PR is merged, we should open an issue for PEP 101.

Thanks!

jdufresne:
Thanks for the feedback! I believe I have addressed all the points. Please let me know if you like to see any other changes or adjustments.

willingc:
Thanks @jdufresne for the contribution.

PR #431, opened 2018-10-26
Introduce `hub pr checkout` command

methane:
This is very comfortable tool when reviewing other's pull request.

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -330,6 +330,15 @@ local copy of a pull request as follows::
 
    git pr <pr_number>
 
+.. note::
+
+   [hub](https://github.com/github/hub) command line utility makes this
+   workflow very easy.  You can check out the branch by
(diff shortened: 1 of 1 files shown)


terryjreedy:
For me, the note is incomplete.  Installing git for Windows does not install a hub command.  What arguments follow 'hub push'?  Why is this better than 'git pr'?

willingc:
@terryjreedy The link for hub that @menthane shared directs to the installation page (README) for hub. For Windows, specifically, the install is at: https://github.com/github/hub#windows or installing via Powershell or as an executable.

Many Mac users prefer hub to the `git pr` alias as hub has additional commands used by maintainers (`git fetch <GitHub username` which sets up a remote and fetches and `git checkout <pr url>`). 

It's really a matter of choice to use hub or not. Feel free to suggest additional wording in a followup PR.

PR #432, opened 2018-10-30
Adding self to documentation experts.

JulienPalard:


willingc:
Seems reasonable to add as a new interest perhaps "Release Management-Docs" or add yourself to Release Management. The other place where it seems to make sense would be to add a sentence on the index.rst page that discusses releases.

JulienPalard:
Added myself to the release team line.

PR #441, opened 2018-12-14
Fix link on the git bootcamp page

lysnikolaou:
On the git bootcamp page, Markdown syntax was used for the link to [hub](https://github.com/github/hub), instead of rst syntax.


Mariatta:
Thanks!

PR #452, opened 2019-02-04
Provide some example on how to leave PR reviews

Mariatta:


Mariatta:
For documentation, sometimes it is still worth rendering the docs.

For real typos, at the very least, you can leave comment such as "assert is indeed the right spelling, not assret". You can also provide more context such as "I've confirmed that this is the only place where this typo happened".

pullrequest.rst
```diff
@@ -393,6 +393,22 @@ code and leave comments in the pull request or issue tracker.
...
+Instead of simply "approving" the pull request, leave comments, for example:
+
+1. Whether you have tested the PR
+
+2. Whether the PR fixes the linked bug/issue
+
+3. If you requested changes to the pull request, provide suggestions on how it
+   can be improved.
+
+4. Leave comments on what's "good" on the pull request, not just the "bad".
```
terryjreedy:
``` suggestion
4. Comment on what is "good" about the pull request, not just the "bad".
```
(While this can be overdone,) I think this is important. But I expect most people to not see why, or why they should bother.  So I think something more is needed.

When someone dismisses one of my IDLE improvements as 'useless', it is disheartening.  It shows that likely no effort was made to understand how it might benefit some users.  It also makes it hard to extract anything useful from the communication to make a better improvement.

My suggested addition: `Doing so will make it easier for the PR author to find the good in your comments.`

pullrequest.rst
```diff
@@ -393,6 +393,22 @@ code and leave comments in the pull request or issue tracker.
...
+---------------------------------------
+
+When you review a pull request, you should provide additional details and context
+of your review process. 
+
+Instead of simply "approving" the pull request, leave comments, for example:
+
+1. Whether you have tested the PR
+
+2. Whether the PR fixes the linked bug/issue
```
terryjreedy: I think these should be consolidated with an important item added.
```
`1. If you tested the PR, report the result and the system and version tested on, such as 'Windows 10', 'Ubuntu 16.4', or 'Mac High Sierra'.

File: pullrequest.rst [hunk 3]
```diff
@@ -393,6 +393,22 @@ code and leave comments in the pull request or issue tracker.
...
+
+3. If you requested changes to the pull request, provide suggestions on how it
```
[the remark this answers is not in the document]
Mariatta: Yep I would uppercase PR, but not pull request. Thanks. GitHub's help page about pull request also does not upper case the "pull request". https://help.github.com/articles/about-pull-requests/

Mariatta:
Thanks!

PR #457, opened 2019-02-06
Add entry for Jason R. Coombs based on his best recollection.

### LINKED ISSUE #455: No entry for  Jason R. Coombs in developers.rst
Hi,

I can see that Jason is a core-dev but I don't find any entry in developers.rst.

Thank you

jaraco:
Fixes #455.

### DIFF
--- a/developers.rst
+++ b/developers.rst
@@ -171,6 +171,9 @@ Permissions History
   Walker Hale were given push privileges on Mar 16 2011 by GFB, for
   contributions to the packaging module.
 
+- Jason R. Coombs was given commit access on Mar 14 2011
+  by Brett Cannon as part of a group sprinting on distutils2.
+
 - Jeff Hardy was given push privileges on Mar 14 2011 by BAC, for stdlib
   compatibility work for IronPython.
 


jaraco:
@brettcannon I vaguely recall that you were the approving member for the distutils2 working group at PyCon 2010. Or maybe it was 2011? I'm pretty sure it was Atlanta. Do you recall? Are there other records we can use to validate the declaration?

brettcannon:
I would just look at when the oldest commit was and which dates aligns the best. Once you do that then feel free to merge.

jaraco:
I don't believe I pushed any commits during that sprint, despite being granted access.

I do see my first major contribution in d40e6f70a5, which was committed as part of bpo-1578269. My first direct commit seems to be fee7745ebb, which would have been after PyCon 2011. So maybe that was the correct year. At the very least, we know it could not have been later than that. I'll update the PR accordingly.

jaraco:
This renewed commit assumes that PyCon 2011 was the place where the change was granted. That aligns better with the history of commits.

Issue #458, opened 2019-02-06
Add git config alias for devguide


eamanu:
Hello everyboy!

I am not sure if here is the appropriate  place to this issue. Sorry if not.

On gitbootcamp.rst there is a recommendation to made easy the creation of a branch for PR to test it. I am talking about this: 
```
$ git config --global alias.pr '!sh -c "git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}" -' 
```

What about if is added something similar for maintain updated the CPython repo. I am thinking about something have (like result)

```
$ git update
```

To avoid:

```
$ git checkout master
$ git pull origin master
$ git pull upstream master
$ git push origin master
```

What do you think about that?

Thanks!

Mariatta:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/python/devguide/issues/61 (devguide-git bot), https://github.com/python/devguide/issues/8 (Add Quick Start: Communications section to devguide), https://github.com/python/devguide/issues/9 (Add tips for effective online communication to devguide), https://github.com/python/devguide/issues/217 (Quoting for config alias.pr), and https://github.com/python/devguide/issues/7 (Add a Sprints organization/preparation section to devguide).

dorothykiz1:
Hello @eamanu and @Mariatta  i would like to work on this issue

asvetlov:
Do you need to update your local `master` copy?
If the master is tracked against `upstream` and `git checkout master && git pull` fetches `upstream` -- there is no need to keep your local (origin) master up to date.
Actually, you can even remove the branch -- everything still works fine.

eamanu:
Hi @asvetlov, 

Thanks for you response
> Do you need to update your local `master` copy?
hmmm I think yes. Because this way you can create a branch from master to make the PR. If you have an old master branch maybe when you made the PR this can have some conflicts. This is for local (your computer) copy. 

Maybe could not be necessary have and updated copy on my own cpython fork (master branch) on Gihub, but if you follow the devguide, this mentions on https://devguide.python.org/gitbootcamp/#syncing-with-upstream the `push to origin master`.

eamanu:
For that reason I create the alias.

matrixise:
I use these ones:

pr = !sh -c \"git fetch upstream pull/${1}/head:pr_${1} && git checkout pr_${1}\" -
[...]

PR #459, opened 2019-02-07
Add alias for sync local and origin master branch with upstream

eamanu:
Hi! 

This is a proposal for #458. I added the setting up of `update` alias to automate the syncing of both local and origin master branch with upstream master branch. Thanks!

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -247,6 +247,16 @@ Solution::
 The ``--rebase`` option is only needed if you have local changes to the
 branch.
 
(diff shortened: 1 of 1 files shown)


willingc:
Hi @eamanu. Thanks for the PR. Instead of giving the example alias (which can be problematic if folks have named their remotes differently), I think it would be better to mention that git aliases can be set up for common tasks. Also a link to the [git doc](https://git-scm.com/book/en/v2/Git-Basics-Git-Aliases) would be helpful.

PR #462, opened 2019-02-11
Mention that news file should be at least 30 characters

Mariatta:
In the past we've seen behavior where people entered an empty news file just to make the "check" news" status pass.
In https://github.com/python/bedevere/issues/127 we've added a restriction that the news entry should be at least 30 characters long.


ned-deily:
I don’t think we need to have such a large requirement. We now have an example of a commit where fewer characters is just fine. In the end it’s up to the release manager to review and edit all news items prior to release. I’d prefer a non-zero test.

Mariatta:
Closing this PR since it's been decided that we shouldn't limit the news file length. Just as long as it is not empty. Thanks.

PR #470, opened 2019-03-17
Remove EOLed Python 3.4 from Status of Python branches

cclauss:
Fixes: [issue36327] Remove EOLed Py34 from "Status of Python branches"

@larryhastings

Remove Python 3.4 https://devguide.python.org/#branchstatus

Add Python 3.4 https://devguide.python.org/devcycle/#end-of-life-branches

----------
assignee: docs@python
components: Documentation
messages: 338131
nosy: cclauss, docs@python
priority: normal
severity: normal
status: open
title: Remove EOLed Py34 from "Status of Python branches"
versions: Python 3.4

_______________________________________
Python tracker <report@bugs.python.org>
<https://bugs.python.org/issue36327>
_______________________________________


gvanrossum:
Is there a definition of EoL somewhere? Does it coincide with the last release date? Or is it some time later?

terryjreedy:
https://devguide.python.org/devcycle/#end-of-life-branches says
 "The code base for a release cycle which has reached end-of-life status is frozen and no longer has a branch in the repo. The final state of the end-of-lifed branch is recorded as a tag with the same name as the former branch, e.g. 3.3 or 2.6."

The branch to tag conversion should not happen until five years are up.  The last release may happen before that.  For 3.3, the last release was 10 days before.  For 3.2, over a year befor.e.   The last release and EOL can past the 5-year date.  I believe Larry Hastings is planning a 3.4.10 any time now, but it is not yet available on python.org and so not 'released'.

I think we can let Larry decide whether to close this or revise after the release and tag conversion.

cclauss:
#472 seems to be making similar changes.

PR #471, opened 2019-03-19
Python 3.4 reached its end-of-life

vstinner:
https://mail.python.org/pipermail/python-dev/2019-March/156705.html


vstinner:
@larryhastings: Is it ok now to declare that Python 3.4 reached its end of life?

vstinner:
Ah, PR #472 is better than my PR. I close mine.

vstinner:
I was confused. My PR modify another page. I merged it.

PR #477, opened 2019-05-04
Update Black URL

jdufresne:
> Black, your uncompromising #Python code formatter, was a project
> created with the community in mind from Day 1. Today we moved it under
> the PSF umbrella. It's now available on GitHub under
> https://github.com/python/black/ . You install and use it just like
> before.

https://twitter.com/llanga/status/1123980466292445190

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ The CPython Developer's Guide
    :target: https://python.zulipchat.com
 
 .. |Codestyle| image:: https://img.shields.io/badge/code%20style-black-000000.svg
-   :target: https://github.com/ambv/black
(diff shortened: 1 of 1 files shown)


gvanrossum:
Please hold off merging this until the SC debate about this is resolved.

hugovk:
I think the SC debate is now resolved, see https://github.com/python/steering-council/issues/9#issuecomment-490519819 and https://github.com/python/devguide/pull/480, but I'm not clear on the decision for Black.

gvanrossum:
We're waiting for Black to move.

jdufresne:
Rebased and updated to now point to https://github.com/psf/black

PR #481, opened 2019-05-10
Add RTD configuration file

willingc:
Create a config file to enable more people to edit settings without access to the RTD GUI.

[Reference: RTD documentation of configuration file](https://docs.readthedocs.io/en/stable/config-file/index.html#configuration-file)

[Rendered docs on my test account](https://test-devguide.readthedocs.io/en/rtd-build/)

PR #484, opened 2019-05-12
Revert "Add RTD configuration file"

Mariatta:
Reverts python/devguide#481

Seems like devguide wasn't building properly.
I'm currently on my phone so I can't do much.
Reverting until we can investigate more.


willingc:
Thanks @Mariatta 👍 I've fixed all through the RTD UI. See issue comments.

PR #472, opened 2019-03-19
Add 3.4 to list of EOL'd versions

crwilcox:
3.4 EOL'd. This PR adds it to the list of EOL'd in the devguide.

PR #486, opened 2019-05-20
Remove EOLed Python 3.4 from Status of Python branches

cclauss:
Fixes: [issue36327] Remove EOLed Py34 from "Status of Python branches"  Completes the cycle of #470, #471, and #472.

Another attempt at the second half of #470 now that a few months have passed since CPython v3.4.10 was released.  @larryhastings Your review please.


willingc:
Thanks for the PR @cclauss. This looks like it needs a rebase to correct the current conflict. I will add Larry as a reviewer since he should have the final call on when he wishes to remove from devguide. Thanks.

hugovk:
3.4 has since been removed from the table:

https://devguide.python.org/#status-of-python-branches

PR #488, opened 2019-05-28
Replace  --force with --force-with-lease

maxking:
`--force-with-lease` is generally less harmful than `--force`  and is usually what people want when they use `--force`.


Mariatta:
Thanks!

PR #480, opened 2019-05-08
Document the policy for repositories under the Python org

brettcannon:

PR #489, opened 2019-05-29
Fix number

hugovk:
Two categories were added during review of https://github.com/python/devguide/pull/480. 

Don't specify the number, so it doesn't need to be updated when categories are added or removed.


encukou:
Thank you!
Normally I wouldn't dare modify a Steering Committee pronouncement, but the fix is obvious, so I'll go ahead and merge.

PR #497, opened 2019-06-16
Add introductory note for new and experienced git users

willingc:
Follow up to #496


willingc:
Thanks for the review @methane.

PR #499, opened 2019-06-18
Add p-ganssle to experts list for times and dates

pganssle:
Adding myself to the experts index.

Does anyone know if the auto-nosy index on bpo is populated from this document automatically, or is there somewhere else I need to update for that?


Mariatta:
I think it needs to be done in bpo side. One of the bpo admins can do it. (not me)

zware:
This will propagate to bpo automatically.

(Also, welcome, Paul!)

PR #504, opened 2019-07-03
Update my affiliations and motivations

pitrou:


pitrou:
I realized those weren't up-to-date anymore :-)

Issue #507, opened 2019-07-12
Add github usernames to the experts index


aeros:
On the [experts index section of the devguide](https://devguide.python.org/experts), it would be helpful to include the GitHub usernames of the experts in each category. This seems like it would be a necessary component of fully implementing [PEP 581](https://www.python.org/dev/peps/pep-0581/). Currently, the names on this list only correspond with the names on the [roundup issue tracker](https://bugs.python.org/).

This would be particularly helpful so that contributors and those in the [proposed triager role](https://discuss.python.org/t/proposal-create-bug-triage-team-on-github/992) can notify the correct the core developers for review of an issue. Sometimes the reviewers automatically added based on the code owners list is enough, but in many cases (particularly for documentation) it is not. For an example, see https://github.com/python/cpython/pull/14492.

I would be glad to help with updating the experts list, but I'm undecided as far as formatting goes. At the moment, it looks something like this:

>**Interest Area** | **Maintainers**

>documentation | ezio.melotti, eric.araujo, mdk, willingc

Here's a few ideas:

1) **Add github username in parenthesis if it is not the same as the one listed**:

>**Interest Area** | **Maintainers**

>documentation | ezio.melotti (ezio-melotti), eric.araujo (merwok), mdk (JulienPalard) , willingc

2) **Change the maintainers column to use their actual name, and have other section(s) for the usernames**:

>**Interest Area** | **Maintainers** | **GitHub**

>documentation | Ezio Melotti, Eric Araujo, Julien Palard, Carol Willing | ezio-melotti, merwok JulienPalard, willingc

3) **Replace roundup usernames with GitHub usernames**:

>**Interest Area** | **Maintainers**

>documentation | ezio-melotti, merwok, JulienPalard, willingc

Since the transition to GitHub issues is still in progress, I would suggest using the first option, as that would allow for the index to still be useful for issues which are on the roundup issue tracker. Once the transition is complete, the roundup usernames could be removed and replaced with just the github usernames (third option).

aeros167:
While working on this issue, it would also be appropriate to spend some time updating the list so that any currently inactive experts are removed and currently active experts are added. Any active core developer who has experience reviewing PRs in a given category and is interested could be added.

Mariatta:
I also prefer option 1. 

I believe there is an existing mechanism to automatically `+nosy` people in experts list to issue in bpo. Not sure if changing the format will break anything. Perhaps someone more familiar with bpo can chime in.

zware:
[...]

PR #517, opened 2019-07-27
Add JSON for storing expert data

aeros:
Component of issue #507

@zware and @Mariatta suggested that an effective method of storing the data from the experts index would be to add a new JSON file to the devguide repository. This would allow for an efficient way to store the expert data and to easily be able to find their respective GitHub names (or any other platform). Currently, `experts.rst` only lists the names on bpo, and the formatting can't be changed due to the nosy list scripts using data from the tables. As a beneficial side effect, this would also likely improve the performance of any expert data extraction scripts.

The formatting of the JSON file is using two primary sections: `categories` and `maintainers`. Initially, I was going to leave the exact formatting from the `experts.rst` tables, but I realized that doing so would result in quite a lot of redundant data. To address that issue, I used the following format:

```javascript
'categories': {
    'category_name': [
        'maintainer_name',
        // ...,
   ],
   // ...,
},
'maintainers': {
    'maintainer_name' : {
        'bpo_name': 'value',
        'github_name': 'value',
        // Other properties?
    },
    // ...,
}
```

Having a separate section for the specific information on each of the maintainers provides normalization, so that the information for any maintainers that are experts for multiple modules aren't repeatedly stored. 

I think that providing their actual name instead of their bpo username makes a bit more sense, since that allows `experts.rst` to be far less platform specific. It also makes more sense for the platform specific usernames to be a property of their actual name, rather than using the bpo username as the key. Additionally, since we're moving away from bpo in the near future, that adds further reason to not use it as the identifying key for the experts.

An alternative would be to instead use the GitHub usernames as the key, but as I mentioned earlier, I find it preferable that this JSON doesn't favor any particular platform. Leaving it platform-independent makes it far more scalable in the long term. Let me know if anyone has any suggestions for alternative formats.

Note: The JSON file is currently missing a few experts (that aren't core devs), since I extracted data from the bpo committers list to convert from their bpo usernames to their actual names. Feel free to add or mention any active experts that are missing. The current experts list has an (inactive) after the names of some of the inactive people, but I'm not sure as to whether or not it's fully up to date.

### DIFF
--- a/experts.json
+++ b/experts.json
@@ -729,9 +729,6 @@
         "Argument Clinic":[
             "Larry Hastings"
         ],
-        "Platform":[
-
-        ],
         "AIX":[
 
         ],
@@ -767,9 +764,6 @@
             "Paul Moore"
         ],
         "JVM/Java":[
-
-        ],
-        "Interest Area":[
 
         ],
         "algorithms":[
@@ -934,9 +928,6 @@
         "version control":[
             "Éric Araujo",
             "Ezio Melotti"
-        ],
-        "Translation":[
-
         ],
         "French":[
             "Julien Palard"
--- a/experts.json
+++ b/experts.json
@@ -1,0 +1,1284 @@
+{
+    "categories":{
+        "__future__":[
+
+        ],
+        "__main__":[
+            "Guido van Rossum",
+            "Nick Coghlan"
+        ],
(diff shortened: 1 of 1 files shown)


aeros:
Closing this PR as the core issue was already solved with https://github.com/python/devguide/pull/552.

PR #524, opened 2019-08-11
Improve gitbootcamp.rst

maggyero:
This PR will apply the following changes to gitbootcamp.rst:

- fix capitalization in some section titles;
- use the long form of will consistently;
- use variables `<x>` consistently.

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -40,7 +40,7 @@ You will only need to do this once.
 
 .. _clone-your-fork:
 
-Cloning your Forked CPython Repository
+Cloning a Forked CPython Repository
 --------------------------------------
 
 You will only need to do this once.  From your command line::
@@ -157,7 +157,7 @@ Staging and Committing Files
 
 2. To stage the files to be included in your commit::
 
-      git add <filename>
+      git add <filename1> <filename2>
 
 3. To commit the files that have been staged (done in step 2):
 
@@ -201,7 +201,7 @@ Commit the files:
 
 .. code-block:: bash
 
-   git commit -m "<message>"
+   git commit -m <message>
 
 .. _push-changes:
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -201,7 +201,7 @@ Commit the files:
 
 .. code-block:: bash
 
-   git commit -m <message>
+   git commit -m "<message>"
 
 .. _push-changes:
 
@@ -469,7 +469,7 @@ To edit an open pull request that targets ``master``:
       git fetch upstream
       git merge upstream/master
       git add <filename>
-      git commit -m <message>
+      git commit -m "<message>"
 
 4. Push the changes back to the contributor's PR branch::
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -469,7 +469,7 @@ To edit an open pull request that targets ``master``:
       git fetch upstream
       git merge upstream/master
       git add <filename>
-      git commit -m "<message>"
+      git commit -m <message>
 
 4. Push the changes back to the contributor's PR branch::
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -107,7 +107,7 @@ Create a new branch and switch to it::
 
 This is equivalent to::
 
-   # create a new branch off master, without checking it out
+   # create a new branch from master, without checking it out
    git branch <branch-name> master
    # check out the branch
    git checkout <branch-name>
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -41,7 +41,7 @@ You will only need to do this once.
 .. _clone-your-fork:
 
 Cloning a Forked CPython Repository
---------------------------------------
+-----------------------------------
 
 You will only need to do this once.  From your command line::
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -77,7 +77,7 @@ Setting Up Your Name and Email Address
 
 .. code-block:: bash
 
-   git config --global user.name <name>
+   git config --global user.name "<name>"
    git config --global user.email <email>
 
 The ``--global`` flag sets these parameters globally while
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -18,7 +18,7 @@ Git Bootcamp and Cheat Sheet
    desire a different approach.
 
 
-In this section, we'll go over some commonly used Git commands that are
+In this section, we will go over some commonly used Git commands that are
 relevant to CPython's workflow.
 
 .. contents::
@@ -28,7 +28,7 @@ relevant to CPython's workflow.
 Forking CPython GitHub Repository
 ---------------------------------
 
-You'll only need to do this once.
+You will only need to do this once.
 
 1. Go to https://github.com/python/cpython.
 
@@ -36,18 +36,18 @@ You'll only need to do this once.
 
 3. When asked where to fork the repository, choose to fork it to your username.
 
-4. Your fork will be created at https://github.com/<username>/cpython.
+4. Your forked CPython repository will be created at https://github.com/<username>/cpython.
 
 .. _clone-your-fork:
 
-Cloning The Forked CPython Repository
--------------------------------------
+Cloning your Forked CPython Repository
+--------------------------------------
 
-You'll only need to do this once.  From your command line::
+You will only need to do this once.  From your command line::
 
    git clone git@github.com:<username>/cpython.git
 
-It is also recommended to configure an ``upstream`` remote::
+It is also recommended to configure an ``upstream`` remote repository::
 
    cd cpython
    git remote add upstream git@github.com:python/cpython.git
@@ -61,11 +61,11 @@ To list the remote repositories that are configured, along with their URLs::
 
    git remote -v
 
-You should have two remotes: ``origin`` pointing to your fork,
+You should have two remote repositories: ``origin`` pointing to your forked CPython repository,
 and ``upstream`` pointing to the official CPython repository::
 
-   origin  git@github.com:<your-username>/cpython.git (fetch)
-   origin  git@github.com:<your-username>/cpython.git (push)
+   origin  git@github.com:<username>/cpython.git (fetch)
+   origin  git@github.com:<username>/cpython.git (push)
    upstream        git@github.com:python/cpython.git (fetch)
    upstream        git@github.com:python/cpython.git (push)
 
@@ -77,18 +77,18 @@ Setting Up Your Name and Email Address
 
 .. code-block:: bash
 
-   git config --global user.name "Your Name"
-   git config --global user.email email@example.org
+   git config --global user.name <name>
+   git config --global user.email <email>
 
-The ``--global`` flag sets these globally,
-``--local`` sets them only for the current project.
+The ``--global`` flag sets these parameters globally while
+the ``--local`` flag sets them only for the current project.
 
 .. _autocrlf:
 
 Enabling ``autocrlf`` on Windows
 --------------------------------
 
-The *autocrlf* option will fix automatically any Windows-specific line endings.
+The ``autocrlf`` option will fix automatically any Windows-specific line endings.
 This should be enabled on Windows, since the public repository has a hook which
 will reject all changesets having the wrong line endings::
 
@@ -107,7 +107,7 @@ Create a new branch and switch to it::
 
 This is equivalent to::
 
-   # create a new branch off 'master', without checking it out
+   # create a new branch off master, without checking it out
    git branch <branch-name> master
    # check out the branch
    git checkout <branch-name>
@@ -132,7 +132,6 @@ on the 2.7 release from the ``upstream`` remote::
 
    git checkout -b 2.7 upstream/2.7
 
-
 .. _deleting_branches:
 
 Deleting Branches
@@ -149,7 +148,6 @@ To delete a **remote** branch::
 
 You may specify more than one branch for deletion.
 
-
 Staging and Committing Files
 ----------------------------
 
@@ -159,7 +157,7 @@ Staging and Committing Files
 
 2. To stage the files to be included in your commit::
 
-      git add path/to/file1 path/to/file2 path/to/file3
+      git add <filename>
 
 3. To commit the files that have been staged (done in step 2):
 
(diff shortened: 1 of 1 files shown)


gitbootcamp.rst
```diff
@@ -18,7 +18,7 @@ Git Bootcamp and Cheat Sheet
    desire a different approach.
 
 
-In this section, we'll go over some commonly used Git commands that are
+In this section, we will go over some commonly used Git commands that are
```
aeros167: Personally, I don't think there's an issue with using contractions, only that within a given article, the form chosen for a term should be consistent. Since ``gitbootcamp`` uses ``we'll``, it should not use ``we will``. However, either option is okay.

gitbootcamp.rst
```diff
@@ -28,26 +28,26 @@ relevant to CPython's workflow.
...
 3. When asked where to fork the repository, choose to fork it to your username.
 
-4. Your fork will be created at https://github.com/<username>/cpython.
+4. Your forked CPython repository will be created at https://github.com/<username>/cpython.
 
 .. _clone-your-fork:
 
-Cloning The Forked CPython Repository
--------------------------------------
+Cloning your Forked CPython Repository
```
aeros167: Generally, it is preferable to avoid the usage of second and third person pronouns in technical writing. It would be outside of the scope of this PR to correct every usage, but it would be preferable to avoid adding more instances. If/when this PR is merged, I can make the other corrections in a separate PR.
```suggestion
Cloning a Forked CPython Repository
```

gitbootcamp.rst
```diff
@@ -77,18 +77,18 @@ Setting Up Your Name and Email Address
...
 
-   git config --global user.name "Your Name"
-   git config --global user.email email@example.org
+   git config --global user.name <name>
+   git config --global user.email <email>
 
-The ``--global`` flag sets these globally,
-``--local`` sets them only for the current project.
+The ``--global`` flag sets these parameters globally while
+the ``--local`` flag sets them only for the current project.
```
aeros167: Since ``--global`` was explicitly mentioned as a flag earlier within the same sentence, it might be a bit redundant to do the same for ``--local``. I agree with the addition of "parameters" though.

gitbootcamp.rst
```diff
@@ -107,7 +107,7 @@ Create a new branch and switch to it::
 
 This is equivalent to::
 
-   # create a new branch off 'master', without checking it out
+   # create a new branch off master, without checking it out
```
aeros167: I agree with removing the quotations, but I think this provides a further improvement to the wording:
```suggestion
   # create a new branch from master, without checking it out
```

gitbootcamp.rst
```diff
@@ -107,7 +107,7 @@ Create a new branch and switch to it::
...
    git branch <branch-name> master
    # check out the branch
```
aeros167: This suggestion is along the same lines of making wording more explicit. Also, the usage of "checkout" is more appropriate here than "check out", and more closely matches the Git command.
```suggestion
   # checkout the newly created branch
```

gitbootcamp.rst
```diff
@@ -159,28 +157,26 @@ Staging and Committing Files
 
 2. To stage the files to be included in your commit::
 
-      git add path/to/file1 path/to/file2 path/to/file3
+      git add <filename>
```
aeros167: I agree with the usage of ``<filename>`` instead of the current formatting, but I think it is also helpful to mention that multiple files can be added at once.
```suggestion
      git add <filename1> <filename2>
```

gitbootcamp.rst
```diff
@@ -205,51 +201,48 @@ Commit the files:
...
+2. Press the ``New pull request`` button.
 
-3. Click ``compare across forks`` link.
+3. Click the ``compare across forks`` link.
 
-4. Select the base fork: ``python/cpython`` and base branch: ``master``.
+4. Select the base repository: ``python/cpython`` and base branch: ``master``.
 
-5. Select the head fork: ``<username>/cpython`` and base branch: the branch
+5. Select the head repository: ``<username>/cpython`` and head branch: the branch
```
aeros167: For step 5, this usage of "fork" is appropriate. I don't think this change is required.

gitbootcamp.rst
```diff
@@ -205,51 +201,48 @@ Commit the files:
...
    containing your changes.
 
-6. Press ``Create Pull Request`` button.
-
+6. Press the ``Create pull request`` button.
 
-Syncing With Upstream
+Syncing with Upstream
```
aeros167: The capitalization of "with" in a title is somewhat subjective and depends on the particular styling guidelines, but the most common rule that I've encountered is to lowercase any prepositions which are four or less letters. So, using lower case "with" is probably more appropriate.

gitbootcamp.rst
```diff
@@ -205,51 +201,48 @@ Commit the files:
 
 .. code-block:: bash
 
-   git commit -m '<message>'
-
+   git commit -m "<message>"
```
aeros167: The quotation marks aren't needed here, since it is already using <>.
```suggestion
   git commit -m <message>
```

gitbootcamp.rst
```diff
@@ -262,8 +255,9 @@ Another scenario:
 - You created ``some-branch`` some time ago.
 - Time passes.
 - You made some commits to ``some-branch``.
-- Meanwhile, there are recent changes from upstream CPython repository.
-- You want to incorporate the recent changes from upstream into ``some-branch``.
+- Meanwhile, there are recent changes from the upstream CPython repository.
+- You want to incorporate the recent changes from the upstream CPython
```
aeros167: See above suggestion.
```suggestion
- You want to incorporate the recent changes from the upstream
```

gitbootcamp.rst
```diff
@@ -205,51 +201,48 @@ Commit the files:
...
-
+   git commit -m "<message>"
 
 .. _push-changes:
 
 Pushing Changes
 ---------------
 
-Once your changes are ready for a review or a pull request, you'll need to push
+Once your changes are ready for a review or a pull request, you will need to push
```
aeros167: See above comment regarding contractions.

gitbootcamp.rst
```diff
@@ -28,26 +28,26 @@ relevant to CPython's workflow.
...
 
 .. _clone-your-fork:
 
-Cloning The Forked CPython Repository
--------------------------------------
+Cloning your Forked CPython Repository
+--------------------------------------
 
-You'll only need to do this once.  From your command line::
+You will only need to do this once.  From your command line::
```
aeros167: See above comment.

gitbootcamp.rst
```diff
@@ -28,26 +28,26 @@ relevant to CPython's workflow.
 Forking CPython GitHub Repository
 ---------------------------------
 
-You'll only need to do this once.
+You will only need to do this once.
```
aeros167: See above comment.

gitbootcamp.rst
maggyero: Actually gitbootcamp.rst was using both.

gitbootcamp.rst
maggyero: See above comment.

gitbootcamp.rst
maggyero: See above comment.

gitbootcamp.rst
maggyero: I agree.

gitbootcamp.rst
maggyero: I disagree here, I think that comments should not necessarily follow the command names but rather be correct English, so the infinitive "check out" or "switch to".

gitbootcamp.rst
maggyero: Okay.

gitbootcamp.rst
maggyero: Okay.

gitbootcamp.rst
maggyero: See above comment.

gitbootcamp.rst
maggyero: Yes.

gitbootcamp.rst
maggyero: To me it the asymmetry sounded odd.

gitbootcamp.rst
maggyero: I agree.

gitbootcamp.rst
maggyero: I disagree as GitHub uses exactly "base repository" and "head repository".

maggyero:
@aeros167 Thanks for the review. I have applied some of your suggestions.

gitbootcamp.rst
aeros167: @maggyero Oh, interesting. In that case, we should probably use whichever form was encountered first and stick with it.

gitbootcamp.rst
aeros167: Generally speaking, code comments have a tendency to be significantly less restricted by proper English. However, both "checkout" and "check out" are probably acceptable in this case.

gitbootcamp.rst
```diff
@@ -28,26 +28,26 @@ relevant to CPython's workflow.
...
 
 1. Go to https://github.com/python/cpython.
 
 2. Press ``Fork`` on the top right.
 
 3. When asked where to fork the repository, choose to fork it to your username.
 
-4. Your fork will be created at https://github.com/<username>/cpython.
+4. Your forked CPython repository will be created at https://github.com/<username>/cpython.
```
eamanu: What about :
```suggestion
4. Your CPython repository forked will be created at https://github.com/<username>/cpython.
```

gitbootcamp.rst
```diff
@@ -28,26 +28,26 @@ relevant to CPython's workflow.
...
 
 .. _clone-your-fork:
 
-Cloning The Forked CPython Repository
--------------------------------------
+Cloning a Forked CPython Repository
+--------------------------------------
```
eamanu:
```suggestion
-----------------------------------
```

gitbootcamp.rst
```diff
@@ -205,51 +201,48 @@ Commit the files:
 
 .. code-block:: bash
 
-   git commit -m '<message>'
-
+   git commit -m <message>
```
eamanu: I think this that is better used ```' '``` because when you write a message to the commit you need use ```' '```

gitbootcamp.rst
```diff
@@ -480,10 +469,10 @@ To edit an open pull request that targets ``master``:
       git fetch upstream
       git merge upstream/master
       git add <filename>
-      git commit -m "<commit message>"
+      git commit -m <message>
```
eamanu: Same comment above. I think that the double quote is necessary because when you wirte
a commit message, tue ```"```` is necessary

gitbootcamp.rst
aeros167: This is not proper grammar:
> Your CPython repository forked ...

"Your forked CPython repository" is more correct, but an alternative could be:

> Your fork of the CPython repository

gitbootcamp.rst
aeros167: Good catch, the dashes in the title need to have the same character count as the text of the title.

gitbootcamp.rst
aeros167: Oops, I was the one who initially suggested to remove the quotations as it initially looked like they were a styling decision. I forgot they were required.

gitbootcamp.rst
aeros167: Agreed, but since both the single quote ``'`` and double quote ``"`` are both acceptable, we should use the single quote since it was used in the previous one for consistency.

gitbootcamp.rst
maggyero: Thanks @eamanu.

gitbootcamp.rst
maggyero: Okay, but let's use the double quotes as they allow the apostrophe punctuation mark in the commit message, and as they were already used in the two other `git commit` commands of this document. Done.

gitbootcamp.rst
maggyero: Okay, but let's use the double quotes as they allow the apostrophe punctuation mark in the commit message, and as they were already used in the two `git commit` commands of this document. Done.

gitbootcamp.rst
eamanu: Sorry for the delay of me response. 

> This is not proper grammar:

Sorry for the noise I am not English speaker, I believe that was a good change. Thanks and sorry.

aeros167: > Sorry for the noise I am not English speaker, I believe that was a good change. Thanks and sorry.

No problem! It's always better to try to better to provide feedback than not submit any at all. It's appreciated either way. (:

methane:
> * use variables `<x>` consistently.

I don't think this is a good idea.  Examples helps readers than variables sometime.

For example, `"Your Name"` looks more clear than `"<your name>"`.  It is easy to understand that I should write "Inada Naoki", not "<inada naoki>" or "methane".

aeros167:
> I don't think this is a good idea. Examples helps readers than variables sometime.

> For example, "Your Name" looks more clear than "<your name>". It is easy to understand that I should write "Inada Naoki", not "" or "methane".

In that case, I think we should aim to consistently use one format or the other within the article, as switching between the two could cause some degree of confusion. I'm okay with using quotes instead of tags though.

Another alternative would be to use italics, that format is used commonly in a number of different areas of documentation to tell the reader that they should insert their own value:

"*Your Name*"

methane:
> In that case, I think we should aim to consistently use one format or the other within the article, as switching between the two could cause some degree of confusion.

I don't think it causes real confusion.  Variables are useful when the same word is used several times.  But it is not so useful when it is used only once like `<name>` or `<email>`.

As I described above, `"Your Name"` will have less confusion than `"<name>"` or `<name>`.

aeros167:
@methane 
> I don't think it causes real confusion. Variables are useful when the same word is used several times. But it is not so useful when it is used only once like ``<name>`` or ``<email>``.

Okay, I see. As an example, do you think it's useful in the following context? (taken from an existing section in gitbootcamp):
```
git clone git@github.com:<username>/cpython.git
```

> As I described above, "Your Name" will have less confusion than ``"<name>"`` or ``<name>``.

I agree, I just thought that the italics might be useful, such as "*Your Name*". It helps to make the text stand out to the reader, and indicate they should substitute their own text there. In the specific case of "Your Name" it's fairly clear either way, it just makes it more obvious.

methane:
> Okay, I see. As an example, do you think it's useful in the following context for `<username>`? (taken from an existing section in gitbootcamp):

Yes.  `<username>` is used above (L48):

```
   git clone git@github.com:<username>/cpython.git
```

So `<your-username>` should be replaced with `<username>` for consistency.
This is the case consistency is important for readers.

It may be more helpful to add more comment like this:

```
   # Replace <username> with your github user name.
   git clone git@github.com:<username>/cpython.git
```

maggyero:
@methane Alright, updated, thanks for the review.

PR #528, opened 2019-08-23
Contribute article to the exploring page

tonybaloney:
Adds another article to the list of external references to explore the source code.

https://realpython.com/cpython-source-code-guide/

Article is based on 3.8.0b3


ncoghlan:
Thanks! I was thinking we should add this after you posted it, so it's nice to be able to just merge the PR :)

PR #530, opened 2019-08-24
Add guidelines for triagers closing PRs

aeros167:
Based on the conversation in https://github.com/python/devguide/issues/527, I added a note after the "Responsibilities include" list which specifies that triagers should only close PRs that could be considered spam, but they can suggest for other PRs to be closed. For anything potentially valid (including recently active and older PRs), a core developer should make the final decision.

PR #531, opened 2019-08-28
Adding Stale label

eamanu:
In this PR I propose add the stale label on devguide.

Also, move OS-X label to above for a correct alphabetic order.

I would like document the ```awaiting *``` labels, but I don't know if in this section (Github label for PR) is the correct. Some opinion about it?

### DIFF
--- a/triaging.rst
+++ b/triaging.rst
@@ -131,7 +131,7 @@ sprint
 
 stale
     Used for stale PRs or inactive for a long period of time. This label
-    help to core dev identify quickly the PRs candidate to be closed or make
+    help to core dev identify quickly the PR candidate to be closed or make
     a ping to the author.
 
 type-bugfix
--- a/triaging.rst
+++ b/triaging.rst
@@ -131,7 +131,7 @@ sprint
 
 stale
     Used for stale PRs or inactive for a long period of time. This label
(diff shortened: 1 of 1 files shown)


eamanu:
> No objections to me, we just have to make sure to actually make the label if this gets merged. :)

Hi @brettcannon thanks for the review. I think the label is already created https://github.com/python/cpython/labels

eamanu:
Hi @aeros167, thank you for the review.  

> Thanks for the PR @eamanu! I recently added a section for the stale label to my PR (#530), but I removed it since it would probably be better to do it a separate PR (as you had done).

I would like to document the ```awaiting-* ``` labels. But I don't know if here is the correct sections. What do you think.

```
stale
    Used for PRs that include changes which are no longer relevant or when the
    author hasn't responded to feedback in a long period of time. This label
    helps core developers quickly identify PRs that are candidates for closure 
    or require a ping to the author.
```

I like how sound this, so I will add the suggestion. Thanks

aeros167:
@eamanu:
> I would like to document the awaiting-* labels. But I don't know if here is the correct sections. What do you think.

We didn't include them in the initial iteration of the PR label section since those ones are automatically managed by the bots, but I don't think there would be an issue with adding them. That should probably be done in a separate PR after this one is finished though.

PR #532, opened 2019-09-09
Touch up Getting Started

brettcannon:
1. Mention that `make -j` works w/o specifying the number of cores
2. Add gdbm as a package to install from Homebrew

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -17,7 +17,6 @@ might want to use `our official images
 contain the latest releases of several Python versions, along with git head,
 and are provided for development and testing purposes only.
 
-
 .. seealso::
 
    The :ref:`quick-reference` gives brief summary of the process from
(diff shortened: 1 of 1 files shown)


setup.rst
```diff
@@ -166,7 +166,9 @@ Once ``configure`` is done, you can then compile CPython with:
 This will build CPython with only warnings and errors being printed to
 stderr and utilize up to 2 CPU cores. If you are using a multi-core machine
 with more than 2 cores (or a single-core machine), you can adjust the number
-passed into the ``-j`` flag to match the number of cores you have.
+passed into the ``-j`` flag to match the number of cores you have (or if your
+version of Make supports it, you can leave the number off and Make will use
+as many cores as your machine has).
```
ned-deily: > and Make will use as many cores as your machine has

I don't think that's quite right.  For example, on macOS 10.14, the current `man make` (which is for `GNU Make 3.81`) states:

> If the -j option is given without an argument, make will not limit the  number of jobs that can run simultaneously.

and, looking at the output of `make -d` leads me to believe that the number is indeed not related to the number of cores but can be much more.

setup.rst
```diff
@@ -378,7 +380,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz
+    $ brew install openssl xz gdbm
```
ned-deily: You should make a similar change for MacPorts below adding `gdbm`.

ned-deily:
@brettcannon How's that look?

PR #533, opened 2019-09-12
Update the developer log based on the official historical record

### LINKED ISSUE #390: Simplify developer log

brettcannon:
Closes #390

### DIFF
--- a/developers.rst
+++ b/developers.rst
@@ -3,7 +3,7 @@
 Developer Log
 =============
 
-This page list the historical members of the Python development team. (The
+This page lists the historical members of the Python development team. (The
 master list is kept in a private repository due to containing sensitive contact
 information.)
 


aeros167:
@brettcannon Could we also add a column to this list that provides a GitHub username for each core developer?

The more long term solution would be to update the experts list as described in https://github.com/python/devguide/issues/507. But, there has not been any progress from when I opened a PR to create the JSON to store the expert data (GH-507) approx two months ago. In the meantime, it might be a decent option to add the GitHub usernames for all of the core developers that have one to the developer log.

If the information is not available in the private committers repository, I would not mind opening a new PR to complete the process manually. For my own purposes during PR review, it can be quite tedious to find the GitHub username of core developers on the experts index (since that only contains the bpo usernames). We could really use at least one comprehensive list.

brettcannon:
@aeros167 Technically it could be private who is what GitHub username (we have no requirement that you can't have a pseudo-anonymous GitHub account which doesn't expose your real name). Basically someone has to ask the core devs if anyone objects.

aeros167:
@brettcannon:
> Basically someone has to ask the core devs if anyone objects.

I'll open up a new thread in python-dev in the near future to gather some feedback on the subject. My idea for this would be to make it an opt-out system, and to initially only include those that have their real name mentioned on their GitHub user page or included as part of their username (such as yours). If anyone doesn't want to be included, we could simply leave the GitHub column for them as null or an empty string (depending on the implementation details of the `csv-table` directive). 

The main useful part would be for the purpose of easily finding the GitHub usernames for the experts, but it seems like that's going to be more of a long term project. We definitely could benefit from having a single location where there's a public listing of GitHub usernames for each of the core developers that isn't tied to bpo, since we're working on moving away from there.

aeros167:
If it is accepted, would you prefer for the usernames to be manually added to the .csv so that any developers who don't have their GitHub username associated with their real name, or those who wish to opt-out? From my understanding, this would be the only change needed for the .rst:
```
.. csv-table::
   :header: "Name", "Joined", "GitHub"
   :file: developers.csv
   :encoding: "utf-8"
```
It might be a bit tedious, but I think it will ultimately be worthwhile. It'll give me a good chance to familiarize myself with all of the core dev GitHub usernames. I'm familiar with most of the currently active ones at this point, but I'm sure there's several sporadically active ones that I haven't encountered.

brettcannon:
@aeros167 I really don't want to make this an opt-out system. Either no one will care and we will do it or some people will and so we won't do it since you could infer who is who on GitHub through process of elimination.

And I will handle updating everything if this happens so don't worry about the changes to make (there are other things that may have to change).

aeros167:
@brettcannon:
> I really don't want to make this an opt-out system.

I'm good with it either way. I just figured that might simplify things a bit. If it is rejected, the experts index should at least contain GitHub names as well once we get further along on that project. Those are the primary useful GitHub usernames for triagers and PR authors to be aware of, and should definitely be posted somewhere public (outside of bpo). 

> And I will handle updating everything if this happens so don't worry about the changes to make (there are other things that may have to change).

Okay, sounds good. I figured you already had a decent amount on your plate and could use the assistance, but I have no issues with that. Let me know if you need any help with any part of it though, this particular project is important to me. (:

Also, I'm eager to help with the any part of the general GitHub migration process (if possible). Although bpo seems to have served the Python development community well overall, I very much appreciate a lot of the features in GitHub. Feel free to mention or directly assign me to any related issues that I would be able to help with.

PR #535, opened 2019-09-17
Sort developers list

matrixise:
@brettcannon I have sorted the list of the developers by date. 

1. commit with dos2unix
2. commit where I sort the developers by date.


brettcannon:
Please don't update here without updating the source data in python/voters as this file is auto-generated. It's best to simply fix the TOML file to be sorted by date again and update `coredev validate` to check that the order is maintained.

PR #540, opened 2019-10-06
Activate the environment created by make venv

bsolomon1124:
The Doc/Makefile venv
(https://github.com/python/cpython/blob/24ddd9c2d6ab61cbce7e68d6de36d4df9bd2c3fb/Doc/Makefile#L143)
does not actually activate the virtual environment.

Subsequently, the build command (via make html) would fail
because none of the programs just installed are actually
available, hence user would always get

    Missing the required blurb or sphinx-build tools

if they did not activate the virtualenv.

This also voids the usability of `make -C Doc/ venv html`.

### DIFF
--- a/documenting.rst
+++ b/documenting.rst
@@ -1542,9 +1542,8 @@ Using make / make.bat
 
    cd Doc
    make venv
+   source ./venv/bin/activate
    make html
-
-or alternatively ``make -C Doc/ venv html``.
 
 You can also use ``make help`` to see a list of targets supported by
 :command:`make`.  Note that ``make check`` is automatically run when


encukou:
A venv does not need to be activated; it is enough to run its Python, i.e. `$(VENVDIR)/bin/python3` or `$(VENVDIR)/bin/blurb`. The Makefile rules should either run this directly, or add ``$(VENVDIR)/bin` to `$PATH`.

You shouldn't be getting `Missing the required blurb or sphinx-build tools`, but the error is somewhere else than unactivated venv.

Do you still get the error if you run these sequentially – `make env`; `make html`?

bsolomon1124:
@encukou you're right, I missed the `BLURB = PATH=$(VENVDIR)/bin:$$PATH blurb` part.

I think this was an error on my part.  I had (a while back) made my own venv and installed from `Doc/requirements.txt`, which has `blurb` but not `sphinx-build`.

Anyways, `rf -rf venv && make venv && make html` works fine now.  Thanks for looking over this PR.

PR #546, opened 2019-10-15
Python docs translations: add Indonesian

JulienPalard:


oonid:
dear @JulienPalard 
would you please do a small revision?
Indonesian (country and) language code: **id** 
and here's my account: https://bugs.python.org/user32660
thank you.

JulienPalard:
@oonid Fixed in https://github.com/python/devguide/pull/547 can you proofread? Sorry for the language code. I double checked on Wikipedia and still wrote it wrong, not proud :)

oonid:
@JulienPalard I have checked it (#547), looks good!
it's ok Julien, just sometimes it happened.

PR #550, opened 2019-11-04
change wording re. who defines what is "Pythonic"

taleinat:
Guido is no longer the BDFL :(

This is now, in a general sense, defined by the Steering Council.


willingc (approved):

willingc:
Thanks Tal.

Issue #555, opened 2019-11-23
Add a note about running a make clean before or after re-running configure


DahlitzFlorian:
As mentioned in [bpo-38877](https://bugs.python.org/issue38877) it might be necessary to run a `make clean` before of after running re-running `configure` in a particular build directory. Therefore, a note should be displayed in the setup section of the devguide.

ned-deily:
Thanks for the PR (and sorry for the delay in reviewing it).

PR #556, opened 2019-11-23
Add a note about running make clean before or after re-running configure

DahlitzFlorian:
See issue: #555


DahlitzFlorian:
@ned-deily Hi Ned, we talked about this note in [bpo-38877](https://bugs.python.org/issue38877). Could you please review this PR if you have the time? If I missed something, let me know!

PR #561, opened 2020-01-14
Fix a typo in the Quick Links section

ngie-eign:
The phrase "Get Setup" should be spelled like "Git Setup". The
former doesn't make any semantic sense, whereas the latter does.

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>


ncoghlan:
Thanks!

PR #566, opened 2020-01-28
Purge removed modules from expert's list

isidentical:
Some modules that are removed but not included in this patch (like `_dummy_thread`) kept for bugs in the back versions (like 3.8).


ncoghlan:
Thanks!

PR #568, opened 2020-01-29
Added a security notice to buildworker.rst

opensource-assist:


buildworker.rst
```diff
@@ -348,6 +348,8 @@ python-dev about buildbot security that includes examples of the tests for
 which privilege matters.  There was no final consensus, but the information is
 useful as a point of reference.
 
+Third-party buildbots are only used for testing fixes. None of those buildbots nor the system that they run on are utilised to produce any CPython release materials.
```
pablogsal:
```suggestion
Third-party buildbots are only used for testing purposes. None of the buildbots nor the system that they run on are utilised to produce any release materials.
```

zware:
This seems out of place and unnecessary to me.

opensource-assist:
@zware
> This seems out of place and unnecessary to me.
What do you mean by 'out of place?
This is indeed necessary, because I was concerned with the security of using third-party buildbots and if they can tamper Python releases. This seemed important to me.

zware:
This page is geared towards buildbot worker operators, not so much general information about the buildbots (there's the `buildbots.rst` page that's a bit closer to that goal).

As far as "unnecessary", it's fairly clear by looking at the build configurations or logs (which would be a lot more trustworthy than a random note in documentation that frequently drifts out of date, for anyone who is truly concerned about security) that we don't even currently build installers on the buildbots, much less produce release materials.

opensource-assist:
@zware 
> This page is geared towards buildbot worker operators, not so much general information about the buildbots (there's the `buildbots.rst` page that's a bit closer to that goal).

I was thinking about that too; but as I see there was a section on buildworker.rst I just wanted to put it in there.
I also wanted to notify any evil-minded entities about this security precaution when they browse the buildworker.rst page. 

> As far as "unnecessary", it's fairly clear by looking at the build configurations or logs (which would be a lot more trustworthy than a random note in documentation that frequently drifts out of date, for anyone who is truly concerned about security) that we don't even currently build installers on the buildbots, much less produce release materials.

That's the whole point of documenting something, you wanna make it easier for other people to know about the inner workings of the buildbots. It gives a peace of mind to anybody reading it, if they were concerned.

willingc:
Hi @opensource-assist. Thanks for the documentation suggestion. Looking at the suggested text and others' comments I am closing this PR as out of scope for the devguide. Thanks! 🌻

PR #575, opened 2020-03-05
Add Python Language Reference docs to Grammar checklist.

brandtbucher:
It looks like these docs (specifically the grammar reference for [expressions](https://docs.python.org/3/reference/expressions.html)) got stale when the walrus was added. It's making documenting PEP 614 very tricky.

Also, this checklist is an amazing resource. I was halfway through debugging segfaults in `importlib` before I found this!


brettcannon:
Thanks!

Issue #336, opened 2018-03-07
Missing alternate Windows step for installing coverage


abrezovsky:
In /coverage/ section 5.2.1.1
When installing coverage on Windows, the second command step can not be used:

`source ../cpython-venv/bin/activate`

Instead, there should probably be some mention of the proper step:

`..\cpython-venv\Scripts\activate.bat`

Which allows you to enter the virtual environment and continue with the instructions.

abrezovsky:
As a side-note, it might be worth mentioning that activate.ps1 needs to be used if trying to go through PowerShell.

brettcannon:
That command also doesn't necessary work on UNIX either depending on your shell. A comment saying to change that line accordingly based on your shell should be sufficient rather than listing every possible invocation.

Mariatta:
In that same section there is a link to venv's documentation, where it says: `For more info on venv see Virtual Envrionment documentation.`
The venv documentation has more details for Windows, `csh` or `fish` shell users.

willingc:
Next action: Revise https://github.com/python/devguide/blob/master/coverage.rst#install-coverage to add these suggestions:
- [ ] More info on venv as suggested by @Mariatta 
- [ ] Comment on command may need slight modifications based on shell and os per @brettcannon - [ ] Add Windows command for illustration of a modification for os per @abrezovsky

PR #584, opened 2020-05-16
Improve steps to install coverage in Virtualenv on Windows (#336)

scrool:
Also on Mac OS X systems.

Instead of code section followed by alternative solutions introduce multiple code sections by system name. This way one can easily skip to section one is intersted in. This way tutorial on virtual env is written as well.


willingc (approved):
Thanks @scrool for the clarifications to the coverage installation for Mac and Window. Happy to merge.

Issue #577, opened 2020-03-09
Use imperative for commit title; fix example


terryjreedy:
(See GvR's messages 363691 and 363740 on 2020 March 9 for https://bugs.python.org/issue39852.)
https://devguide.python.org/pullrequest/#making-good-commits has this descriptive example:
`bpo-42: the spam module is now more spammy (GH-NNNN)`
Guido strongly prefers the prevailing imperative style and would have that read
 `bpo-42: make the spam module more spammy (GH-NNNN)`

Possibly split the following paragraph into two and after "... what the purpose of the commit is." add something like
"The imperative form is strongly preferred to a descriptive form such as 'the spam module is now more spammy'.  Use `git log --oneline` to see existing title lines."

gvanrossum:
While we're at it, let's also codify that the commit "title" shouldn't end in a period.

Also, let's more directly reference this guide to writing commit messages: https://chris.beams.io/posts/git-commit/

DahlitzFlorian:
I would like to work on this if nobody is currently doing it.

PR #587, opened 2020-05-21
Fix good commit example to use imperative title

DahlitzFlorian:
Closes: #577 

- Fix good commit example to use imperative title
- Explicitly mention the preference for imperative over descriptive titles
- Link to the article from Chris Beams

@gvanrossum I added the link to the article as suggested by you. The [Git Bootcamp](https://devguide.python.org/gitbootcamp/#accepting-and-merging-a-pr) is mentioning the article, too, but I agree with you that it is better to explicitly mention it at this point, too.


pullrequest.rst
```diff
@@ -258,21 +258,30 @@ and for each pull request there may be several commits.  In particular:
...
 
    The spam module sporadically came up short on spam. This change
    raises the amount of spam in the module by making it more spammy.
 
 The first line or sentence is meant to be a dense, to-the-point explanation
-of what the purpose of the commit is.  If this is not enough detail for a
-commit, a new paragraph(s) can be added to explain in proper depth what has
-happened (detail should be good enough that a core developer reading the
-commit message understands the justification for the change).
+of what the purpose of the commit is. The imperative form is strongly
```
aeros, line 267: Since we have many developers with varying levels of English proficiency, some of them may not be overly familiar with the difference between imperative and descriptive form. As a result, I think it's worth having `(used in the example above)` after "imperative form" to make it a bit more explicit as to what is being referred to.

E.g. "The imperative form (used in the example above) is strongly preferred..."

DahlitzFlorian:
Thanks for the fast feedback @aeros! I applied your suggestion.

pullrequest.rst
```diff
@@ -258,21 +258,30 @@ and for each pull request there may be several commits.  In particular:
 
 Commit messages should follow the following structure::
 
-   bpo-42: the spam module is now more spammy (GH-NNNN)
+   bpo-42: make the spam module more spammy (GH-NNNN)
```
terryjreedy, line 261: This comment is based on my preference and looking at two pages of current  `git log --oneline` output.
```suggestion
   bpo-42: Make the spam module more spammy (GH-NNNN)
```
An initial capital is nearly always used if there is no bpo number.  The exception was of the form 's/type/fix/'.  When there is a bpo number, capitals are more common than not.
@gvanrossum Do you disagree with the change?

DahlitzFlorian: @terryjreedy I agree with you. I saw the capital letter on various commit messages, too, and I'm using it as well. Furthermore, the devguide's [quick reference](https://devguide.python.org/#quick-reference) uses a capital letter in point seven, too.
If @gvanrossum agrees on this, I'm happy to apply the suggested change.

aeros: I'm personally neutral about the change because I don't think it has a significant effect on the readability of the commit messages for the purposes of detailing the changes (unlike imperative vs descriptive, IMO). So while the prevailing style is sentence case capitalization after the `bpo-NNN:`, I don't know that it's a real concern.

That being said, if we just modified the example to use it (as proposed above) rather than writing about in the guidelines, I don't have an issue with it.

terryjreedy:
LGTM.  Kyle, leaving this for you if you get to it.

pullrequest.rst
```diff
@@ -258,21 +258,30 @@ and for each pull request there may be several commits.  In particular:
...
+that a core developer reading the commit message understands the
+justification for the change).
 
 Check :ref:`the git bootcamp <accepting-and-merging-a-pr>` for further
 instructions on how the commit message should look like when merging a pull
 request.
 
+.. note::
+   `How to Write a Git Commit Message <https://chris.beams.io/posts/git-commit/>`_
+   is a nice article describing how to write a good commit message.
```
aeros, line 283:
```suggestion
   is a nice article that describes how to write a good commit message.
```

Issue #593, opened 2020-06-16
Test coverage doc fix


lrjball:
Currently in the test coverage section, specifically 5.2.1.4, it mentions building the C extension for coverage.py. For a while now coverage.py has come with the C extension by default, so this step is almost never required (see here: https://coverage.readthedocs.io/en/coverage-5.1/install.html#c-extension). 
I'm thinking it would be worth mentioning in the guide that installing the C extension is only required if 'coverage --version' says 'Without C extension'.

PR #594, opened 2020-06-16
Updated docs for coverage c extension

lrjball:
Related issue is #593

The documentation around building the c extension for coverage.py is usually not required, as the c extension is usually installed by default. I've added some clarification of this in the docs.


willingc (approved):
Thanks @lrjball for the clarification on coverage and the c extension. 👍

PR #598, opened 2020-07-13
Clarify branch/release terminology in Development Cycle page.

ned-deily:
Try to be more consistent about use of major.minor.micro version
numbers and of feature, bugfix, and security releases.  Reflect
current policy and practice for branch transitions as of 3.9.
Other minor cleanups.


willingc (approved):
Thanks @ned-deily. Sorry for the delayed review.

PR #606, opened 2020-08-13
pl translation migrated to python org.

JulienPalard:

PR #607, opened 2020-08-29
Add links to GitHub repository and builds on docs.python.org for Polish docs translations

m-aciek:
Plus remove e-mail link.

For about two weeks translation repository is within Python organisation on GitHub and builds are being published on docs.python.org.


m-aciek:
This is almost duplicate PR of #606, but it adds link to initial doc builds.

willingc (approved):
Thanks for the change @m-aciek.

PR #614, opened 2020-09-30
developers.csv: Escape dot to prevent <ol>

encukou:
Re-apply commit 0427168c1ae4f69aff299fa215dd2ae7aaa0ff15.
This time the issue was fixed in the script that generates the data,
so it shouldn't reappear.

Co-Authored-By: Hugo van Kemenade <hugovk@users.noreply.github.com>


hugovk (approved):
Thanks!

PR #621, opened 2020-10-14
Remove italics: Python 3.9.0 was released on schedule

hugovk:
https://www.python.org/downloads/release/python-390/


willingc (approved):
Thanks @hugovk

Issue #625, opened 2020-10-19
Change Zulip badge in README to Discourse


willingc:
Discourse is found at discuss.python.org. Remove Zulip badge.

aliya-rahmani:
I've made PR #628 for this issue.

PR #630, opened 2020-10-20
Updated zulip to discourse #625

aliya-rahmani:


hugovk:
I guess the first two commits from Mariatta aren't needed here?

willingc:
Hi @aliya-rahmani, 

I've updated your PR to only include the desired commits. The devguide has some helpful git tips for changing pull requests and keeping your fork in sync with this repo. Here's the page: https://devguide.python.org/gitbootcamp/

Thanks for the PR too :sunny:

PR #633, opened 2020-10-21
CONTRIBUTING: Link to bugs.python.org and python-dev

hugovk:


willingc (approved):
Good catches. Thank you :sunny:

PR #639, opened 2020-10-27
Add an entry for Doc/library/ast.rst

isidentical:
Since 3.9, we started to document each each individual node in the [AST](https://docs.python.org/library/3/ast.html) docs. It might be a common catch to forget when a new node is introduced or existing ones are updated / removed, so added a checklist entry.

### DIFF
--- a/grammar.rst
+++ b/grammar.rst
@@ -49,6 +49,9 @@ Note: sometimes things mysteriously don't work.  Before giving up, try ``make cl
 * ``_Unparser`` in the :file:`Lib/ast.py` file may need changes to accommodate
   any modifications in the AST nodes.
 
+* If any AST node introduced or updated, changes needs to be reflacted to
+  :file:`Doc/library/ast.rst`. 
+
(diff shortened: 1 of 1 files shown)


grammar.rst
```diff
@@ -49,6 +49,9 @@ Note: sometimes things mysteriously don't work.  Before giving up, try ``make cl
 * ``_Unparser`` in the :file:`Lib/ast.py` file may need changes to accommodate
   any modifications in the AST nodes.
 
+* If any AST node introduced or updated, changes needs to be reflacted to
+  :file:`Doc/library/ast.rst`.
```
encukou, line 53: Most of the checkpoints start with the name of the file, which makes them easier to scan. May I suggest:

```suggestion
* :file:`Doc/library/ast.rst` may need to be updated to reflect changes to AST nodes.
```

isidentical: Makes sense, thanks!

pablogsal (approved):

PR #641, opened 2020-11-09
Changing Arabic coordinator

Abdur-rahmaanJ:
Changing the arabic coordinator. Don't merge before getting a written confirmation here from Mr Ibrahim.

### DIFF
--- a/documenting.rst
+++ b/documenting.rst
@@ -1595,7 +1595,7 @@ in production, other are work in progress:
 +-----------------+-------------------------------+----------------------------+
 | Language        | Contact                       | Links                      |
 +=================+===============================+============================+
-| Arabic (ar)     | Abdur-Rahmaan Janhangeer      | `GitHub <github_ar_>`_     |
+| Arabic (ar)     | Ibrahim Elbouhissi            | `GitHub <github_ar_>`_     |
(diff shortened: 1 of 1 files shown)


JulienPalard:
In case it helps someone in the future disambuguate a situation: this is the right way to do it.

I mean, there's no need for me or anyone in particular to validate how a language choose its translator as long as everyone agrees.

(This mean, in case a translation can't agree on a translation coordinator, they can either don't have one, or organize a vote, or anything else, autonomously).

IbrahimElbouhissi:
Hi everyone, 

Nice to be with you in this Glorious mission. I am happy to accept this task and be the Arabic Translation Coordinator of Python docs. 

To be clear and honest, it is important to tell you that this is my first try to work with a team in Github. This makes me work more hard, and do my best. Any help will be appreciated.

@Abdur-rahmaanJ it is the time to next step. 

BR

JulienPalard:
Welcome on board! Don't hesitate to ask if something is not straightforward in the translation process. There's the [translation@](https://mail.python.org/mailman3/lists/translation.python.org/) list, or even in private, I'm mdk on freenode in #python-docs, or julien at python dot org),

PR #643, opened 2020-11-13
Advice user to use `make html` from within the `Doc` directory

jugmac00:


Mariatta:
This particular section is for building the devguide documentation, and there is no Doc directory.
I think you're thinking about building the CPython documentation, https://devguide.python.org/documenting/#building-the-documentation, and in there, we already include the `cd Doc` command line before `make html`

jugmac00:
Oops, thanks for getting back to me and sorry for the noise.

PR #666, opened 2021-03-21
Fix link to custom builders

pitrou:
The name of the tags classifying the various builders has changed at some point, but the devguide hasn't been updated, so the current link points to a list of 0 builders.


pitrou:
I don't think this should be controversial, so I'm gonna merge this PR without waiting for a review.

zware (commented):
LGTM after the fact :)

PR #673, opened 2021-04-08
Add advice on building all modules on Debian-like

Maroloccio:
Current advice regarding installing dependencies is insufficient to build all optional modules:

```
The necessary bits to build these optional modules were not found:                                                            
_bz2                  _curses               _curses_panel
_dbm                  _gdbm                 _hashlib
_lzma                 _sqlite3              _ssl
_tkinter              _uuid                 readline
zlib                                                   
To find the necessary bits, look in setup.py in detect_modules() for the module's name. 
```

Some users may like the convenience of getting a head start on choosing OS packages that get them closer to a complete build.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -325,10 +325,14 @@ Then you should update the packages index::
 
 Now you can install the build dependencies via ``apt``::
 
-   $ sudo apt-get build-dep python3.6
+   $ sudo apt-get build-dep python3
 
-If that package is not available for your system, try reducing the minor
-version until you find a package that is available.
+If you want to build all optional modules, install the following packages and
+their dependencies::
+
+   $ sudo apt-get install build-essential gdb lcov libbz2-dev libffi-dev \
+         libgdbm-dev liblzma-dev libncurses5-dev libreadline6-dev \
+         libsqlite3-dev libssl-dev lzma lzma-dev tk-dev uuid-dev zlib1g-dev
 
 
 .. _MacOS:


the-knights-who-say-ni:

I'm a bot set up to make sure that the project can legally accept this contribution by verifying everyone involved has signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).

## Recognized GitHub username

We couldn't find a [bugs.python.org](https://bugs.python.org/) (b.p.o) account corresponding to the following GitHub usernames:

@Maroloccio

This might be simply due to a missing "GitHub Name" entry in one's b.p.o account settings. This is necessary for legal reasons before we can look at this contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.




You can [check yourself](https://check-python-cla.herokuapp.com/) to see if the CLA has been received.

Thanks again for the contribution, we look forward to reviewing it!

Maroloccio:
@erlend-aasland you :+1: this and the PR has been waiting for a merge for the last couple of months; what can I do to make the last step happen, please?

terryjreedy:
@Mariatta Do you know linux well enough to review this addition to the linux setup part of the devguide?  Or do you know who else commits devguide changes who might? (I am neither.) The idea seem straightforward, and has gotten 4 likes, but from non-committers.

erlend-aasland (approved):
LGTM, but I'm not sure about the _"versions you prefer"_ bit; CPython has version requirements for some (all?) dependencies. For example SQLite >= 3.7.15.

setup.rst
```diff
@@ -341,6 +341,14 @@ Now you can install the build dependencies via ``apt``::
 If that package is not available for your system, try reducing the minor
 version until you find a package that is available.
 
+If you want to build all modules, install these packages in the versions you
+prefer, changing the version name as above::
```
zware, line 345: I think this can be combined with the above point, as I'm pretty sure the explicit package names cover everything that would be installed by `build-dep`.  The above point should also be updated to a more recent version (or the minor version removed entirely, and the suggestion amended to suggest adding a minor version if `python3` doesn't cut it).

zware, line 345: I'm also not a fan of the "versions you prefer..." bit, I would suggest to trim it back to just 
```suggestion
If you want to build all optional modules, install the following packages and
their dependencies::
```

setup.rst
```diff
@@ -341,6 +341,14 @@ Now you can install the build dependencies via ``apt``::
...
+
+   $ sudo apt-get install build-essential gdb lcov libbz2-dev libffi-dev \
+         libgdbm-dev liblzma-dev libncurses5-dev libreadline6-dev \
+         libsqlite3-dev libssl-dev lzma lzma-dev tk-dev uuid-dev xvfb \
```
zware, line 349: `xvfb` is not actually necessary, it would only allow you to run Tk tests on a headless machine.

zware (commented):
Looks like a good addition, but I have some quibbles :)

setup.rst
erlend-aasland: +1

Maroloccio: I have removed the minor version from the Python package for which the command installs the build dependencies.

Maroloccio: I have accepted this suggestion entirely, merging it with the deletion of the previous paragraph.

setup.rst
Maroloccio: I have removed `xvfb` from the list.

Maroloccio:
Thank you @zware for your suggestions, please either Resolve the contextual exchanges or recommend further work needed on the PR.

erlend-aasland (approved):
LGTM with changes applied!

zware (approved):

PR #676, opened 2021-04-20
devcycle: Clarify doc & test changes in Beta & RC

encukou:
This is how I imagine things work in the beta & RC, but I'm likely wrong:

- beta: docs & test changes are welcome
- RC: no trivial changes

@pablogsal, as the RM, could you please review and make any changes necessary?

### DIFF
--- a/devcycle.rst
+++ b/devcycle.rst
@@ -185,9 +185,10 @@ Beta
 ----
 
 After a first beta release is published, no new features are accepted.  Only
-bug fixes can now be committed.  This is when core developers should concentrate
-on the task of fixing regressions and other new issues filed by users who have
-downloaded the alpha and beta releases.
+bug fixes and improvements to documentation and tests can now be committed.
+This is when core developers should concentrate on the task of fixing
+regressions and other new issues filed by users who have downloaded the alpha
+and beta releases.
 
 Being in beta can be viewed much like being in RC_ but without the extra
 overhead of needing commit reviews.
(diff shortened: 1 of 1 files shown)


pablogsal (approved):

pablogsal:
> * beta: docs & test changes are welcome
> * RC: no trivial changes
> 
> @pablogsal, as the RM, could you please review and make any changes necessary?

I think this is correct, at least it matches my current understanding.

pablogsal:
Chiming other RMs to double check:

CC: @ambv @ned-deily

ambv:
I like the clarification that beta still accepts doc and test improvements. I don't like the implication that release candidates don't allow those. Since we squished the beta phase a bit to hit RC1 just in time for inclusion in Fall editions of Linux distributions, we now spend two months of every year in RC phase. Saying that we shouldn't allow for documentation and tests to shape up then is unrealistic.

I understand that we don't want major refactors of tests and docs in RCs but for new functionality it's not uncommon for documentation to only be written when we pester for it while working on the What's New document before the final release -_-

ned-deily:
I agree with @ambv.  Perhaps change the RC wording to something like: "While the goal is to have no code changes between a RC and a final release, there may be a need for final documentation or test fixes. Any such proposed changes should be discussed first with the release manager."

devcycle.rst
```diff
@@ -204,8 +205,9 @@ Release Candidate (RC)
 A branch preparing for an RC release can only have bugfixes applied that have
 been reviewed by other core developers.  Generally, these issues must be
 severe enough (e.g. crashes) that they deserve fixing before the final release.
-All other issues should be deferred to the next development cycle, since
-stability is the strongest concern at this point.
+All other issues (including documentation and test changes) should be deferred
+to the next development cycle, since stability is the strongest concern at this
+point.
```
encukou, line 210:
```suggestion
All other issues should be deferred to the next development cycle, since
stability is the strongest concern at this point.

While the goal is to have no code changes between a RC and a final release,
there may be a need for final documentation or test fixes. Any such proposed
changes should be discussed first with the release manager.
```

encukou:
Thanks for the clarification!

pablogsal (approved):

PR #682, opened 2021-04-29
Add a page on changing the C API (PEP-652)

encukou:
This repeats points currently said in `Include/README.rst`.
I intend to replace most of that README with a link to this document.

The main change from that document are:
- The three "rings" are not defined by where things are declared.
  It would be good to have them in the right place, but it's
  not yet the case, and for technical reasons we might never reach
  the ideal. (In `Include/README.rst` it makes more sense to make
  the locations more prominent.)
- The order is switched. For changing the Limited API you generally
  need to read (and follow guidelines in) the preceding sections
  as well.

I added points from PEP-652 (Maintaining the Stable ABI).


encukou:
The page, rendered on RTD: https://cpython-devguide--682.org.readthedocs.build/c-api/

@erlend-aasland, you wrote `Include/README.rst`. Does this PR look OK to you?

Note that I also want to improve ``Doc/c-api/stable.rst`: https://github.com/python/cpython/pull/25668

c-api.rst
```diff
@@ -0,0 +1,191 @@
...
+The Limited API should be defined in ``Include/``, excluding the
+``cpython`` and ``internal`` subdirectories.
+
+Guidelines for changing the Limited API
+---------------------------------------
+
+- Guidelines for the general :ref:`public-capi` apply.
+
+- New Limited API should only be defined if ``Py_LIMITED_API`` is set
+  to a the version the API was added in or higher.
```
erlend-aasland, line 109: Looks like there's left-overs from an earlier formulation here :)

c-api.rst
```diff
@@ -0,0 +1,191 @@
...
+
+  - Macros should not expose implementation details
+  - Functions must be exported as actual functions, not (only)
+    as functions-like macros.
+  - If possible, avoid macros. This makes the Limited API more usable in
+    languages that don't use the C preprocessor.
+
+- Please start a public discussion before expanding the Limited API
+
+- The limited API and its intended use must follow standard C, not just
```
erlend-aasland, line 127:
```suggestion
- The Limited API and its intended use must follow standard C, not just
```

(Nit: I find the _"its intended use"_ part a tiny bit weird. Can it be reformulated?)

c-api.rst
```diff
@@ -0,0 +1,191 @@
...
+- Add the declaration to a header file directly under ``Include/``, into a
+  block guarded with the following:
+
+  .. code-block:: c
+
+    #if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x03yy0000
+
+  with the ``yy`` corresponding to the target CPython version, e.g.
+  ``0x030A0000`` for Python 3.10.
+- Add an entry to the end of the Stable ABI manifest, ``Misc/stable_abi.txt``.
```
erlend-aasland, line 178:
```suggestion
- Append an entry to the Stable ABI manifest, ``Misc/stable_abi.txt``.
```

c-api.rst
```diff
@@ -0,0 +1,191 @@
...
+
+.. _"Finalizing the API" in PEP 590: https://www.python.org/dev/peps/pep-0590/#finalizing-the-api
+
+
+.. _public-capi:
+
+Public C API
+============
+
+CPython's public C API if available when ``Python.h`` is included normally.
```
erlend-aasland, line 71:
```suggestion
CPython's public C API is available when ``Python.h`` is included normally.
```

c-api.rst
```diff
@@ -0,0 +1,191 @@
+=======================
+Changing Python's C API
+=======================
+
+The C API is divided into three sections:
+
+1. The internal, private API, available with ``Py_BUILD_CORE`` defined.
+   Ideally declared in ``Include/internal/``. Any API named with a leading
+   underscore is also considered private.
+2. The public C API, available when ``Python.h`` is included normally.
```
erlend-aasland, line 10: Maybe drop _"normally"_?

c-api.rst
erlend-aasland: Also, the "follow standard C" part is covered in line 130.

erlend-aasland (commented):
LGTM, with some nitpicks :)

c-api.rst
encukou: By *normally* I mean *without defining macros to select the other variants*. I'll expand that in the section below.

encukou:
Thanks for the comments!
Do things look better now?

erlend-aasland (approved):
LGTM :)

encukou:
Thanks!

vstinner:
https://devguide.python.org/c-api/ is great, thanks @encukou!

Next step: document C API **removal** ;-)

PR #80, opened 2016-12-09
Add a checklist for committers

brettcannon:

PR #685, opened 2021-04-30
Remove mention of old coverage comment from checklist

ammaraskar:
This comment seems to be an old revised mention about codecov's coverage comments from here: https://github.com/python/devguide/pull/80/files#diff-eb1b0e0746ead0f2c05fe91ab3cf886a06b30c20459147869209616b5d2279cbR20-R21

We've long disabled codecov posting comments on pull requests about coverage, and with https://github.com/python/cpython/pull/25679 working to remove the coverage build, we can get rid of this mention.


Mariatta (approved):

Mariatta:
Thanks!

PR #688, opened 2021-05-05
Pip updates during the installation of the virtual env

matrixise:


JulienPalard:
I don't think it make sense to update pip while installing requirements. It would make sense to upgrade pip first, then use the new pip to install requirements.

Do you hit an issue with a venv created with a too-old pip?

matrixise:
@JulienPalard 

Do you prefer this commit? https://github.com/python/devguide/pull/688/commits/f38e5a546adffd728b97b7c6c5aa5e13d2f09e1d

About the issue, just the warning (outdated version of pip) that pollutes my terminal

JulienPalard:
> Do you prefer this commit? [f38e5a5](https://github.com/python/devguide/commit/f38e5a546adffd728b97b7c6c5aa5e13d2f09e1d)

Yes I prefer it this way, so the pip used to install requirement is the new one, not the old one.

matrixise:
@JulienPalard updated.

JulienPalard:
Thanks @matrixise ☺

PR #690, opened 2021-05-09
Remove the venv dir and introduce VENVDIR

matrixise:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,9 @@
 
 # You can set these variables from the command line.
 PYTHON        = python3
+VENVDIR       = ./venv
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Looks good, it allows people to set their own venv dir, similar to the CPython docs `Makefile`:

https://github.com/python/cpython/blob/e8e737bcf6d22927caebc30c5d57ac4634063219/Doc/Makefile#L8-L11

ezio-melotti:
I already fixed this as part of #856 and just saw this PR now.
Since it's now superseded, I'm going to close it.
Thanks anyway for the PR!

PR #691, opened 2021-05-10
Remove replace description for py27 in setup step

00Kai0:
Remove replace description for python2. (It's no longer needed after EOL)


00Kai0:
@methane , committed. Now just remove description for py27 :)

PR #696, opened 2021-05-15
minor grammar improvement

Bonifacio2:
The former sentence used to say "Changes (...) follows (...) and is (...)", but since the sentence subject is a plural then the verbs should be replaced with (they) "follow" and "are". Does that make sense?


encukou:
Looks good; thank you!

PR #701, opened 2021-05-18
fixes #502: include infromation about auto-closure of bpo isssues

### LINKED ISSUE #502: Document that bpo issues can be automatically closed using commit message
According to PEP 595, bpo now has a feature where issues can be automatically closed, if the PR commit message looks like `closes bpo-NNNN: ...`

This should be documented in devguide and announced to core devs/core-workflow mailing list/discourse.

Some questions: what magic keywords accepted? "closes" "fixes" "close" "fix"?
Is it case sensitive? (I assume not)

Does it also update the "resolution" / "stage"?

rahul-kumi:
Fixes #502

Feature that was included in PEP-595 to automatically close issue
in the bug tracker when a PR with commit message including words like-
close, closes, closed, closing, fix, fixes or fixed being used with
issue number is merged is documented in the devguide.

### DIFF
--- a/pullrequest.rst
+++ b/pullrequest.rst
@@ -286,6 +286,11 @@ to explain in proper depth what has happened (detail should be good enough
 that a core developer reading the commit message understands the
 justification for the change).
 
+Additionally, commit message when started with these words-
+**close**, **closes**, **closed**, **closing**, **fix**, **fixes** or **fixed**
+along with the issue number(bpo-NNN) like ``closes bpo-NNN: ....`` ensure
+auto-closure of the issue in bug tracker as the PR containing the commit is merged.
+
 Check :ref:`the git bootcamp <accepting-and-merging-a-pr>` for further
 instructions on how the commit message should look like when merging a pull
 request.


rahul-kumi:
@hugovk, the CLA is signed...still the CLA signing status is stated as expected.
Can you please help me with this.

JelleZijlstra (commented):
This still talks about bpo, which is no longer relevant.

Additionally, it would fit better at https://devguide.python.org/gitbootcamp/#accepting-and-merging-a-pr where we already talk about commit messages.

ezio-melotti:
Closing since we now migrated to GitHub.

gvanrossum:
But GitHub has a similar feature. Is *that* documented?

ezio-melotti:
> Is that documented?

Apparently not. 

Recently there was a discussion about the fact that using the auto-closing keywords is not always desirable, e.g.:
* when the PR needs to be backported 
* when multiple PRs address the same issue

In these cases it might be better to avoid keywords and just close the issue manually once all the relevant PRs are merged.  Both the keywords and recommendations on where they should be used/avoid can be documented.

I would really like to see sections with basic instructions/tips/reference for all the tools we use, but I'm still not sure what would be the best way to organize this.

PR #706, opened 2021-05-23
Bring compiler docs up to speed with Python 3.10

### LINKED ISSUE #678: Update compiler docs with the new parser changes
The documentation regarding the design of CPython's compiler should be updated to include the new changes in the parser:

* AST nodes are not generated anymore by hand (`ast.c` is almost gone) but generated directly by the parser.
* There is a new API for working with `asdl_seq` nodes.

Fidget-Spinner:
Fixes #678.

* Updated renamed functions and files in 3.10.
* Small note for Windows ``build.bat``.
* Updated compiler docs with new PEG parser and typed ASDL sequences.


Fidget-Spinner:
Reviews are welcome and very appreciated -- I'm no expert in the compiler.

compiler.rst
```diff
@@ -159,49 +159,81 @@ are very rare.  However, if you've allocated a PyObject, you must tell
 the arena about it by calling ``PyArena_AddPyObject()``.
 
 
-Parse Tree to AST
------------------
-
-The AST is generated from the parse tree (see :file:`Python/ast.c`) using the
-function ``PyAST_FromNode()``.
+Tokens to AST
```
pablogsal, line 162: Using tokens here is a bit confusing because the source of all the "public" APIs used by interpreter takes the source code as a string buffer, and in the process of parsing, it transforms it to tokens. I suggest to change these occurrences to "source code to AST".

compiler.rst
```diff
@@ -159,49 +159,81 @@ are very rare.  However, if you've allocated a PyObject, you must tell
...
+-------------
 
-The function begins a tree walk of the parse tree, creating various AST
-nodes as it goes along.  It does this by allocating all new nodes it
-needs, calling the proper AST node creation functions for any required
-supporting functions, and connecting them as needed.
+The AST is generated from tokens using the function
+``_PyPegen_run_parser()`` (:file:`Parser/pegen.c`), which is a thin wrapper over
+the function ``_PyPegen_parse()`` (:file:`Parser/parser.c`).
```
pablogsal, line 167: `_PyPegen_run_parser` and `_PyPegen_parse` are technically implementation details. I prefer to reference the actual APis used by the interepter: `_PyParser_ASTFromFile` and `_PyParser_ASTFromString` (and other variants).

compiler.rst
```diff
@@ -159,49 +159,81 @@ are very rare.  However, if you've allocated a PyObject, you must tell
...
+:file:`Parser/asdl_c.py` from :file:`Parser/Python.asdl`).  This all leads to a
 sequence of AST nodes stored in ``asdl_seq`` structs.
 
-Function and macros for creating and using ``asdl_seq *`` types as found
-in :file:`Python/asdl.c` and :file:`Include/asdl.h` are as follows:
+To avoid bad performance during backtracking, rules (other than the top level
+ones) are memoized.  Each rule checks if a memoized version exists and returns
+that if so, else it continues in the manner stated in the previous paragraph.
+
+Macros for using ``asdl_seq *`` types as found
```
pablogsal, line 192: I would prefer to mention *first* the typed macros and then mention that for generic functions, there are untyped counterparts that should be used as a last resort when some function that acts on generic sequences is required.

compiler.rst
```diff
@@ -159,49 +159,81 @@ are very rare.  However, if you've allocated a PyObject, you must tell
...
-``asdl_seq_GET(asdl_seq *, int)``
+``asdl_seq_GET_UNTYPED(asdl_seq *, int)``
         Get item held at a specific position in an ``asdl_seq``
-``asdl_seq_SET(asdl_seq *, int, stmt_ty)``
+``asdl_seq_SET_UNTYPED(asdl_seq *, int, stmt_ty)``
         Set a specific index in an ``asdl_seq`` to the specified value
 ``asdl_seq_LEN(asdl_seq *)``
         Return the length of an ``asdl_seq``
 
+There are also typed counterparts for some of the macros above.  These
```
pablogsal, line 203: Let's mention that these sequence functions and another part of the API is automatically generated by the AST generation code, while the generic headers live in https://github.com/python/cpython/blob/main/Include/internal/pycore_asdl.h

compiler.rst
Fidget-Spinner: Thank you. I was initially confused too but I was hesitant to change it -- lots of the online diagrams I read didn't seem to actually reflect what the source code was doing. Eventually I just gave up and stepped through the source code to write these docs :P.

compiler.rst
```diff
@@ -419,7 +461,27 @@ Important Files
...
+
+    parser.c
+        The parser generated by :file:`Tools/peg_generator/pegen/c_generator.py`
+        from the grammar :file:`Grammar/python.gram`.  Creates the AST from
+        source code.  Rule functions for their corresponding production rules
+        are found here.
+
+    peg_api.c
+        Contains high-level functions to create an AST from source code which
+        are used by the interpreter.
```
lysnikolaou, line 474:
```suggestion
        Contains high-level functions, that are part of the C API, to create an AST from source code which
        are used by the interpreter.
```

compiler.rst
```diff
@@ -419,7 +461,27 @@ Important Files
...
+
+    pegen.c
+        The new PEG parser introduced in 3.9.  Contains helper functions which
```
lysnikolaou, line 477: Let's move the first sentence, `The new PEG parser introduced in 3.9.` to `parser.c`. `pegen.c` only contains helper functions.

compiler.rst
```diff
@@ -159,49 +159,88 @@ are very rare.  However, if you've allocated a PyObject, you must tell
...
 sequence of AST nodes stored in ``asdl_seq`` structs.
 
-Function and macros for creating and using ``asdl_seq *`` types as found
-in :file:`Python/asdl.c` and :file:`Include/asdl.h` are as follows:
-
-``_Py_asdl_seq_new(Py_ssize_t, PyArena *)``
-        Allocate memory for an ``asdl_seq`` for the specified length
-``asdl_seq_GET(asdl_seq *, int)``
+To avoid bad performance during backtracking, rules (other than the top level
+ones) are memoized.  Each rule function checks if a memoized version exists and
```
lysnikolaou, line 191:
```suggestion
To avoid bad performance during backtracking, some rules, chosen by applying a
flag in the grammar file, are memoized.  Each rule function checks if a memoized
version exists and
```

lysnikolaou (commented):
Thanks a lot for this @Fidget-Spinner! I left some minor comments.

compiler.rst
Fidget-Spinner: Wow thanks! I didn't know that a flag existed. Cool, I just noticed the ``(memo)`` flag in the grammar file.

compiler.rst
Fidget-Spinner: Hmm I think they were removed from the C API https://github.com/python/cpython/commit/57364ce34e0492fbc8b0a6b8c882f384bb489457.

lysnikolaou: Oh, you're right!

pablogsal (approved):

compiler.rst
```diff
@@ -419,7 +504,28 @@ Important Files
 
     asdl_c.py
         "Generate C code from an ASDL description."  Generates
-        :file:`Python/Python-ast.c` and :file:`Include/Python-ast.h`.
+        :file:`Python/Python-ast.c` and :file:`Include/Internal/pycore_ast.h`.
```
pablogsal, line 507: I like these descriptions but the header files get refactored quite a lot (specially the ast ones) so I am a bit scared that this gets out of date, but is still an improvement so I say: let's merge it.

pablogsal:
Fantastic job @Fidget-Spinner 🎉

Fidget-Spinner:
Pablo and Lysandros, thank you both for the helpful reviews! This PR has improved greatly thanks to your feedback and I've also learnt a lot.

Re:

> I like these descriptions but the header files get refactored quite a lot (specially the ast ones) so I am a bit scared that this gets out of date, but is still an improvement so I say: let's merge it.

Yup I agree.

PR #707, opened 2021-05-25
Clarify how to pass the header and library path to configure command

chrisxiao:
Using shell variable to pass header and library path to configure command does not works
so change to use environment variables

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -380,9 +380,9 @@ and ``configure`` Python versions >= 3.7::
 
 or ``configure`` Python versions < 3.7::
 
-    $ CPPFLAGS="-I$(brew --prefix openssl)/include" \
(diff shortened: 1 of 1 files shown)


pablogsal:
> Using shell variable to pass header and library path to configure command does not works
> so change to use environment variables

Why it doesn't work? This should work as this only sets the environment vars in the script execution. There must be something odd on your side

chrisxiao:
> > Using shell variable to pass header and library path to configure command does not works
> > so change to use environment variables
> 
> Why it doesn't work? This should work as this only sets the environment vars in the script execution. There must be something odd on your side

chrisxiao:
@pablogsal  Thanks for comment, I recheck and find it is my fault. the doc is right.

Issue #711, opened 2021-06-06
Error in table of contents on the gitbootcamp page


shreyanavigyan:
In the git bootcamp cheatsheet, there's an error in the table of contents. Screenshot -

![image](https://user-images.githubusercontent.com/74560907/120925230-49a88480-c6f5-11eb-8b97-39b81257e5f1.png)

akulakov:
I will make a PR to delete this list of contents since an exact copy already shows up on the right hand panel, as shown in the screenshot.

shreyanavigyan:
@akulakov Thanks! https://github.com/python/devguide/pull/712 fixes this! Closing this.

PR #712, opened 2021-06-08
Remove superfluous content listing from git bootcamp

akulakov:
Fixes issue #711


Mariatta (approved):
Thanks. Yes, the `contents` directive was a leftover from the previous theme. In the new theme, this is not needed.

PR #714, opened 2021-06-13
bpo-42349: Clarify basicblocks and give some examples

sweeneyde:
This follows dicucssion here: https://bugs.python.org/issue42349

### DIFF
--- a/compiler.rst
+++ b/compiler.rst
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
 Control Flow Graphs
 -------------------
 
-A control flow graph (often referenced by its acronym, CFG) is a
-directed graph that models the flow of a program using basic blocks that
-contain the intermediate representation (abbreviated "IR", and in this
-case is Python bytecode) within the blocks.  Basic blocks themselves are
-a block of IR that has a single entry point but possibly multiple exit
-points.  The single entry point is the key to basic blocks; it all has
-to do with jumps.  An entry point is the target of something that
-changes control flow (such as a function call or a jump) while exit
-points are instructions that would change the flow of the program (such
-as jumps and 'return' statements).  What this means is that a basic
-block is a chunk of code that starts at the entry point and runs to an
-exit point or the end of the block.
+A *control flow graph* (often referenced by its acronym, CFG) is a
+directed graph that models the flow of a program.  A node of a CFG is
+not an individual bytecode instruction, but instead represents a
+sequence of bytecode instructions that always execute sequentially.
+Each node is called a *basic block* and must always execute from
+start to finish, with a single entry point at the beginning and a
+single exit point at the end.  If some bytecode instruction *a* needs
+to jump to some other bytecode instruction *b*, then *a* must occur at
+the end of its basic block, and *b* must occur at the start of its
+basic block.
 
-As an example, consider an 'if' statement with an 'else' block.  The
-guard on the 'if' is a basic block which is pointed to by the basic
-block containing the code leading to the 'if' statement.  The 'if'
-statement block contains jumps (which are exit points) to the true body
-of the 'if' and the 'else' body (which may be ``NULL``), each of which are
-their own basic blocks.  Both of those blocks in turn point to the
-basic block representing the code following the entire 'if' statement.
+As an example, consider the following code snippet:
+
+.. code-block:: Python
+
+   if x < 10:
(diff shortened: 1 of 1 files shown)


sweeneyde:
It looks like @brettcannon wrote this doc originally. The compiler implementation details used to allow multiple exits per basic block, but that is "incorrect" for the common definition of basic block, and it was changed in that issue.

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+.. code-block:: Python
+
+   if x < 10:
+       f1()
+       f2()
+   else:
+       g()
+   end()
+
+The `x < 10` guard is represented by its own basic block that
```
Mariatta, line 313: I think it would be rendered better if you use double backticks instead of single.
```suggestion
The ``x < 10`` guard is represented by its own basic block that
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+compares `x` with `10` and then ends in a conditional jump based on
```
Mariatta, line 314:
```suggestion
compares ``x`` with ``10`` and then ends in a conditional jump based on
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+the result of the comparison.  This conditional jump allows the block
+to point to both the body of the `if` and the body of the `else`.  The
```
Mariatta, line 316:
```suggestion
to point to both the body of the ``if`` and the body of the ``else``.  The
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+`if` basic block contains the `f1()` and `f2()` calls and points to
```
Mariatta, line 317:
```suggestion
``if`` basic block contains the ``f1()`` and ``f2()`` calls and points to
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+the `end()` basic block. The `else` basic block contains the `g()`
```
Mariatta, line 318:
```suggestion
the ``end()`` basic block. The ``else`` basic block contains the ``g()``
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+call and similarly points to the `end()` block.
```
Mariatta, line 319:
```suggestion
call and similarly points to the ``end()`` block.
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+
+Note that more complex code in the guard, the `if` body, or the `else`
```
Mariatta, line 321:
```suggestion
Note that more complex code in the guard, the ``if`` body, or the ``else``
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+body may be represented by multiple basic blocks. For instance,
+short-circuiting boolean logic in a guard like `if x or y:`
```
Mariatta, line 323:
```suggestion
short-circuiting boolean logic in a guard like ``if x or y:``
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+will produce one basic block that tests the truth value of `x`
```
Mariatta, line 324:
```suggestion
will produce one basic block that tests the truth value of ``x``
```

compiler.rst
```diff
@@ -288,26 +288,42 @@ number is passed as the last parameter to each ``stmt_ty`` function.
...
+and then points both (1) to the start of the `if` body and (2) to
```
Mariatta, line 325:
```suggestion
and then points both (1) to the start of the ``if`` body and (2) to
```

Mariatta (commented):
I simply made comment about the markup, did not really review the changed text thoroughly.

I was also wondering if the examples could be something more concrete than *f1* or *a*, *b*, but I don't have better suggestion. Just a thought.

PR #718, opened 2021-06-20
Fixes #717: Example for Running a single test case fails

### LINKED ISSUE #717: Example for Running a single test case fails
**Describe the bug**
When I run `./python -m unittest -v test.test_abc.TestABC` as suggested in the dev-guide,
I get an error:
```
AttributeError: module 'test.test_abc' has no attribute 'TestABC'
```

**Expected behavior**
fix this by adding "_Py" and run it as
`./python -m unittest -v test.test_abc.TestABC_Py`

SimiCode:
Cause of the bug:
- The class seems to be reloaded using a different name at the bottom of `https://github.com/python/cpython/blob/main/Lib/test/test_abc.py`

Fix:
- Modify class name in guide.

### DIFF
--- a/runtests.rst
+++ b/runtests.rst
@@ -48,7 +48,7 @@ verbose mode (using ``-v``), so that individual failures are detailed::
 To run a single test case, use the ``unittest`` module, providing the import
 path to the test case::
 
-   ./python -m unittest -v test.test_abc.TestABC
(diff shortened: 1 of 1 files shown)


Mariatta (approved):
Thanks!

PR #719, opened 2021-06-21
fixed capitalization on PCbuild\build.bat

Harry-Lees:
On Windows, the build instructions prompt the user to run `PCBuild\build.bat`. The correct path is `PCbuild\build.bad` if a user copies the path from the copy block in the docs, it won't find the file correctly.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -258,9 +258,9 @@ are downloaded:
 
 .. code-block:: dosbatch
 
-   PCBuild\build.bat
+   PCbuild\build.bat
 
-After this build succeeds, you can open the ``PCBuild\pcbuild.sln`` solution in
+After this build succeeds, you can open the ``PCbuild\pcbuild.sln`` solution in
 Visual Studio to continue development.
 
 See the `readme`_ for more details on what other software is necessary and how


zware (approved):
Windows is (almost always) on a case-insensitive file system, so this generally doesn't matter.  However, you're right that this is the wrong capitalization, so we might as well fix it :)

Harry-Lees:
> 
> 
> Windows is (almost always) on a case-insensitive file system, so this generally doesn't matter. However, you're right that this is the wrong capitalization, so we might as well fix it :)

I only noticed because I had a failing windows test on another PR :( and I got this nice message from Powershell when trying to build. Perhaps this doesn't happen when using cmd? I'm not sure since I don't use Windows that much for development.

```
PS Z:\home\harry\Github\cpython> PCBuild\build.bat
PCBuild\build.bat : The module 'PCBuild' could not be loaded. For more information, run 'Import-Module PCBuild'.
At line:1 char:1
+ PCBuild\build.bat
+ ~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (PCBuild\build.bat:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CouldNotAutoLoadModule
```

Issue #681, opened 2021-04-29
Fix broken links in the Devguide


Mariatta:
Looks like the `Link check` step in the CI is broken. It has always been allowed to fail for a long time.
Would be great to fix them all.

Please see the output here:
https://github.com/python/devguide/runs/2463428458?check_suite_focus=true

~While looking at the output, I noticed that pypy.org is listed as broken. I think that's the correct url though. I've filed the issue here:~ 

https://github.com/pypy/pypy.org/issues/23
Update: the url seems to work now.

But it would be good to look through the rest and update the urls accordingly.

Mariatta:
I think it would be great if `link check` becomes a required CI check, once we fixed all the urls.

himanshu007-creator:
Hi, i would like to fix the broken links. Please assign me this issue

Mariatta:
We don't assign issues to non core developers. You can just start creating the PR and we go from there.

himanshu007-creator:
There are total 11 broken links. I have fixed 7 of 'em. But there are some links i would like to ask about.

<a href="https://github.com/python/devguide/runs/2463428458?check_suite_focus=true#step:7:65">this </a> and <a href="https://github.com/python/devguide/runs/2463428458?check_suite_focus=true#step:7:66">this</a> link  point to same website and this website just shows a white screen

<a href="https://github.com/python/devguide/runs/2463428458?check_suite_focus=true#step:7:245">This</a> link doesn't work because the repo is either made private or deleted
<a href="https://github.com/python/devguide/runs/2463428458?check_suite_focus=true#step:7:392">This</a> website shows a 522 error


I would like to know how should they be resolved.

Mariatta:
> This website shows a 522 error

Looks like gmane.org moved to https://gmane.io/

> This link doesn't work because the repo is either made private or deleted 
Yes the python/voters repo is a private repo that is available for Python core devs only. So it looks like we will have to allow that the linkcheck to be broken :( Is there a way to tell sphinx linkcheck to ignore this specific url?

> This metadata is not available as mentioned in issue
Hmm yes, I also don't know what the correct URL is. It has been discussed previously in 
  https://github.com/python/devguide/issues/456.

Does anyone know where is the the Issue tracker for the tracker is? I thought that it's now being tracked at github.com/python/bugs.python.org, but this is the issue tracker for our own instance of it. It sounds like there is another issue tracker just for Roundup. But I don't know where that is.

> this and this link point to same website and this website just shows a white screen
[...]

PR #727, opened 2021-07-15
Remove the 'continue-on-error' flag when linkcheck fails

Mariatta:
Let's see if we still have linkcheck errors.


Mariatta:
Looks like this URL is returning 500 error at the moment.
```
(     buildworker: line  178) broken    http://trac.buildbot.net/wiki/RunningBuildbotOnWindows#Service - 500 Server Error: Internal Server Error for url: http://trac.buildbot.net/wiki/RunningBuildbotOnWindows
```

Mariatta:
I think buildbot is not using trac.buildbot.net anymore.

I found the webarchive equivalent of this, not sure if there is a more recent url:
https://web.archive.org/web/20201113084744/http://trac.buildbot.net/wiki/RunningBuildbotOnWindows

zware:
It looks like https://docs.buildbot.net/latest/manual/installation/misc.html#launching-worker-as-windows-service is probably about as close as we'll get to a modern equivalent.

ezio-melotti:
I just noticed this PR while replying to #727.

I think we should make `make html` and `make check` required, and `make linkcheck` optional, since it might fail sporadically.  In order to do this we could:
* Split them in two separate jobs, one required, the other optional;
* Keep them together with `continue-on-error` on the `linkcheck` and make it required

The former would be ideal, but I'm not sure how to set up the workflow to only set up one environment and doing one checkout before running two separate jobs in the same environment.  If that's not possible, I guess we can do it twice, and we could skip the `linkcheck` job if the `build-and-check` job fails.

If we choose the latter, it would be useful to get some notification in case of `linkcheck` failure.  I investigated this in https://github.com/python/devguide/issues/681#issuecomment-1100775734, but I couldn't find a way to report the `linkcheck` failure.

hugovk:
I agree, I don't think we can ever have `linkcheck` as mandatory on normal builds. Links come and go, the network can be flakey.

Instead, it's good to run it as non-failing.

> The former would be ideal, but I'm not sure how to set up the workflow to only set up one environment and doing one checkout before running two separate jobs in the same environment. If that's not possible, I guess we can do it twice, and we could skip the `linkcheck` job if the `build-and-check` job fails.

It's not possible, each job runs in a fresh environment. Having said that, you can upload/download artifacts and use a cache for re-using things.

How about we keep this like this: `html` and `check` required, `linkcheck` runn but non-failing.

* This we can check the current state for each build, go and check if we fixed sone.

And add a (nightly? weekly?) cron that only does `linkcheck` is allowed to fail?

* This way we get a report and can easily see if there's links to fix

One concern: if the cron is often failing (which it might, and often), we don't want it to show the repo as failed at the top:

image

ezio-melotti:
> And add a (nightly? weekly?) cron that only does linkcheck is allowed to fail?

It's somewhat unlikely that recent changes will include invalid links (unless some copy-and-paste mishap happened), so running `linkcheck` on a fixed schedule sounds like a good idea to me.

> One concern: if the cron is often failing (which it might, and often), we don't want it to show the repo as failed at the top:

How does reporting work for this kind of checks that are not directly linked to specific PRs?  Where can the results be reported?

AlexWaygood:
> How does reporting work for this kind of checks that are not directly linked to specific PRs? Where can the results be reported?

Over at typeshed, we have a test that runs every night. If it fails, a bot auto-creates an issue for us. It works pretty well!

Here's the `.yml` file: https://github.com/python/typeshed/blob/master/.github/workflows/daily.yml

And here's an example of an issue that the bot auto-created when the test failed: https://github.com/python/typeshed/issues/7572

hugovk:
> How does reporting work for this kind of checks that are not directly linked to specific PRs? Where can the results be reported?

You'd need to go and check the last run on the cron. Due to frequent flakey failures, I see it more as something to check up on once in a while.

> Over at typeshed, we have a test that runs every night. If it fails, a bot auto-creates an issue for us. It works pretty well!

Sounds good! Although I doubt we'd want a new issue created every day/week (whatever the schedule is)? Perhaps update a single long-lived issue?

Or if it was monthly, it might be fine to open a new one, if we (or the bot) closes the old one?

ezio-melotti:
Removing `continue-on-error` doesn't seem a viable option, so I'm going to close this PR.
I created #917 to track the proposal of adding a workflow that is run regularly.

PR #730, opened 2021-07-23
bpo-44727: Avoid `enum` in the stable ABI

encukou:
Adding a new enumerator to a C enum can change the size of the type,
which would break the ABI.
This is not often a problem in practice, but the rules around when it
is a problem and when it isn't are complicated enough that I believe
enum should not be used in the stable ABI (possibly with well-reasoned
exceptions)

AFAICS, the rules are:
- In C++, an incompatible change to an enum is one that changes the
  size of the *smallest bit field large enough to hold all enumerators*.
  Values outside the range cause undefined/unspecified behavior.
- In C, it looks like enums that fit in `char` are safe.

Also, the compiler-defined size of enums will make it more cumbersome
to formally define an ABI for non-C languages.

### DIFF
--- a/c-api.rst
+++ b/c-api.rst
@@ -124,6 +124,13 @@ Guidelines for changing the Limited API
   - If possible, avoid macros. This makes the Limited API more usable in
     languages that don't use the C preprocessor.
 
+- Think twice when using ``enum``.
+
+  - If a new identifier is added to an ``enum`` in the future, passing it to
+    older extensions can lead to issues (unspecified/undefined behavior in
+    C++; possible size change in C).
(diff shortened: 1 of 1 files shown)


ncoghlan:
What do you think of the approach of splitting the enum definition, so we still have an enum available for internal use, but the public type is an int typedef?

encukou:
Let me close this and continue the discussion on the mailing list: https://mail.python.org/archives/list/python-dev@python.org/thread/RR62MUNEQPH2HLI6YE4VOQH6JQT64LPO/#RNR6AVDT45PEKPY6TLQGF7ZPAB4NRBCD

PR #734, opened 2021-07-31
Add a new guide on how to understand and use the PEG parser

pablogsal:

### DIFF
--- a/parser.rst
+++ b/parser.rst
@@ -1,7 +1,7 @@
 .. _parser:
 
-How to use CPython's Parser
-===========================
+Guide of CPython's Parser
+=========================
 
 :Author: Pablo Galindo Salgado
 
@@ -67,8 +67,7 @@ technique called "packrat parsing" [1]_ which not only loads the entire
 program in memory before parsing it but also allows the parser to backtrack
 arbitrarily. This is made efficient by memoizing the rules already matched for
 each position. The cost of the memoization cache is that the parser will naturally
-use more memory than a simple LL(1) parser, which normally are table-based. We
-will explain later in this document why we consider this cost acceptable.
+use more memory than a simple LL(1) parser, which normally are table-based. 
 
 
 Key ideas
@@ -745,12 +744,30 @@ two phases:
     * Make sure **all** invalid rules start with the ``invalid_`` prefix to not
       impact performance of parsing correct Python code.
     * Make sure the parser doesn't behave differently for regular rules when you introduce invalid rules
-      (see the :ref:`how PEG parsers work section <how-peg-parsers-work>` for more information)>
+      (see the :ref:`how PEG parsers work section <how-peg-parsers-work>` for more information).
 
 You can find a collection of macros to raise specialized syntax errors in the
 :file:`Parser/pegen.h` header file. These macros allow also to report ranges for
 the custom errors that will be highlighted in the tracebacks that will be
 displayed when the error is reported.
+
+.. tip::
+    A good way to test if an invalid rule will be triggered when you expect is to test if introducing
+    a syntax error **after** valid code triggers or not the rule. For example: ::
+
+        <valid python code> $ 42
+    
+    Should trigger the syntax error in the ``$`` character. If your rule is not correctly defined this
+    won't happen. For example, if you try to define a rule to match Python 2 style ``print`` statements
+    to make a better error message and you define it as: ::
+
+        invalid_print: "print" expression
+    
+    This will **seem** to work because the parser will correctly parse ``print(something)`` because is valid
+    code and the second phase will never execute but if you try to parse ``print(something) $ 3`` the first pass
+    of the parser will fail (because of the ``$``) and in the second phase, the rule will match the
+    ``print(something)`` as ``print`` followed by the variable ``something`` between parenthesis and the error
+    will be reported there instead in the ``$`` character.
 
 Generating AST objects
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/parser.rst
+++ b/parser.rst
@@ -83,9 +83,9 @@ Key ideas
 * Alternatives are ordered ( ``A | B`` is not the same as ``B | A`` ).
 * If a rule returns a failure doesn't mean that the parsing has failed,
   it just means "try something else".
-* By default PEG parsers run in exponential time and they are keep linear By
+* By default PEG parsers run in exponential time, which can be optimized to linear by
   using memoization.
-* If parsing fails completely (no rule can succeed parsing all the input text), the
+* If parsing fails completely (no rule succeeds in parsing all the input text), the
   PEG parser doesn't have a concept of "where the syntax error is".
 
 
@@ -100,7 +100,7 @@ parsers work) has deep consequences, other than removing ambiguity.
 
 If a rule has two alternatives and the first of them succeeds, the second one is
 **not** attempted even if the caller rule fails to parse the rest of the input.
-This is often described that the parser is "eager". To illustrate this, consider
+Thus the parser is said to be "eager". To illustrate this, consider
 the following two rules (in these examples, a token is an individual character): ::
 
     first_rule:  ( 'a' | 'aa' ) 'a'
@@ -122,7 +122,7 @@ tried.
 
     The effects of ordered choice such as illustrated above, may be hidden by many levels of rules.
 
-For this reason, avoid writing rules where an alternative is contained in the next one,
+For this reason, writing rules where an alternative is contained in the next one is in almost all cases a mistake,
 for example: ::
 
     my_rule:
@@ -500,7 +500,7 @@ A similar grammar written to target Python AST objects:
 Pegen
 -----
 
-Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. Is the
+Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. It is the
 program that can be used to read the python grammar located in :file:`Grammar/Python.gram` and produce the final C
 parser. It contains the following pieces:
 
@@ -534,7 +534,7 @@ How to regenerate the meta-parser
 
 The meta-grammar (the grammar that describes the grammar for the grammar files
 themselves) is located at :file:`Tools/peg_generator/pegen/metagrammar.gram`.
-Although is very unlikely that you ever need to modify it, if you make any modifications
+Although it is very unlikely that you will ever need to modify it, if you make any modifications
 to this file (in order to implement new Pegen features) you will need to regenerate
 the meta-parser (the parser that parses the grammar files). To do so just execute: ::
 
@@ -592,7 +592,7 @@ Memoization
 
 As described previously, to avoid exponential time complexity in the parser, memoization is used. In the Python
 
-The C parser used by Python is highly optimized and memoization can expensive both in memory and time. Although
+The C parser used by Python is highly optimized and memoization can be expensive both in memory and time. Although
 the memory cost is obvious (the parser needs memory for storing previous results in the cache) the execution time
 cost comes for continuously checking if the given rule has a cache hit or not. In many situations, just parsing it
 again can be faster. Pegen **disables memoization by default** except for rules with the special marker `memo` after
@@ -601,7 +601,7 @@ the rule name (and type, if present): ::
     rule_name[typr] (memo):
         ...
 
-Selectively turning on memoization for a handful of rules, the parser becomes faster and uses less memory.
+By selectively turning on memoization for a handful of rules, the parser becomes faster and uses less memory.
 
 .. note::
     Left-recursive rules always use memoization, since the implementation of left-recursion depends on it.
@@ -616,7 +616,7 @@ needs to be manually activated.
 Automatic variables
 ~~~~~~~~~~~~~~~~~~~
 
-To make writing actions easier, Pegen inject some automatic variables in the namespace available
+To make writing actions easier, Pegen injects some automatic variables in the namespace available
 when writing actions. In the C parser, some of these automatic variable names are:
 
 * ``p``: The parser structure.
@@ -700,7 +700,7 @@ error messages.
 
 .. note::
     Tokenizer errors are normally reported by raising exceptions but some special
-    tokenizer errors such as unclosed parenthesis will be reported only after themselves
+    tokenizer errors such as unclosed parenthesis will be reported only after the
     parser finishes without returning anything.
 
 How Syntax errors are reported
@@ -708,10 +708,10 @@ How Syntax errors are reported
 
 As described previously in the :ref:`how PEG parsers work section
 <how-peg-parsers-work>`, PEG parsers doesn't have a defined concept of where
-errors happened in the grammar, because if a rule fails that doesn't imply a
+errors happened in the grammar, because a rule failure doesn't imply a
 parsing failure like in context free grammars. This means that some heuristic
 has to be used to report generic errors unless something is explicitly declared
-as an error in there grammar.
+as an error in the grammar.
 
 To report generic syntax errors, pegen uses a common heuristic in PEG parsers:
 the location of *generic* syntax errors is reported in the furthest token that
@@ -721,18 +721,18 @@ been raised.
 
 .. caution::
     Positive and negative lookaheads will try to match a token so they will affect
-    the location of generic syntax errors. Use them carefully around at boundaries
+    the location of generic syntax errors. Use them carefully at boundaries
     between rules.
 
 As the Python grammar was primordially written as an LL(1) grammar, this heuristic
-has an extremely high success rate, but some of PEG features can have small effects,
+has an extremely high success rate, but some PEG features can have small effects,
 such as :ref:`positive lookaheads <peg-positive-lookahead>` and
 :ref:`negative lookaheads <peg-negative-lookahead>`.
 
 To generate more precise syntax errors, custom rules are used. This is a common practice
 also on context free grammars: the parser will try to accept some construct that is known
 to be incorrect just to report a specific syntax error for that construct. In pegen grammars,
-these rules start with ``invalid_`` prefix. This is because trying to match this rules
+these rules start with ``invalid_`` prefix. This is because trying to match these rules
 normally has a performance impact on parsing (and can also affect the 'correct' grammar itself
 in some tricky cases, depending on the ordering of the rules) so the generated parser acts in
 two phases:
@@ -743,9 +743,9 @@ two phases:
 
 2. If the first phase failed, a second parsing attempt is done including the rules that start
    with an ``invalid_`` prefix. By design this attempt **cannot succeed** and is only executed
-   to give it a change to the invalid rules to detect specific situations where custom, more precise,
+   to give to the invalid rules a chance to detect specific situations where custom, more precise,
    syntax errors can be raised. This also allows to trade a bit of performance for precision reporting
-   errors: given that we know that the input text is invalid, so there is no need to be fast because
+   errors: given that we know that the input text is invalid, there is no need to be fast because
    the interpreter is going to stop anyway.
 
 .. important::
@@ -764,7 +764,7 @@ displayed when the error is reported.
 
 .. tip::
     A good way to test if an invalid rule will be triggered when you expect is to test if introducing
-    a syntax error **after** valid code triggers or not the rule. For example: ::
+    a syntax error **after** valid code triggers the rule or not. For example: ::
 
         <valid python code> $ 42
     
@@ -778,7 +778,7 @@ displayed when the error is reported.
     code and the second phase will never execute but if you try to parse ``print(something) $ 3`` the first pass
     of the parser will fail (because of the ``$``) and in the second phase, the rule will match the
     ``print(something)`` as ``print`` followed by the variable ``something`` between parenthesis and the error
-    will be reported there instead in the ``$`` character.
+    will be reported there instead of the ``$`` character.
 
 Generating AST objects
 ~~~~~~~~~~~~~~~~~~~~~~
@@ -787,7 +787,7 @@ The output of the C parser used by CPython that is generated by the
 :file:`Grammar/Python.gram` grammar file is a Python AST object (using C
 structures). This means that the actions in the grammar file generate AST objects
 when they succeed. Constructing these objects can be quite cumbersome (see
-themselves :ref:`AST compiler section <compiler-ast-trees>` for more information
+the :ref:`AST compiler section <compiler-ast-trees>` for more information
 on how these objects are constructed and how they are used by the compiler) so
 special helper functions are used. These functions are declared in the
 :file:`Parser/pegen.h` header file and defined in the :file:`Parser/pegen.c`
@@ -834,7 +834,7 @@ As the generated C parser is the one used by Python, this means that if somethin
 new rules to the grammar you cannot correctly compile and execute Python anymore. This makes it a bit challenging
 to debug when something goes wrong, specially when making experiments.
 
-For this reason, is a good idea to experiment first by generating a Python parser. To do this, you can go to the
+For this reason it is a good idea to experiment first by generating a Python parser. To do this, you can go to the
 :file:`Tools/peg_generator/` directory on the CPython repository and manually call the parser generator by executing:
 
 .. code-block:: shell
@@ -861,8 +861,8 @@ can be a bit hard to understand at first.
 
 .. note::
 
-    When activating verbose mode in the Python parser, is better to not use interactive mode as it can be much harder to
-    understand, as interactive mode involves some special steps compared with regular parsing.
+    When activating verbose mode in the Python parser, it is better to not use interactive mode as it can be much harder to
+    understand, because interactive mode involves some special steps compared to regular parsing.
 
 To activate verbose mode you can add the ``-d`` flag when executing Python:
 
@@ -885,7 +885,7 @@ character marks the type of the trace:
 
 The <token_location> part indicates the current index in the token array, the
 <rule_name> part indicates what rule is being parsed and the <alternative> part
-indicates what alternative withing that rule is being attempted.
+indicates what alternative within that rule is being attempted.
 
 
 References
--- a/parser.rst
+++ b/parser.rst
@@ -61,7 +61,7 @@ Notice that "failure" results do not imply that the program is incorrect or a
 parsing failure because as the choice operator is ordered, a "failure" result
 merely indicates "try the following option". A direct implementation of a PEG
 parser as a recursive descent parser will present exponential time performance
-the worst case, because PEG parsers have infinite lookahead (this means that
+in the worst case, because PEG parsers have infinite lookahead (this means that
 they can consider an arbitrary number of tokens before deciding for a rule).
 Usually, PEG parsers avoid this exponential time complexity with a technique called
 "packrat parsing" [1]_ which not only loads the entire program in memory before
@@ -78,11 +78,11 @@ Key ideas
     Don't try to reason about a PEG grammar in the same way you would to with an EBNF
     or context free grammar. PEG is optimized to describe **how** input strings will
     be parsed, while context-free grammars are optimized to generate strings of the
-    language they describe (in EBNF, to know if given string is in the language you need
-    to do work to find out as is not immediately obvious from the grammar).
+    language they describe (in EBNF, to know if a given string is in the language, you need
+    to do work to find out as it is not immediately obvious from the grammar).
 
 * Alternatives are ordered ( ``A | B`` is not the same as ``B | A`` ).
-* If a rule returns a failure doesn't mean that the parsing has failed,
+* If a rule returns a failure, it doesn't mean that the parsing has failed,
   it just means "try something else".
 * By default PEG parsers run in exponential time, which can be optimized to linear by
   using memoization.
@@ -121,7 +121,7 @@ tried.
 
 .. caution::
 
-    The effects of ordered choice such as illustrated above, may be hidden by many levels of rules.
+    The effects of ordered choice, such as the ones illustrated above, may be hidden by many levels of rules.
 
 For this reason, writing rules where an alternative is contained in the next one is in almost all cases a mistake,
 for example: ::
@@ -131,7 +131,7 @@ for example: ::
         | 'if' expression 'then' block 'else' block
 
 In this example, the second alternative will never be tried because the first one will
-succeed first (even if the input string has a ``'else' block`` that follows). To correctly
+succeed first (even if the input string has an ``'else' block`` that follows). To correctly
 write this rule you can simply alter the order: ::
 
     my_rule:
@@ -609,7 +609,7 @@ By selectively turning on memoization for a handful of rules, the parser becomes
 To know if a new rule needs memoization or not, benchmarking is required
 (comparing execution times and memory usage of some considerably big files with
 and without memoization). There is a very simple instrumentation API available
-in the C generated parse code that allows to measure how much each rule uses
+in the generated C parse code that allows to measure how much each rule uses
 memoization (check the :file:`Parser/pegen.c` file for more information) but it
 needs to be manually activated.
 
@@ -650,7 +650,7 @@ keyword as a variable will always fail:
             ^^^^^
     SyntaxError: invalid syntax
 
-While soft keywords don't have this limitation if used in a context other when they
+While soft keywords don't have this limitation if used in a context other the one where they
 are defined as keywords:
 
 .. code-block:: python
@@ -684,7 +684,7 @@ as well as soft keywords:
 .. caution::
     Soft keywords can be a bit challenging to manage as they can be accepted in
     places you don't intend to, given how the order alternatives behave in PEG
-    parser (see :ref:`consequences of ordered choice section
+    parsers (see :ref:`consequences of ordered choice section
     <consequences-of-ordered-choice>` for some background on this). In general,
     try to define them in places where there is not a lot of alternatives.
 
@@ -709,7 +709,7 @@ How Syntax errors are reported
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 As described previously in the :ref:`how PEG parsers work section
-<how-peg-parsers-work>`, PEG parsers doesn't have a defined concept of where
+<how-peg-parsers-work>`, PEG parsers don't have a defined concept of where
 errors happened in the grammar, because a rule failure doesn't imply a
 parsing failure like in context free grammars. This means that some heuristic
 has to be used to report generic errors unless something is explicitly declared
@@ -732,9 +732,9 @@ such as :ref:`positive lookaheads <peg-positive-lookahead>` and
 :ref:`negative lookaheads <peg-negative-lookahead>`.
 
 To generate more precise syntax errors, custom rules are used. This is a common practice
-also on context free grammars: the parser will try to accept some construct that is known
+also in context free grammars: the parser will try to accept some construct that is known
 to be incorrect just to report a specific syntax error for that construct. In pegen grammars,
-these rules start with ``invalid_`` prefix. This is because trying to match these rules
+these rules start with the ``invalid_`` prefix. This is because trying to match these rules
 normally has a performance impact on parsing (and can also affect the 'correct' grammar itself
 in some tricky cases, depending on the ordering of the rules) so the generated parser acts in
 two phases:
@@ -753,7 +753,7 @@ two phases:
 .. important::
     When defining invalid rules:
 
-    * Make sure all custom invalid rules raise :exc:`SyntaxError` exceptions (or a subclass if it).
+    * Make sure all custom invalid rules raise :exc:`SyntaxError` exceptions (or a subclass of it).
     * Make sure **all** invalid rules start with the ``invalid_`` prefix to not
       impact performance of parsing correct Python code.
     * Make sure the parser doesn't behave differently for regular rules when you introduce invalid rules
@@ -779,7 +779,7 @@ displayed when the error is reported.
     This will **seem** to work because the parser will correctly parse ``print(something)`` because it is valid
     code and the second phase will never execute but if you try to parse ``print(something) $ 3`` the first pass
     of the parser will fail (because of the ``$``) and in the second phase, the rule will match the
-    ``print(something)`` as ``print`` followed by the variable ``something`` between parenthesis and the error
+    ``print(something)`` as ``print`` followed by the variable ``something`` between parentheses and the error
     will be reported there instead of the ``$`` character.
 
 Generating AST objects
@@ -834,7 +834,7 @@ Making experiments
 
 As the generated C parser is the one used by Python, this means that if something goes wrong when adding some
 new rules to the grammar you cannot correctly compile and execute Python anymore. This makes it a bit challenging
-to debug when something goes wrong, specially when making experiments.
+to debug when something goes wrong, especially when making experiments.
 
 For this reason it is a good idea to experiment first by generating a Python parser. To do this, you can go to the
 :file:`Tools/peg_generator/` directory on the CPython repository and manually call the parser generator by executing:
@@ -873,7 +873,7 @@ To activate verbose mode you can add the ``-d`` flag when executing Python:
     $ python -d file_to_test.py
 
 This will print **a lot** of output to ``stderr`` so is probably better to dump it to a file for further analysis. The output
-consist of trace lines with the following structure:
+consists of trace lines with the following structure:
 
     <indentation> ('>'|'-'|'+'|'!') <rule_name>[<token_location>]: <alternative> ...
 
--- a/parser.rst
+++ b/parser.rst
@@ -100,7 +100,7 @@ that in PEG parsers alternatives are ordered (which is the core of how PEG
 parsers work) has deep consecuences, other than removing ambiguity.
 
 If a rule has two alternatives and the first of them suceeds, the second one is **not** attempted
-even if the caller rule failes to parse the rest of the input. This is often described that the parser
+even if the caller rule fails to parse the rest of the input. This is often described that the parser
 is "eager". To illustrate, this, consider the following two rules: ::
 
     first_rule:  ( 'a' | 'aa' ) 'a'
--- a/parser.rst
+++ b/parser.rst
@@ -550,7 +550,7 @@ Although it is very unlikely that you will ever need to modify it, if you make a
 to this file (in order to implement new Pegen features) you will need to regenerate
 the meta-parser (the parser that parses the grammar files). To do so just execute: ::
 
-    make regen-metaparser
+    make regen-pegen-metaparser
 
 If you are on Windows you can use the Visual Studio project files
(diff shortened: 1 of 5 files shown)


terryjreedy:
I cannot review and make suggestions while you are pushing multiple changes.  Let me know when done for a while.

pablogsal:
> I cannot review and make suggestions while you are pushing multiple changes. Let me know when done for a while.

Something is wrong with the CI, I am trying to understand what's going on first. Will mark it as draft meanwhile

pablogsal:
I don't understand this error:

```
Run sphinx-build -n -W -q -b html -d _build/doctrees . _build/html
  sphinx-build -n -W -q -b html -d _build/doctrees . _build/html
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.6/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.6/x64/lib

Extension error (furo):
Handler <function _overwrite_pygments_css at 0x7f1b0f26e670> for event 'build-finished' threw an exception (exception: [Errno 2] No such file or directory: '/home/runner/work/devguide/devguide/_build/html/_static/pygments.css')
Error: Process completed with exit code 2.

```

@pradyunsg any idea of what's going on? I cannot reproduce this locally :(

pablogsal:
I am unable to reproduce the `furo` failure in the CI. Hopefully @pradyunsg can explain what's going on here, meanwhile, the text itself is ready for review.

pablogsal:
I am a bit suspicious that this is due to this commit:

https://github.com/pradyunsg/furo/commit/d467db25edc515bd240fd13e803d6a617a7412f1

I am going to pin to furo<2021.07.28.beta39 to avoid the problem due to the new code.

pablogsal:
@terryjreedy ready for review!

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+reflects how the parser will operate when parsing it. The fundamental technical
+difference is that the choice operator is ordered. This means that when writing::
+
+  rule: A | B | C
+
+a context-free-grammar parser (like an LL(1) parser) will generate constructions
+that given an input string will *deduce* which alternative (``A``, ``B`` or ``C``)
+must be expanded, while a PEG parser will check if the first alternative succeeds
+and only if it fails, will it continue with the second or the third one in the
+order in which they are written. This makes the choice operator not commutative.
```
terryjreedy, line 42: This is basically the same as order being important in ``and`` and ``or`` expressions and sequences of conditions in ``if`` and ``match`` statements.  Once I saw this, understanding the rest became easy.  So I think adding a sentence here would be useful.

Some theoretical computer scientists have advocated that multiple conditions be considered unordered like parsing patterns, so that multiple matches are possible, making a statement ambiguous.

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+an input string as its argument, and yields one of the following results:
+
+* A "success" result. This result indicates that the expression can be parsed by
+  that rule and the function may optionally move forward or consume one or more
+  characters of the input string supplied to it.
+* A "failure" result, in which case no input is consumed.
+
+Notice that "failure" results do not imply that the program is incorrect or a
+parsing failure because as the choice operator is ordered, a "failure" result
+merely indicates "try the following option". A direct implementation of a PEG
```
terryjreedy, line 61: This seems directed at people who are so enmeshed in current parsers (as oppose to programming), that sequential rather thanparallel testing or matching seems weird rather than natural.

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. _consecuences-of-ordered-choice:
+
+Although PEG may look like EBNF, but its meaning is quite different. The fact
+that in PEG parsers alternatives are ordered (which is the core of how PEG
+parsers work) has deep consecuences, other than removing ambiguity.
+
+If a rule has two alternatives and the first of them suceeds, the second one is **not** attempted
+even if the caller rule failes to parse the rest of the input. This is often described that the parser
```
terryjreedy, line 103: 'failes'
```suggestion
even if the caller rule fails to parse the rest of the input. This is often described that the parser
```

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+is "eager". To illustrate, this, consider the following two rules: ::
+
+    first_rule:  ( 'a' | 'aa' ) 'a'
+    second_rule: ('aa' | 'a'  ) 'a'
+
+In a relgular EBNF grammar, both rules specify the language ``{aa, aaa}`` but
+in PEG, one of these two rules accepts the string ``aaa`` but not the string
+``aa``. The other does the opposite -- it accepts the string the string ``aa``
+but not the string ``aaa``. The reason is that the rule ``('a'/'aa')'a'`` does
```
terryjreedy, line 112:
```suggestion
but not the string ``aaa``. The rule ``('a'/'aa')'a'`` does
```

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+not accept ``aaa`` because ``'a'/'aa'`` consumes the first ``a``, letting the
+final ``a`` in the rule consume the second, and leaving out the third ``a``.
+As the rule has suceeded, no attempt is ever made to go back and let
+``'a'/'aa'`` try the second alternative. The expression ``('aa'/'a')'a'`` does
+not accept ``aa`` because ``'aa'/'a'`` accepts all of ``aa``, leaving nothing
+for the final ``a``. Again, the second alternative of ``'aa'/'a'`` is not
+tried.
```
terryjreedy, line 119: A subtle point here is that catenation does not distribute across choices.  I presume that ``'aaa'|'aa' *will* match ``aa`` because upon failure of ``aaa`` to match, ``aa`` will be tried, so that the pattern is *not* equal to ``('aa'/'a')'a'``.

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+        | 'if' expression 'then' block
+        | 'if' expression 'then' block 'else' block
+
+In this example, the second alternative will never be tried because the first one will
+suceed first (even if the input string has a ``'else' block`` that follows). To correctly
+write this rule you can simply alter the order: ::
+
+    my_rule:
+        | 'if' expression 'then' block
+        | 'if' expression 'then' block 'else' block
```
terryjreedy, line 138:
```suggestion
        | 'if' expression 'then' block 'else' block
        | 'if' expression 'then' block
```

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+``e1 e2``
+'''''''''
+
+Match e1, then match e2.
+
+::
+
+    rule_name: first_rule second_rule
+
+.. _e1-e2-1:
```
terryjreedy, line 176: I don't understand this block.  It may be my fuzziness about .rst markup, but something seems extraneous or missing.  Same comment about other .. e1...(underscored) lines below.

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+``Any`` in Python.
+
+Grammar Expressions
+~~~~~~~~~~~~~~~~~~~
+
+``# comment``
+'''''''''''''
+
+Python-style comments.
+
```
terryjreedy, line 166: Is '::' missing here?  I see it at the beginning of other grammar expressions.

parser.rst
```diff
@@ -0,0 +1,884 @@
...
+    rule_name: (e)
+
+A slightly more complex and useful example includes using the grouping
+operator together with the repeat operators:
+
+::
+
+    rule_name: (e1 e2)*
+
```
terryjreedy, line 208: The 'black' explanation sentence and '::' seem to be missing.

I stopped a little below here.  Will read more later.

terryjreedy (changes requested):

parser.rst
pablogsal: In the word or grammars the "|" operator doesn't imply parallel or sequential testing because the grammar is *not* describing how strings in the language will be matched. What makes PEG unique is they the parser is describing how to generate a parser, so there is a much closer relationship between the grammar and "execution".

parser.rst
pablogsal: Yes and no. Grammars in general doesn't describe execution or matching. Indeed, EBNF grammars don't describe how to mach strings in the language but how to generate them. Algorithmic transformation is required to know how an EBNF grammar matches strings. PEG is unique here in the sense that th grammar is more related to the execution of the parser so what you mean is indeed closer, but care must be taken because it doesn't translate well to general grammars.

parser.rst
pablogsal: > so that the pattern is *not* equal to ``('aa'/'a')'a'``.

Indeed, that is the consequence of the parser being "eager" and the ordered alternative operator. The key here is that groping is its own rule (so when you write `('aa'/'a')'a'` you are writing `rule_1 'a'`. And the alternative eagerness is evaluated at the rule granularity

parser.rst
pablogsal: No, ```# comment``` is a section name,  `Python-style comments.` is the description of that section and ```e1 e2``` is another section name. Check the generated docs to see how it looks

parser.rst
pablogsal: These are unused labels, I will remove them

parser.rst
pablogsal: > The 'black' explanation sentence and '::' seem to be missing.

What do you mean with the `black` explanation?

terryjreedy: The black type in the diff.

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+        <valid python code> $ 42
+    
+    Should trigger the syntax error in the ``$`` character. If your rule is not correctly defined this
+    won't happen. For example, if you try to define a rule to match Python 2 style ``print`` statements
+    to make a better error message and you define it as: ::
+
+        invalid_print: "print" expression
+    
+    This will **seem** to work because the parser will correctly parse ``print(something)`` because is valid
```
terryjreedy, line 777:
```suggestion
    This will **seem** to work because the parser will correctly parse ``print(something)`` because it is valid
```

parser.rst
```diff
@@ -568,8 +568,8 @@ Pegen has some special grammatical elements and rules:
 Tokenization
 ~~~~~~~~~~~~
 
-Is commong among PEG parser frameworks that the parser does both the parsing and the tokenization,
-but this does not happen in pegen. The reason is that the Python language needs a custom tokenizer
+Is common among PEG parser frameworks that the parser does both the parsing and the tokenization,
```
terryjreedy, line 571:
```suggestion
It is common among PEG parser frameworks that the parser does both the parsing and the tokenization,
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+* A "success" result. This result indicates that the expression can be parsed by
+  that rule and the function may optionally move forward or consume one or more
+  characters of the input string supplied to it.
+* A "failure" result, in which case no input is consumed.
+
+Notice that "failure" results do not imply that the program is incorrect or a
+parsing failure because as the choice operator is ordered, a "failure" result
+merely indicates "try the following option". A direct implementation of a PEG
+parser as a recursive descent parser will present exponential time performance in
+the worst case as compared with LL(1) parsers, because PEG parsers have infinite lookahead
```
lysnikolaou, line 63: I don't really understand what "as compared with LL(1) parsers" means here. What are we comparing? Is your intention to say that "exponential time is much worse than the typical implementation of an LL(1)"? If yes, it's not easy to get that here.

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+    Don't try to reason about a PEG grammar in the same way you would to with an EBNF
+    or context free grammar. PEG is optimized to describe **how** input strings will
+    be parsed, while context-free grammars are optimized to generate strings of the
+    language they describe (in EBNF, to know if given string is in the language you need
+    to do work to find out as is not immediately obvious from the grammar).
+
+* Alternatives are ordered ( ``A | B`` is not the same as ``B | A`` ).
+* If a rule returns a failure doesn't mean that the parsing has failed,
+  it just means "try something else".
+* By default PEG parsers run in exponential time and they are keep linear By
```
lysnikolaou, line 86:
```suggestion
* By default PEG parsers run in exponential time, which can be optimized to linear by
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+  using memoization.
+* If parsing fails completely (no rule can succeed parsing all the input text), the
```
lysnikolaou, line 88:
```suggestion
* If parsing fails completely (no rule succeeds in parsing all the input text), the
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+.. _consequences-of-ordered-choice:
+
+Although PEG may look like EBNF, its meaning is quite different. The fact
+that in PEG parsers alternatives are ordered (which is at the core of how PEG
+parsers work) has deep consequences, other than removing ambiguity.
+
+If a rule has two alternatives and the first of them succeeds, the second one is
+**not** attempted even if the caller rule fails to parse the rest of the input.
+This is often described that the parser is "eager". To illustrate this, consider
```
lysnikolaou, line 103:
```suggestion
Thus the parser is said to be "eager". To illustrate this, consider
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+``'a'|'aa'`` try the second alternative. The expression ``('aa'|'a')'a'`` does
+not accept ``aa`` because ``'aa'|'a'`` accepts all of ``aa``, leaving nothing
+for the final ``a``. Again, the second alternative of ``'aa'|'a'`` is not
+tried.
+
+.. caution::
+
+    The effects of ordered choice such as illustrated above, may be hidden by many levels of rules.
+
+For this reason, avoid writing rules where an alternative is contained in the next one,
```
lysnikolaou, line 125: I think that the wording here should be stronger than just "avoid writing this". Maybe something like this?

```suggestion
For this reason, writing rules where an alternative is contained in the next one is in almost all cases a mistake,
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+        | f=factor { f }
+
+    factor[expr_ty]: 
+        | '(' e=expr ')' { e }
+        | a=atom { a }
+
+    atom[expr_ty]: 
+        | n=NAME { n }
+        | n=NUMBER { n }
+        | s=STRING { s }
```
lysnikolaou, line 469: Let's remove the string here, since it doesn't appear in the Python example below and we're talking about simple arithmetic expressions.

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+    atom: 
+        | NAME { ast.Name(id=name.string, ctx=ast.Load()) }
+        | NUMBER { ast.Constant(value=ast.literal_eval(number.string)) }
+
+
+Pegen
+-----
+
+Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. Is the
```
lysnikolaou, line 503:
```suggestion
Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. It is the
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+    ./PCbuild/build.bat --regen
+
+The generated parser file is located at :file:`Parser/parser.c`.
+
+How to regenerate the meta-parser
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The meta-grammar (the grammar that describes the grammar for the grammar files
+themselves) is located at :file:`Tools/peg_generator/pegen/metagrammar.gram`.
+Although is very unlikely that you ever need to modify it, if you make any modifications
```
lysnikolaou, line 537:
```suggestion
Although it is very unlikely that you will ever need to modify it, if you make any modifications
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+    ./PCbuild/build.bat --regen
+
+How tokens are generated and the rules governing this is completely up to the tokenizer (:file:`Parser/tokenizer.c`)
+and the parser just receives tokens from it.
+
+Memoization
+~~~~~~~~~~~
+
+As described previously, to avoid exponential time complexity in the parser, memoization is used. In the Python
```
lysnikolaou, line 593: There seems to be something missing here?

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+The C parser used by Python is highly optimized and memoization can expensive both in memory and time. Although
```
lysnikolaou, line 595:
```suggestion
The C parser used by Python is highly optimized and memoization can be expensive both in memory and time. Although
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+the memory cost is obvious (the parser needs memory for storing previous results in the cache) the execution time
+cost comes for continuously checking if the given rule has a cache hit or not. In many situations, just parsing it
+again can be faster. Pegen **disables memoization by default** except for rules with the special marker `memo` after
+the rule name (and type, if present): ::
+
+    rule_name[typr] (memo):
+        ...
+
+Selectively turning on memoization for a handful of rules, the parser becomes faster and uses less memory.
```
lysnikolaou, line 604:
```suggestion
By selectively turning on memoization for a handful of rules, the parser becomes faster and uses less memory.
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+The ``match`` and ``case`` keywords are soft keywords, so that they are recognized as
+keywords at the beginning of a match statement or case block respectively, but are
+allowed to be used in other places as variable or argument names.
+
+You can get a list of all keywords defined in the grammar from Python:
+
+.. code-block:: python
+
+    >>> keyword.kwlist
```
lysnikolaou, line 669: It might be obvious, but what about adding an import in this code-block?

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+:ref:`rule action <peg-grammar-actions>` raises an exception all parsing will
+stop at that exact point. This is done to allow to correctly propagate any
+exception set by calling Python C-API functions. This also includes :exc:`SyntaxError`
+exceptions and this is the main mechanism the parser uses to report custom syntax
+error messages.
+
+.. note::
+    Tokenizer errors are normally reported by raising exceptions but some special
+    tokenizer errors such as unclosed parenthesis will be reported only after themselves
+    parser finishes without returning anything.
```
lysnikolaou, line 704:
```suggestion
    tokenizer errors such as unclosed parenthesis will be reported only after the
    parser finishes without returning anything.
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+How Syntax errors are reported
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As described previously in the :ref:`how PEG parsers work section
+<how-peg-parsers-work>`, PEG parsers doesn't have a defined concept of where
+errors happened in the grammar, because if a rule fails that doesn't imply a
```
lysnikolaou, line 711:
```suggestion
errors happened in the grammar, because s rule failure doesn't imply a
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+parsing failure like in context free grammars. This means that some heuristic
+has to be used to report generic errors unless something is explicitly declared
+as an error in there grammar.
```
lysnikolaou, line 714:
```suggestion
as an error in the grammar.
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+To report generic syntax errors, pegen uses a common heuristic in PEG parsers:
+the location of *generic* syntax errors is reported in the furthest token that
+was attempted to be matched but failed. This is only done if parsing has failed
+(the parser returns ``NULL`` in C or ``None`` in Python) but no exception has
+been raised.
+
+.. caution::
+    Positive and negative lookaheads will try to match a token so they will affect
+    the location of generic syntax errors. Use them carefully around at boundaries
```
lysnikolaou, line 724:
```suggestion
    the location of generic syntax errors. Use them carefully at boundaries
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+    between rules.
+
+As the Python grammar was primordially written as an LL(1) grammar, this heuristic
+has an extremely high success rate, but some of PEG features can have small effects,
```
lysnikolaou, line 728:
```suggestion
has an extremely high success rate, but some PEG features can have small effects,
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+such as :ref:`positive lookaheads <peg-positive-lookahead>` and
+:ref:`negative lookaheads <peg-negative-lookahead>`.
+
+To generate more precise syntax errors, custom rules are used. This is a common practice
+also on context free grammars: the parser will try to accept some construct that is known
+to be incorrect just to report a specific syntax error for that construct. In pegen grammars,
+these rules start with ``invalid_`` prefix. This is because trying to match this rules
```
lysnikolaou, line 735:
```suggestion
these rules start with ``invalid_`` prefix. This is because trying to match these rules
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+in some tricky cases, depending on the ordering of the rules) so the generated parser acts in
+two phases:
+
+1. The first phase will try to parse the input stream without taking into account rules that
+   start with the ``invalid_`` prefix. If the parsing succeeds it will return the generated AST
+   and the second phase will not be attempted.
+
+2. If the first phase failed, a second parsing attempt is done including the rules that start
+   with an ``invalid_`` prefix. By design this attempt **cannot succeed** and is only executed
+   to give it a change to the invalid rules to detect specific situations where custom, more precise,
```
lysnikolaou, line 746:
```suggestion
   to give to the invalid rules a chance to detect specific situations where custom, more precise,
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+   syntax errors can be raised. This also allows to trade a bit of performance for precision reporting
+   errors: given that we know that the input text is invalid, so there is no need to be fast because
```
lysnikolaou, line 748:
```suggestion
   errors: given that we know that the input text is invalid, there is no need to be fast because
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+      (see the :ref:`how PEG parsers work section <how-peg-parsers-work>` for more information).
+
+You can find a collection of macros to raise specialized syntax errors in the
+:file:`Parser/pegen.h` header file. These macros allow also to report ranges for
+the custom errors that will be highlighted in the tracebacks that will be
+displayed when the error is reported.
+
+.. tip::
+    A good way to test if an invalid rule will be triggered when you expect is to test if introducing
+    a syntax error **after** valid code triggers or not the rule. For example: ::
```
lysnikolaou, line 767:
```suggestion
    a syntax error **after** valid code triggers the rule or not. For example: ::
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+    won't happen. For example, if you try to define a rule to match Python 2 style ``print`` statements
+    to make a better error message and you define it as: ::
+
+        invalid_print: "print" expression
+    
+    This will **seem** to work because the parser will correctly parse ``print(something)`` because it is valid
+    code and the second phase will never execute but if you try to parse ``print(something) $ 3`` the first pass
+    of the parser will fail (because of the ``$``) and in the second phase, the rule will match the
+    ``print(something)`` as ``print`` followed by the variable ``something`` between parenthesis and the error
+    will be reported there instead in the ``$`` character.
```
lysnikolaou, line 781:
```suggestion
    will be reported there instead of the ``$`` character.
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+---------------------------
+
+Making experiments
+~~~~~~~~~~~~~~~~~~
+
+As the generated C parser is the one used by Python, this means that if something goes wrong when adding some
+new rules to the grammar you cannot correctly compile and execute Python anymore. This makes it a bit challenging
+to debug when something goes wrong, specially when making experiments.
+
+For this reason, is a good idea to experiment first by generating a Python parser. To do this, you can go to the
```
lysnikolaou, line 837:
```suggestion
For this reason it is a good idea to experiment first by generating a Python parser. To do this, you can go to the
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+When Python is compiled in debug mode (by adding ``--with-pydebug`` when running the configure step in Linux or by
+adding ``-d`` when calling the :file:`PCbuild/python.bat` script in Windows), is possible to activate a **very** verbose
+mode in the generated parser. This is very useful to debug the generated parser and to understand how it works, but it
+can be a bit hard to understand at first. 
+
+.. note::
+
+    When activating verbose mode in the Python parser, is better to not use interactive mode as it can be much harder to
+    understand, as interactive mode involves some special steps compared with regular parsing.
```
lysnikolaou, line 865:
```suggestion
    When activating verbose mode in the Python parser, it is better to not use interactive mode as it can be much harder to
    understand, because interactive mode involves some special steps compared to regular parsing.
```

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+character marks the type of the trace:
+
+* ``>`` indicates that a rule is going to be attempted to be parsed.
+* ``-`` indicates that a rule has failed to be parsed.
+* ``+`` indicates that a rule has been parsed correctly.
+* ``!`` indicates that an exception or an error has ben detected and the parser is unwinding.
+
+The <token_location> part indicates the current index in the token array, the
+<rule_name> part indicates what rule is being parsed and the <alternative> part
+indicates what alternative withing that rule is being attempted.
```
lysnikolaou, line 888:
```suggestion
indicates what alternative within that rule is being attempted.
```

lysnikolaou (commented):
Thanks for this, Pablo! It's truly amazing work! It'll be so helpful to newcomers that may be interested in working on parser-related tasks.

Also, I left some comments. Most of them are wording-related, so feel free to ignore them.

encukou (commented):
Thank you for writing this up!

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+
+a context-free-grammar parser (like an LL(1) parser) will generate constructions
+that given an input string will *deduce* which alternative (``A``, ``B`` or ``C``)
+must be expanded, while a PEG parser will check if the first alternative succeeds
+and only if it fails, will it continue with the second or the third one in the
+order in which they are written. This makes the choice operator not commutative.
+
+Unlike LL(1) parsers, PEG-based parsers cannot be ambiguous: if a string parses,
+it has exactly one valid parse tree. This means that a PEG-based parser cannot
+suffer from the ambiguity problems described in the previous section.
```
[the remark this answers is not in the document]
pablogsal: Yeah, let me fix this sentence

parser.rst
pablogsal: I will remove the LL(1) parser reference

parser.rst
```diff
@@ -0,0 +1,901 @@
...
+    rule_name: '(' ~ some_rule ')' | some_alt
+
+In this example, if a left parenthesis is parsed, then the other
+alternative won’t be considered, even if some_rule or ‘)’ fail to be
+parsed.
+
+Left recursion
+~~~~~~~~~~~~~~
+
+PEG parsers normally do not support left recursion but Pegen implements a
```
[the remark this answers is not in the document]
pablogsal: Hummmm, indeed, I will find out where to define it first

parser.rst
pablogsal: Thanks for the catch! "In the Python" should be removed.

parser.rst
pablogsal: 👍

parser.rst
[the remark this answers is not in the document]
pablogsal: I removed the reference and substitute it for "CPython's parser generator" as we have a section about pegen later

parser.rst
pablogsal: I apologize, but unfortunately, I still don't know what "the black type" means. I tried to search for "black" but I cannot find anything.

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+    <indentation> ('>'|'-'|'+'|'!') <rule_name>[<token_location>]: <alternative> ...
+
+Every line is indented by a different amount (``<indentation>``) depending on how deep the call stack is. The next
+character marks the type of the trace:
+
+* ``>`` indicates that a rule is going to be attempted to be parsed.
+* ``-`` indicates that a rule has failed to be parsed.
+* ``+`` indicates that a rule has been parsed correctly.
+* ``!`` indicates that an exception or an error has ben detected and the parser is unwinding.
```
terryjreedy, line 886: sp
```suggestion
* ``!`` indicates that an exception or an error has been detected and the parser is unwinding.
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+    When activating verbose mode in the Python parser, it is better to not use interactive mode as it can be much harder to
+    understand, because interactive mode involves some special steps compared to regular parsing.
+
+To activate verbose mode you can add the ``-d`` flag when executing Python:
+
+.. code-block:: shell
+
+    $ python -d file_to_test.py
+
+This will print **a lot** of output to `stderr` so is probably better to dump it to a file for further analysis. The output
```
pablogsal, line 875:
```suggestion
This will print **a lot** of output to ``stderr`` so is probably better to dump it to a file for further analysis. The output
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+represents the "code" in said function. Each parsing function conceptually takes
+an input string as its argument, and yields one of the following results:
+
+* A "success" result. This result indicates that the expression can be parsed by
+  that rule and the function may optionally move forward or consume one or more
+  characters of the input string supplied to it.
+* A "failure" result, in which case no input is consumed.
+
+Notice that "failure" results do not imply that the program is incorrect or a
+parsing failure because as the choice operator is ordered, a "failure" result
```
lysnikolaou, line 61: > Notice that "failure" results do not imply that the program is incorrect or a
parsing failure or a parsing failure

This is a bit hard to read at first and needs a second pass to understand correctly, I feel. Maybe something like this would work a bit better?

Notice that "failure" results do not imply that the program is incorrect, nor do they necessarily mean that the parsing has failed. Since the choice operator is ordered, a failure very often merely indicates "try the following option".

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+merely indicates "try the following option". A direct implementation of a PEG
+parser as a recursive descent parser will present exponential time performance
+the worst case, because PEG parsers have infinite lookahead (this means that
```
lysnikolaou, line 64:
```suggestion
in the worst case, because PEG parsers have infinite lookahead (this means that
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+Key ideas
+~~~~~~~~~
+
+.. important::
+    Don't try to reason about a PEG grammar in the same way you would to with an EBNF
+    or context free grammar. PEG is optimized to describe **how** input strings will
+    be parsed, while context-free grammars are optimized to generate strings of the
+    language they describe (in EBNF, to know if given string is in the language you need
+    to do work to find out as is not immediately obvious from the grammar).
```
lysnikolaou, line 82:
```suggestion
    language they describe (in EBNF, to know if a given string is in the language, you need
    to do work to find out as it is not immediately obvious from the grammar).
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+* Alternatives are ordered ( ``A | B`` is not the same as ``B | A`` ).
+* If a rule returns a failure doesn't mean that the parsing has failed,
```
lysnikolaou, line 85:
```suggestion
* If a rule returns a failure, it doesn't mean that the parsing has failed,
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+final ``a`` in the rule consume the second, and leaving out the third ``a``.
+As the rule has succeeded, no attempt is ever made to go back and let
+``'a'|'aa'`` try the second alternative. The expression ``('aa'|'a')'a'`` does
+not accept ``aa`` because ``'aa'|'a'`` accepts all of ``aa``, leaving nothing
+for the final ``a``. Again, the second alternative of ``'aa'|'a'`` is not
+tried.
+
+.. caution::
+
+    The effects of ordered choice such as illustrated above, may be hidden by many levels of rules.
```
lysnikolaou, line 124:
```suggestion
    The effects of ordered choice, such as the ones illustrated above, may be hidden by many levels of rules.
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+For this reason, writing rules where an alternative is contained in the next one is in almost all cases a mistake,
+for example: ::
+
+    my_rule:
+        | 'if' expression 'then' block
+        | 'if' expression 'then' block 'else' block
+
+In this example, the second alternative will never be tried because the first one will
+succeed first (even if the input string has a ``'else' block`` that follows). To correctly
```
lysnikolaou, line 134:
```suggestion
succeed first (even if the input string has an ``'else' block`` that follows). To correctly
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+By selectively turning on memoization for a handful of rules, the parser becomes faster and uses less memory.
+
+.. note::
+    Left-recursive rules always use memoization, since the implementation of left-recursion depends on it.
+
+To know if a new rule needs memoization or not, benchmarking is required
+(comparing execution times and memory usage of some considerably big files with
+and without memoization). There is a very simple instrumentation API available
+in the C generated parse code that allows to measure how much each rule uses
```
lysnikolaou, line 612:
```suggestion
in the generated C parse code that allows to measure how much each rule uses
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+            ^
+    SyntaxError: invalid syntax
+    >>> foo(class=3)
+    File "<stdin>", line 1
+        foo(class=3)
+            ^^^^^
+    SyntaxError: invalid syntax
+
+While soft keywords don't have this limitation if used in a context other when they
+are defined as keywords:
```
lysnikolaou, line 654:
```suggestion
While soft keywords don't have this limitation if used in a context other the one where they
are defined as keywords:
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+.. code-block:: python
+
+    >>> import keyword
+    >>> keyword.softkwlist
+    ['_', 'case', 'match']
+
+.. caution::
+    Soft keywords can be a bit challenging to manage as they can be accepted in
+    places you don't intend to, given how the order alternatives behave in PEG
+    parser (see :ref:`consequences of ordered choice section
```
lysnikolaou, line 687:
```suggestion
    parsers (see :ref:`consequences of ordered choice section
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+.. note::
+    Tokenizer errors are normally reported by raising exceptions but some special
+    tokenizer errors such as unclosed parenthesis will be reported only after the
+    parser finishes without returning anything.
+
+How Syntax errors are reported
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As described previously in the :ref:`how PEG parsers work section
+<how-peg-parsers-work>`, PEG parsers doesn't have a defined concept of where
```
lysnikolaou, line 712:
```suggestion
<how-peg-parsers-work>`, PEG parsers don't have a defined concept of where
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+    the location of generic syntax errors. Use them carefully at boundaries
+    between rules.
+
+As the Python grammar was primordially written as an LL(1) grammar, this heuristic
+has an extremely high success rate, but some PEG features can have small effects,
+such as :ref:`positive lookaheads <peg-positive-lookahead>` and
+:ref:`negative lookaheads <peg-negative-lookahead>`.
+
+To generate more precise syntax errors, custom rules are used. This is a common practice
+also on context free grammars: the parser will try to accept some construct that is known
```
lysnikolaou, line 735:
```suggestion
also in context free grammars: the parser will try to accept some construct that is known
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+to be incorrect just to report a specific syntax error for that construct. In pegen grammars,
+these rules start with ``invalid_`` prefix. This is because trying to match these rules
```
lysnikolaou, line 737:
```suggestion
these rules start with the ``invalid_`` prefix. This is because trying to match these rules
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+   with an ``invalid_`` prefix. By design this attempt **cannot succeed** and is only executed
+   to give to the invalid rules a chance to detect specific situations where custom, more precise,
+   syntax errors can be raised. This also allows to trade a bit of performance for precision reporting
+   errors: given that we know that the input text is invalid, there is no need to be fast because
+   the interpreter is going to stop anyway.
+
+.. important::
+    When defining invalid rules:
+
+    * Make sure all custom invalid rules raise :exc:`SyntaxError` exceptions (or a subclass if it).
```
lysnikolaou, line 756:
```suggestion
    * Make sure all custom invalid rules raise :exc:`SyntaxError` exceptions (or a subclass of it).
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+    Should trigger the syntax error in the ``$`` character. If your rule is not correctly defined this
+    won't happen. For example, if you try to define a rule to match Python 2 style ``print`` statements
+    to make a better error message and you define it as: ::
+
+        invalid_print: "print" expression
+    
+    This will **seem** to work because the parser will correctly parse ``print(something)`` because it is valid
+    code and the second phase will never execute but if you try to parse ``print(something) $ 3`` the first pass
+    of the parser will fail (because of the ``$``) and in the second phase, the rule will match the
+    ``print(something)`` as ``print`` followed by the variable ``something`` between parenthesis and the error
```
lysnikolaou, line 782:
```suggestion
    ``print(something)`` as ``print`` followed by the variable ``something`` between parentheses and the error
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+
+Debugging generated parsers
+---------------------------
+
+Making experiments
+~~~~~~~~~~~~~~~~~~
+
+As the generated C parser is the one used by Python, this means that if something goes wrong when adding some
+new rules to the grammar you cannot correctly compile and execute Python anymore. This makes it a bit challenging
+to debug when something goes wrong, specially when making experiments.
```
lysnikolaou, line 837:
```suggestion
to debug when something goes wrong, especially when making experiments.
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+    understand, because interactive mode involves some special steps compared to regular parsing.
+
+To activate verbose mode you can add the ``-d`` flag when executing Python:
+
+.. code-block:: shell
+
+    $ python -d file_to_test.py
+
+This will print **a lot** of output to ``stderr`` so is probably better to dump it to a file for further analysis. The output
+consist of trace lines with the following structure:
```
lysnikolaou, line 876:
```suggestion
consists of trace lines with the following structure:
```

parser.rst
```diff
@@ -0,0 +1,903 @@
...
+        | second_alt1 second_alt2 { second_alt1 }
+
+If the action is omitted and C code is being generated, then there are two
+different possibilities:
+
+1. If there’s a single name in the alternative, this gets returned.
+2. If not, a dummy name object gets returned (this case should be avoided).
+
+If the action is omitted and Python code is being generated, then a list
+with all the parsed expressions gets returned (this is meant for debugging).
```
lysnikolaou, line 352: I don't think that's true anymore. I think I remember Matthieu (I guess the state of the python generator in the pegen repo is going to be the standard for this document as well as far as the Python generator goes) changing this to most closely resemble what the C generator does: If there's a single name it gets returned, otherwise the list with all parsed expressions gets returned. Do you think we should change it?

lysnikolaou (commented):
A few more nits here and there. Ignore them if you want to merge this ASAP.

Also, I think that it'd be nice if both of the example grammars generated the exact same parser, just in different languages. They're also somewhat outdated, so here's a suggestion for the ones to use, if you like:

C:
```
start[mod_ty]: a=expr_stmt* ENDMARKER { _PyAST_Module(a, NULL, p->arena) }
expr_stmt[stmt_ty]: a=expr NEWLINE { _PyAST_Expr(a, EXTRA) }

expr[expr_ty]:
    | l=expr '+' r=term { _PyAST_BinOp(l, Add, r, EXTRA) }
    | l=expr '-' r=term { _PyAST_BinOp(l, Sub, r, EXTRA) }
    | term

term[expr_ty]:
    | l=term '*' r=factor { _PyAST_BinOp(l, Mult, r, EXTRA) }
    | l=term '/' r=factor { _PyAST_BinOp(l, Div, r, EXTRA) }
    | factor

factor[expr_ty]:
    | '(' e=expr ')' { e }
    | atom

atom[expr_ty]:
    | NAME
    | NUMBER
```

Python:
```
start[ast.Module]: a=expr_stmt* ENDMARKER { ast.Module(body=a or [], type_ignores=[]) }
expr_stmt: a=expr NEWLINE { ast.Expr(value=a, LOCATIONS) }

expr:
    | l=expr '+' r=term { ast.BinOp(left=l, op=ast.Add(), right=r, LOCATIONS) }
    | l=expr '-' r=term { ast.BinOp(left=l, op=ast.Sub(), right=r, LOCATIONS) }
    | term

term:
    | l=term '*' r=factor { ast.BinOp(left=l, op=ast.Mult(), right=r, LOCATIONS) }
    | l=term '/' r=factor { ast.BinOp(left=l, op=ast.Div(), right=r, LOCATIONS) }
    | factor

factor:
    | '(' e=expr ')' { e }
    | atom

atom:
    | NAME
    | NUMBER
```

parser.rst
pablogsal: :+1: Yeah, I agree, this version reads better. Thanks!

pablogsal:
> start[ast.Module]: a=expr_stmt* ENDMARKER { ast.Module(body=a or [], type_ignores=[]) }
> expr_stmt: a=expr NEWLINE { ast.Expr(value=a, LOCATIONS) }
> 
> expr:
>     | l=expr '+' r=term { ast.BinOp(left=l, op=ast.Add(), right=r, LOCATIONS) }
>     | l=expr '-' r=term { ast.BinOp(left=l, op=ast.Sub(), right=r, LOCATIONS) }
>     | term
> 
> term:
>     | l=term '*' r=factor { ast.BinOp(left=l, op=ast.Mult(), right=r, LOCATIONS) }
>     | l=term '/' r=factor { ast.BinOp(left=l, op=ast.Div(), right=r, LOCATIONS) }
>     | factor
> 
> factor:
>     | '(' e=expr ')' { e }
>     | atom
> 
> atom:
>     | NAME
>     | NUMBER

Great point! I will update the text with a version of these grammars. I will tweak them a bit so they are a bit more similar (name of the macro and some other touches).

parser.rst
pablogsal: Yeah, let's change it to the current behaviour

pablogsal: Also, I still want to keep the Python generator in the CPython repo updated with the one in pegen

lysnikolaou (approved):
Really great work, @pablogsal. Thanks!

pablogsal:
Thanks ❤️

Issue #716, opened 2021-06-16
Document ``build.bat`` counterparts for many instructions using ``make``


Fidget-Spinner:
A few sections using ``make`` don't always have their Windows ``build.bat`` counterpart. One example here:
https://devguide.python.org/grammar/#checklist .

The docs should be updated to have the equivalent ``build.bat`` command beside ``make``.

Fidget-Spinner:
To get the ball rolling, I'll work on the example I mentioned in the first message. If anyone finds anything in other sections, reporting them here or fixing them would be a great help :).

encukou:
There's one in CPython at the end of `Tools/scripts/stable_abi.py`, and the corresponding commands at the end of https://devguide.python.org/c-api/ might also be simplified. If you know the proper incantations for Windows, please add them!
But these might need to be added to `build.bat` first. (And if you want to go deeper: I couldn't find instructions for adding things to `build.bat`.)

Fidget-Spinner:
``./python ./Tools/scripts/stable_abi.py --generate-all ./Misc/stable_abi.txt`` doesn't work on MSVC (Windows) because it seems to require GCC/Clang. I will add a note in the docs somewhere for that.

PR #735, opened 2021-08-03
Specify that you need GCC to regenerate Stable ABI

Fidget-Spinner:
Fixes rest of #716.

### DIFF
--- a/c-api.rst
+++ b/c-api.rst
@@ -186,9 +186,9 @@ Adding a new definition to the Limited API
 
      ./python ./Tools/scripts/stable_abi.py --generate-all ./Misc/stable_abi.txt
 
-   .. note::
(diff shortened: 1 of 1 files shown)


encukou:
This is a bug. GCC should only be needed for checking the result, and "unixy" checks should be skipped on platforms without a CC. But I must admit I only tested on Linux.

What error are you getting?

Fidget-Spinner:
Sorry, I'm unable to reproduce the error anymore :(. At the time I recall some exception about no `CC` and environment variables, but I didn't copy it down and now everything works again.

Tested working on MSVC. Closing this PR now. Thanks for your help Petr.

encukou:
No worries!
Before PEP 652, `stable_abi.py` needed `sysconfig.get_config_var("CC")` for most of what it did. That's probably where you got the issue.

PR #750, opened 2021-10-03
Add instructions on running autoreconf with pkg-config

pablogsal:


tiran:
``configure.ac`` also requires ``autoconf-archive``.

I have updated my build container https://quay.io/repository/tiran/cpython_autoconf to include pkg-config.

ned-deily (approved):

PR #753, opened 2021-10-18
Update Python GitHub Org owners

ewdurbin:
@ezio-melotti was added to be able to develop and test tooling for bpo -> GitHub Issues migration.

Proposes adding @ambv as an org owner. As Developer in Residence this adds one other person with committed time who is closer to CPython day-to-day to help manage membership of the python org, since currently only Owners can perform those invites (for triage team additions, etc).


willingc (approved):
Thanks @ewdurbin. Good call on Lukasz addition.

PR #767, opened 2021-11-14
Update reference to the parser from LL(1) to PEG

itamaro:
not sure if this is considered a minor fix, let me know if I should file an issue for this first

### DIFF
--- a/langchanges.rst
+++ b/langchanges.rst
@@ -46,8 +46,7 @@ the change. This process is the Python Enhancement Proposal (PEP) process.
 You will first need a PEP that you will present to python-ideas. You may be a
 little hazy on the technical details as various core developers can help with
 that, but do realize that if you do not present your idea to python-ideas or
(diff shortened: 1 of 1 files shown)


langchanges.rst
```diff
@@ -47,8 +47,8 @@ You will first need a PEP that you will present to python-ideas. You may be a
 little hazy on the technical details as various core developers can help with
 that, but do realize that if you do not present your idea to python-ideas or
 python-list ahead of time you may find out it is technically not possible
-(e.g., Python's parser will not support the grammar change as it is an LL(1)
-parser). Expect extensive comments on the PEP, some of which will be negative.
+(e.g., Python's parser will not support the grammar change as a PEG parser).
```
pablogsal, line 50: I think the whole sentence in parentheses should be removed

itamaro: removed and updated PR

pablogsal:
Thanks for the PR @itamaro !

PR #779, opened 2021-12-23
So long Python 3.6

hugovk:
And thanks for all the f-strings.

Move 3.6 from the "Status of Python branches" table to the "End-of-life branches" table, and fix a typo in Larry's name.

# Preview

* https://cpython-devguide--779.org.readthedocs.build/#status-of-python-branches
* https://cpython-devguide--779.org.readthedocs.build/devcycle/#end-of-life-branches

# Downloads page?

How to remove it from "Active Python Releases" on https://www.python.org/downloads/ ? And is it time to remove 2.7 from there?


setup.rst
```diff
@@ -240,7 +240,7 @@ Windows
 For a quick guide to building you can read `this documentation`_ from Victor
 Stinner.
 
-**Python 3.6** and later can use Microsoft Visual Studio 2017.  You can download
+**Python 3.7** and later can use Microsoft Visual Studio 2017.  You can download
```
zware, line 243: I believe this could be updated to something like the following:

```suggestion
All current versions of Python can be built using Microsoft Visual Studio 2017
or later.  You can download
```

It might be good to have @zooba confirm that, though :)

zware (approved):

ned-deily:
Thanks for the PR, @hugovk, and the review, @zware!  Rest well, 3.6.

PR #783, opened 2022-01-02
Fix`actions/cache` in `ci.yml`

sobolevn:
There were two problems:
1. `actions/cache` was outdated, currently the active version is `@2`: https://github.com/actions/cache/releases
2. For some reason it was using `pyproject.toml` as a caching key, but this file does not exist. I've changed it to be https://github.com/python/devguide/blob/main/requirements.txt

Question: other repos have `dependabot.yml` job to update GitHub Actions. Do you need one? I can send another PR if it is going to be useful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,10 +17,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         with:
           path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-pip-
       - uses: actions/setup-python@v2


hugovk (commented):
Looks good. 👍 

A further step: https://github.com/actions/setup-python now has pip caching built in.

So we could delete the whole `actions/cache` step here, and add to `actions/setup-python`:

```diff
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
```

And it uses `requirements.txt` as the key by default.

Docs: https://github.com/actions/setup-python#caching-packages-dependencies

hugovk:
> Question: other repos have `dependabot.yml` job to update GitHub Actions. Do you need one? I can send another PR if it is going to be useful.

Dependabot is good but unfortunately it's very spammy for forks:

* https://github.com/dependabot/dependabot-core/issues/2804
* https://github.com/dependabot/dependabot-core/issues/2198

sobolevn:
> A further step: https://github.com/actions/setup-python now has pip caching built in.

I didn't know about that! Thanks!
But, this change might require some further considerations, let's keep it as a separate feature for the future.

hugovk (approved):

Mariatta (approved):

PR #785, opened 2022-01-03
Docs for Python 2.x are no longer built daily

hugovk:
Delete "(as well as 2.x)" from:

> The in-development and most recent 3.x **(as well as 2.x)** maintenance
branches are rebuilt once per day.

The footer of https://docs.python.org/2/ and https://docs.python.org/2.7/ say "Last updated on Jun 19, 2020."

### DIFF
--- a/docquality.rst
+++ b/docquality.rst
@@ -32,8 +32,8 @@ and validates that your new markup is correct.
 
 You can view the documentation built from :ref:`in-development <indevbranch>`
 and :ref:`maintenance <maintbranch>` branches at https://docs.python.org/dev/.
-The in-development and most recent 3.x (as well as 2.x) maintenance
-branches are rebuilt once per day.
+The in-development and most recent 3.x maintenance branches are rebuilt once per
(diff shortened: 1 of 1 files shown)


docquality.rst
```diff
@@ -32,8 +32,8 @@ and validates that your new markup is correct.
 
 You can view the documentation built from :ref:`in-development <indevbranch>`
 and :ref:`maintenance <maintbranch>` branches at https://docs.python.org/dev/.
-The in-development and most recent 3.x (as well as 2.x) maintenance
-branches are rebuilt once per day.
+The in-development and most recent 3.x maintenance branches are rebuilt once per
+day.
```
encukou, line 36: Would it hurt readability if the `3.x` was also removed?

```suggestion
The in-development and recent maintenance branches are rebuilt once per day.
```

hugovk: This is better, thanks.

encukou:
Thanks!

PR #793, opened 2022-01-20
Add padlock emoji to denote voters repo is private

hugovk:
GitHub uses a padlock to indicate private repos.

Add a padlock emoji 🔒 next to the private https://github.com/python/voters/ name.

If you don't have access, GitHub shows a 404 when clicking, making it appear like a broken link: https://github.com/python/devguide/issues/681#issuecomment-1016952381


Mariatta (approved):
Makes sense. Thanks.

PR #795, opened 2022-01-21
Add section about comparing changes to the git bootcamp

### LINKED ISSUE #794: Add git diff tricks to git cheat sheet section
After python/cpython#30745 we can do some neat git diff tricks:

    $ git switch <somebranch-with-autogenerated-changes>
    $ git diff main ':(attr:!generated)'
    $ git config diff.generated.binary true
    $ git diff main

It would be nice to document these in the devguide.

erlend-aasland:
Fixes #794

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -253,13 +253,20 @@ Compare to the ``main`` branch::
 
    git diff main
 
-Exclude generated files from diff (note the single quotes)::
+Exclude generated files from diff using an `attr <_gitattributes>`_
+`pathspec <_pathspec>`_ (note the single quotes)::
 
    git diff main ':(attr:!generated)'
 
 Exclude generated files from diff by default::
 
    git config diff.generated.binary true
+
+The `generated` `attribute <_gitattributes>`_ is defined in `.gitattributes`,
+found in the repository root.
+
+.. _pathspec: https://git-scm.com/docs/gitglossary#def_pathspec
+.. _gitattributes: https://git-scm.com/docs/gitattributes
 
 .. _push-changes:
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -264,7 +264,7 @@ Exclude generated files from diff by default::
    git config diff.generated.binary true
 
 The ``generated`` `attribute <https://git-scm.com/docs/gitattributes>`_ is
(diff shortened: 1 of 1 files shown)


gitbootcamp.rst
```diff
@@ -240,6 +240,27 @@ Commit the files:
...
+   git diff main
+
+Exclude generated files from diff (note the single quotes)::
+
+   git diff main ':(attr:!generated)'
+
+Exclude generated files from diff by default::
+
+   git config diff.generated.binary true
+
```
zware, line 263: Should we mention that both of these options are driven by the `generated` attribute in `.gitattributes`?

erlend-aasland: Sounds good. PTAL.

gitbootcamp.rst
```diff
@@ -253,14 +253,21 @@ Compare to the ``main`` branch::
...
+Exclude generated files from diff using an `attr <_gitattributes>`_
+`pathspec <_pathspec>`_ (note the single quotes)::
 
    git diff main ':(attr:!generated)'
 
 Exclude generated files from diff by default::
 
    git config diff.generated.binary true
 
+The `generated` `attribute <_gitattributes>`_ is defined in `.gitattributes`,
```
zware: reST doesn't like single backticks when it's not a role or link:
```suggestion
The ``generated`` `attribute <_gitattributes>`_ is defined in ``.gitattributes``,
```

I also don't think we need to repeat the link to the gitattributes docs:

```suggestion
The ``generated`` attribute is defined in ``.gitattributes``,
```

gitbootcamp.rst
```diff
@@ -240,6 +240,32 @@ Commit the files:
...
+single quotes)::
+
+   git diff main ':(attr:!generated)'
+
+Exclude generated files from diff by default::
+
+   git config diff.generated.binary true
+
+The ``generated`` `attribute <https://git-scm.com/docs/gitattributes>`_ is
+defined in `.gitattributes`, found in the repository root.
```
zware, line 267:
```suggestion
defined in :file:`.gitattributes`, found in the repository root.
```

erlend-aasland: Oh, I didn't know about that role. Thanks.

erlend-aasland:
Thanks for the improvements, Zachary!

PR #802, opened 2022-01-26
Remove deceased expert and line linking to now-obsolete PEPs with special rules

CAM-Gerlach:
As discussed on python/peps#2270 , looks like [Fredrik "effbot" Lundh](https://mail.python.org/archives/list/python-dev@python.org/thread/36Q5QBILL3QIFIA3KHNGFBNJQKXKN7SD/), who has passed away, is still listed here.

Also, the list still mentions and links to [PEP 291](https://www.python.org/dev/peps/pep-0291/) and [PEP 360](https://www.python.org/dev/peps/pep-0360/), stating they contain special rules for some modules. PEP 291 is thoroughly obsolete, being explicitly limited to Python 2, while PEP 360 has not been updated in many years and with any new additions being explicitly prohibited, it presumably never will be, and all the projects currently listed are maintained in the stdlib and no longer active upstream for at least the past decade or more. This was a source of confusion, so given this sentence no longer applies to any developed CPython version, it should presumably be removed.

@gvanrossum @brettcannon

### DIFF
--- a/experts.rst
+++ b/experts.rst
@@ -50,7 +50,7 @@ __future__
 __main__              gvanrossum, ncoghlan
 _dummy_thread         brett.cannon
 _thread
-_testbuffer           
+_testbuffer
 abc
 aifc                  r.david.murray
 argparse              rhettinger*
@@ -141,7 +141,7 @@ itertools             rhettinger*
 json                  bob.ippolito (inactive), ezio.melotti, rhettinger
 keyword
 lib2to3               benjamin.peterson
-libmpdec              
+libmpdec
 linecache
 locale                lemburg
 logging               vinay.sajip
@@ -187,7 +187,7 @@ pydoc
 queue                 rhettinger*
 quopri
 random                rhettinger, mark.dickinson
-re                    effbot (inactive), ezio.melotti, serhiy.storchaka
+re                    ezio.melotti, serhiy.storchaka
 readline              twouters
 reprlib
 resource              twouters
@@ -254,13 +254,13 @@ wave
 weakref               fdrake
 webbrowser
 winreg                stutzbach
-winsound              effbot (inactive)
+winsound
 wsgiref               pje
 xdrlib
 xml.dom
 xml.dom.minidom
 xml.dom.pulldom
-xml.etree             effbot (inactive), eli.bendersky*, scoder
+xml.etree             eli.bendersky*, scoder
 xml.parsers.expat
 xml.sax
 xml.sax.handler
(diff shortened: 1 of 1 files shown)


AA-Turner:
@CAM-Gerlach The note I proposed in https://github.com/python/peps/pull/2276 requires that the experts index notes that Expat is externally maintained. I also removed Fredrik Lundh from the modules lists:

(Some day GH will allow review suggestions on any line!)

A

```diff
diff --git a/experts.rst b/experts.rst
index a245c91e9..a50c5c453 100644
--- a/experts.rst
+++ b/experts.rst
@@ -184,7 +184,7 @@ pydoc
 queue                 rhettinger*
 quopri
 random                rhettinger, mark.dickinson
-re                    effbot (inactive), ezio.melotti, serhiy.storchaka
+re                    ezio.melotti, serhiy.storchaka
 readline              twouters
 reprlib
 resource              twouters
@@ -251,14 +251,14 @@ wave
 weakref               fdrake
 webbrowser
 winreg                stutzbach
-winsound              effbot (inactive)
+winsound
 wsgiref               pje
 xdrlib
 xml.dom
 xml.dom.minidom
 xml.dom.pulldom
-xml.etree             effbot (inactive), eli.bendersky*, scoder
-xml.parsers.expat
+xml.etree             eli.bendersky*, scoder
+xml.parsers.expat     \*
 xml.sax
 xml.sax.handler
 xml.sax.saxutils
@@ -270,6 +270,8 @@ zipimport             twouters*
 zlib                  twouters
 ====================  =============================================
 
+..note:: \* The expat parser is externally maintained by the
+         `libexpat project <https://libexpat.github.io>`__
 
 Tools
 -----
```

CAM-Gerlach:
> I also removed Fredrik Lundh from the modules lists

Oops, I'd committed that change as a separate commit, but had forgotten to actually `push` it. Thanks for the catch.

> @CAM-Gerlach The note I proposed in python/peps#2276 requires that the experts index notes that Expat is externally maintained.

The issue here is the _Python module_ isn't externally maintained, just `libexpat`, which is no different from `zlib` or the other modules backed by external C libraries that don't have this listed, so I don't think this is really necessary here. Also, I've proposed revising the note, as I"m not sure it totally reflects what Guido and I were suggesting, making this less important. Furthermore, as can be seen on the line immediately above, `*` is already used for a different specific meaning as mentioned in the introduction, so we'd have to use a different symbol to reference it—or, if we do do this, it might be better to just add a short note like `` `libexpat`_ is externally-maintained``...but again, if we do this, we should be consistent and mention it for the other similar modules that are wrappers around external C libraries.

AA-Turner:
> we should be consistent and mention it for the other similar modules that are wrappers around external C libraries.

Agree on this. Something about cans of worms springs to mind!

A

CAM-Gerlach:
> Agree on this. Something about cans of worms springs to mind!

Yeah, that might be beneficial, but is getting kinda out of scope for this PR. That would require tracking down all the various modules that depend on external ones and nothing them; in some cases tightly coupled to the Python module, while in others much more loosely so, with potentially different libraries that could be linked against it depending on how downstreams do it. So I'd rather not open that pandora's box here, and leave that for a separate discussion.

brettcannon (approved):

PR #808, opened 2022-01-28
GitHub Actions updates

hugovk:
# Simplify caching

Follow on from https://github.com/python/devguide/pull/783#pullrequestreview-842402521.

https://github.com/actions/setup-python now has pip caching built in so let's remove the whole `actions/cache` step here. It uses `requirements.txt` as the key by default.

Docs: https://github.com/actions/setup-python#caching-packages-dependencies

# Enable build button

Also, add `workflow_dispatch`, which adds a button to the GitHub Actions UI to let you trigger builds. This is something Travis CI has built in, and is fairly new to GHA and needs enabling.

Docs: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

# Add colour

Finally, GitHub Actions isn't a tty so tools such as sphinx-build that autodetect don't show the output in colour. So let's add `--color` to turn it on, especially useful for the link check. And let's show the docs build output too, useful for debugging failures (and now in colour!).

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches: main
+  workflow_dispatch:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


Mariatta (approved):
Thanks! The colored output looks nice.

PR #812, opened 2022-02-16
Become an expert for typing

JelleZijlstra:
And people can assign issues to me.

I guess there's also some config I need to change so I show up in the autocomplete list on BPO, so I'll go look for that next.


zware (approved):
> I guess there's also some config I need to change so I show up in the autocomplete list on BPO, so I'll go look for that next.

You found it ;)

JelleZijlstra:
> You found it 

It just parses this table? Neat.

PR #814, opened 2022-02-18
Document using Github issues as the issue tracker

ambv:
This change should be landed once issues are migrated from BPO to Github issues. Some information here is based on https://github.com/psf/gh-migration/issues/13 and PEP 588.

The bulk is in `tracker.rst` and `triaging.rst`. Since they must make sense in their entirety, I temporarily deleted the old versions of those files and created new ones so that they can be reviewed without confusing diffs.

### DIFF
--- a/gh-labels.rst
+++ b/gh-labels.rst
@@ -1,0 +1,170 @@
+.. _github-labels:
+
+Github Labels
+=============
+
+We're using labels on Github to categorize issues and pull requests.
+Many labels are shared for both use cases, while some are dedicated
+only to one. Below is a possibly inexhaustive list, but it should get
+you going. For a full list, see `here <https://github.com/python/cpython/issues/labels>`_.
+
+General purpose labels
+----------------------
+
+type-behavior
+    Used for issues/PRs that address unintentional behavior, but do not
+    pose significant security concerns. Generally, bugfixes will be attached
+    to a specific issue where the unintended behavior was first reported.
+
+type-documentation
+    Used for issues/PRs that exclusively involve changes to
+    the documentation. Documentation includes `*.rst` files, docstrings,
+    and code comments.
+
+type-enhancement
+    Used for issues/PRs that provide additional functionality
+    or capabilities beyond the existing specifications.
+
+type-performance
+    Used for issues/PRs that provide performance optimizations.
+
+type-security
+    Used for issues/PRs that involve critical security issues. Less severe
+    security concerns can instead use the type-bugfix label.
+
+type-tests
+    Used for issues/PRs that exclusively involve changes to the tests.
+
+OS-Mac / OS-Windows
+    Used for issues/PRs involving changes which only have an effect upon
+    a specific operating system.
+
+spam
+    Used for issues/PRs that don't include enough eggs or bacon.
+
+Labels specific to issues
+-------------------------
+
+Priority
+^^^^^^^^
+
+release-blocker
+    The highest priority of an issue. If unaddressed, will cause the
+    release manager to hold releasing a new version of Python.
+
+deferred-blocker
+    A release blocker that was pushed one or more releases into the
+    future. Possibly a temporary workaround was employed, or the version
+    of Python the issue is affecting is still in alpha or beta stages
+    of development.
+
+Component
+^^^^^^^^^
+
+library
+    Used for issues involving Python modules in the ``Lib/`` dir.
+
+documentation
+    Used for issues involving documentation in the ``Doc/`` dir.
+
+interpreter-core
+    Used for issues in interpreter core (``Objects/``, ``Python/``,
+    ``Grammar/``, and ``Parser/`` dirs).
+
+extension-modules
+    Used for issues involving C modules in the ``Modules/`` dir.
+
+tests
+    Used for issues involving only Python's regression test suite, i.e.
+    files in the ``Lib/test/`` dir.
+
+Other
+^^^^^
+
+new
+    Denotes that the issue hasn't been looked at by triagers or core
+    developers yet.
+
+easy
+    Denotes that the issue is a good candidate for a newcomer to address.
+
+
+Labels specific to PRs
+----------------------
+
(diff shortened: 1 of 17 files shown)


gh-labels.rst
```diff
@@ -0,0 +1,169 @@
...
+    Used for PRs that include changes which are no longer relevant or when the
+    author hasn't responded to feedback in a long period of time. This label
+    helps core developers quickly identify PRs that are candidates for closure
+    or require a ping to the author.
+
+awaiting review
+    Used for PRs that haven't been reviewed by anyone yet.
+
+awaiting core review
+    Used for PRs that have at least one core developer listed in the
```
JelleZijlstra, line 149: Not sure this is right. I see this label being autoapplied either when:
- A non-core dev has reviewed the PR (even if they requested changes)
- Or when the PR is authored by a core dev

gh-labels.rst
```diff
@@ -0,0 +1,169 @@
...
+    file.
+
+awaiting changes
+    A reviewer required changes to proceed with the PR.
+
+awaiting change review
+    The PR author made requested changes, and they are waiting for review.
+
+awaiting merge
+    A triager or core developer issued an approving review. The PR
```
JelleZijlstra, line 162: Triager review doesn't trigger this label

gh-labels.rst
Mariatta: > I see this label being autoapplied either when:

> A non-core dev has reviewed the PR (even if they requested changes)
> Or when the PR is authored by a core dev

I confirm that this is how bedevere automatically applies the `awaiting core review` label.

gh-labels.rst
JelleZijlstra:
```suggestion
    A core developer issued an approving review. The PR
```

index.rst
```diff
@@ -304,8 +304,8 @@ Full Table of Contents
    documenting
    silencewarnings
    fixingissues
-   tracker
-   triaging
+   tracker2
```
Mariatta, line 307: Just curious of the renaming from `tracker` to `tracker2`. Is this because we're keeping a backup of the old `tracker` document? Does `2` means like `version 2` or something?

JelleZijlstra: I think @ambv did that so it's easier to review the new document as a whole, because it's nearly a full rewrite. Unfortunately, GitHub in the review pane shows a diff between tracker and tracker2 anyway.

ambv: Yes, @JelleZijlstra is right. Git itself is trying to be too smart here but fortunately I foresaw this and the deletion is a separate commit. So if you only look at [699f8f1](https://github.com/python/devguide/pull/814/commits/699f8f1d586e88c64c23268c4bfada962342500b), it reads easier.

gh-labels.rst
ambv: I tried to avoid passive voice. All in vain!

tracker2.rst
```diff
@@ -154,67 +170,53 @@ Doing all of this allows core developers and :ref:`triagers <devrole>` to more
...
+in the issue tracker or use the `advanced search`_ query builder to search for
+specific kinds of issues (e.g. the "Windows" label if you are a Windows
+developer, "Extension Modules" if you are familiar with C, etc.). 
 
 
 .. _devrole:
 
-Gaining the "Developer" Role on the Issue Tracker
-=================================================
+Gaining the "Triager" Role on the Issue Tracker
```
Mariatta, line 185: I think we can remove this whole section completely. I believe this was specific to the role on b.p.o.

For becoming a member of the Python Triage team on GitHub, it was already covered in a different page: https://devguide.python.org/triaging/#becoming-a-member-of-the-python-triage-team

ezio-melotti (changes requested):
I left a few comments but haven't fully reviewed the triaging2 page, since it shows up as a whole new file and it's hard to tell what changed from the original page.  It might be better to just edit the original pages instead of creating new files.

I think it would be better to divide the FAQ/docs in 2-3 sections:
- [ ] the migration (what has been transferred, how bpo things map to GitHub, etc.)
- [ ] GitHub issues in general (general guidelines and tips about using GitHub issues)
- [ ] CPython-specific guidelines (labels, templates, workflows, actions, etc.)

Currently the FAQs are a mix of the first two, but it would be better to separate them.  For example, the "Where is the 'nosy list'?" explains how to subscribe to issues.  This is useful for bpo users, but also for new users, except that they wouldn't know what the "nosy list" is and might have a hard time finding out that answer.

With two separate sections you could have "How do I subscribe/follow an issue?" FAQ in the "Using GitHub" section and then a "Where is the 'nosy list'?" FAQ in the "GitHub for BPO users" section that links to it.

Some CPython-specific guidelines are already covered in the triaging and tracker pages instead, so they don't necessarily have to be in the FAQ (even though there could be FAQs linking to the relevant sections, or some sections could be converted into FAQ).

triaging2.rst
```diff
@@ -0,0 +1,362 @@
+.. _triaging:
```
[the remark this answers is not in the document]
merwok: Sadly git does not track renames, but it tries to guess when computing diffs.  See comment above by ambv, there’s a commit with only the new file added that’s easier to read.

gh-labels.rst
```diff
@@ -0,0 +1,169 @@
...
+sprint
+    Used for PRs authored during an in-person sprint, such as
+    at PyCon, EuroPython, or other official Python events. The label is
+    used to prioritize the review of those PRs during the sprint.
+
+stale
+    Used for PRs that include changes which are no longer relevant or when the
+    author hasn't responded to feedback in a long period of time. This label
+    helps core developers quickly identify PRs that are candidates for closure
+    or require a ping to the author.
```
jacobtylerwalls, line 143: I think stale is just simply no activity, not awaiting changes. No activity might include waiting for an initial review (or a wontfix, as the case may be).

Aof8888 (approved):

PR #818, opened 2022-03-01
Added 'www' in front of links to sphinx-doc.org

slateny:
https://sphinx-doc.org/ doesn't seem to resolve, but https://www.sphinx-doc.org/ does


encukou:
Thank you!

For reference, this is reported at the Sphinx side in https://github.com/sphinx-doc/sphinx/issues/6864

terryjreedy:
Still needed, so good to add.

PR #822, opened 2022-03-26
Update to canonical PEP URLs

hugovk:
With the recent implementation of [PEP 676](https://peps.python.org/pep-0676/) the canonical URL for PEPs has changed from, for example:

https://www.python.org/dev/peps/pep-0008/

to:

https://peps.python.org/pep-0008/

Redirects are in place so the old links still work, but let's update to use the new canonical form.

See also:

* [bpo-47126](https://bugs.python.org/issue47126) / https://github.com/python/cpython/pull/32124
* https://github.com/sphinx-doc/sphinx/pull/10267

### DIFF
--- a/c-api.rst
+++ b/c-api.rst
@@ -58,9 +58,7 @@ definition in ``Include/internal/`` (or directly in a ``.c`` file):
 * “Provisional” API, included in a Python release to test real-world usage
   of new API. Such names should be renamed when stabilized; preferably with
   a macro aliasing the old name to the new one.
-  See `"Finalizing the API" in PEP 590`_ for an example.
-
-.. _"Finalizing the API" in PEP 590: https://www.python.org/dev/peps/pep-0590/#finalizing-the-api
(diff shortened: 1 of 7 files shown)


hugovk:
Actually, will update this to use the `:pep:` role, so they'll be automatically handled by the next Sphinx release (https://github.com/sphinx-doc/sphinx/pull/10267).

hugovk:
Ready for review!

hugovk:
> Actually, will update this to use the [`:pep:` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-pep), so they'll be automatically handled by the next Sphinx release ([sphinx-doc/sphinx#10267](https://github.com/sphinx-doc/sphinx/pull/10267)).

[Sphinx 4.5.0](https://www.sphinx-doc.org/en/master/changes.html#release-4-5-0-released-mar-28-2022) is now out and the `:pep:` role uses the new URLs. 🚀

pablogsal (approved):

pablogsal:
Thanks @hugovk !

PR #552, opened 2019-11-16
Add GitHub sernames for core dev team list

brettcannon:
Closes #507

PR #825, opened 2022-04-01
Use GitHub usernames in the expert index.

ezio-melotti:
This PR is a follow-up of #552 and #507.  The former added a list of GitHub usernames to `developers.csv`, but the list wasn't actually used in #507.  This PR updates the `experts.rst` file with the GitHub names.

Some experts didn't have a GitHub username, so I left their bpo username marked with a `^`, while others now have one on bpo, so I added it to `developers.csv`.  Note that I just checked bpo for people in the expert list that didn't have a GitHub name in `developers.csv` -- I didn't try to find their GitHub name through other means and I didn't check other committers that weren't in the expert list.  I also updated the preamble and fixed a TODO.

While updating the names, I also found 
* 3 users that are not committers but they are listed in the expert index (pradyunsg, gregorlingl, Marcus.Smith)
* 10 users that are marked as committers on bpo, but aren't listed in `developers.csv` (serwy, alanmcintyre, David.Edelsohn, alexis, flowdas, gbtami, rougeth, mattwang44, josix, zhsj -- most of these work on translations)


developers.csv
```diff
@@ -81,7 +81,7 @@ Philip Jenvey,pjenvey,2009-05-07,2020-11-26,For Jython compatibility
 Michael Foord,voidspace,2009-04-01,,For IronPython compatibility
 R\. David Murray,bitdancer,2009-03-30,,
 Chris Withers,cjw296,2009-03-08,,
-Tarek Ziadé,,2008-12-21,2017-02-10,For distutils module; did not make GitHub transition
+Tarek Ziadé,tarekziade,2008-12-21,2017-02-10,For distutils module
```
brettcannon, line 84: This file is generated from https://github.com/python/voters/ , so please update there and then update the file via the GitHub Action which creates the CSV file.

brettcannon (changes requested):

terryjreedy:
I would remove following inactive people (maybe another issue).  (I believe I added the 'inactive' labels to these.)
tkinter: G. Polo was a GSOC student about 2009?, Not active since. 
turtle: Gregor Lingl contributed the current turtle.py in about 2009, but has been inactive since, or soon after commit.
Never signed CLA.  Said he would about 2015 after 2 emails but did not.
idlelib: Roger Serwy got commit rights in 2013 and merged about 10 IDLE issues soon after.  I only remember 1 or 2 comments since. 
idlelib: Kurt Kaiser stopped sometime before 2010, and I do not remember seeing anything other than one message since.

developers.csv
ezio-melotti: I created https://github.com/python/voters/pull/64.

ezio-melotti:
I could remove the ones without a GitHub username in this PR, so I don't have to mark them with a `^` (which might also break the nosy autocomplete of bpo).  While I'm at it I could also remove from the expert index the ones you mentioned, or even all the ones marked as `(inactive)`.  @brettcannon, what do you think?

brettcannon:
I don't really have an opinion as I don't use this file.

ezio-melotti:
I used the generated `developers.csv` to replace the one I had created, and the only difference seems to be with some whitespace.  We can merge this after the GitHub migration, since in its current form it might break the nosy autocomplete on bpo.

PR #826, opened 2022-04-06
bpo-36329: Replace 'make serve' with 'make htmlview'

hugovk:
Devguide version of https://github.com/python/cpython/pull/32354.

I've tested the `Makefile` on macOS, but haven't tested `make.bat` on Windows, so it would be good if someone can test Windows.

Fixes https://github.com/python/cpython/issues/80510.


make.bat
```diff
@@ -72,6 +70,17 @@ if "%1" == "html" (
 	goto end
 )
 
+if "%1" == "htmlview" (
```
ezio-melotti, line 73: This should be documented in the help above.

hugovk: Done!

encukou:
Could you make `serve` echo `The serve target was removed, use htmlview instead (see https://github.com/python/cpython/issues/80510)`, to help anyone still using it?

hugovk:
Updated!

Makefile
```diff
@@ -139,8 +138,12 @@ doctest: venv
...
+htmlview: html
+	$(PYTHON) -c "import os, webbrowser; webbrowser.open('file://' + os.path.realpath('_build/html/index.html'))"
+
 check:
 	$(PYTHON) tools/rstlint.py -i tools -i venv
 
-serve: html
-	tools/serve.py _build/html
+serve:
+	@echo "The serve target was removed, use htmlview instead" \
```
ezio-melotti, line 148: Does this need an extra space between `instead` and `"`?
`linkcheck` and `doctest` have it, but `latex`, `qthelp`, and `htmlhelp` don't.

Maybe add quotes around `'serve'` and  `'htmlview'`.

hugovk: Quotes added.

It doesn't need an extra space.

Without:

```
The 'serve' target was removed, use 'htmlview' instead (see https://github.com/python/cpython/issues/80510)
```

With:

```
The 'serve' target was removed, use 'htmlview' instead  (see https://github.com/python/cpython/issues/80510)
```

ezio-melotti (approved):

PR #832, opened 2022-04-14
Remove whitespace from .rst files, as per make check

slateny:
`sed -z -i 's/ \n/\n/g' *.rst`

PR #833, opened 2022-04-14
Run 'make check' on the CI

hugovk:
PR https://github.com/python/devguide/pull/832 fixes some of the errors found by `make check`. ✅ 

However, the root problem: `make check` isn't run on the CI, so these things slip in.

And interestingly, it must have once run on the CI, because the [devguide says](https://devguide.python.org/docquality/#developer-s-guide-workflow):

> Note that make check runs automatically when you submit a [pull request](https://devguide.python.org/pullrequest/).

This PR:

* Includes PR https://github.com/python/devguide/pull/832 to fix `make check` whitespace errors
* Fixes the other `make check` errors: all cases of using single backticks instead of two for RST
* Run `make check` on the CI again, to prevent these
* Instead of repeating the other long commands on the CI, use the `make` commands (`make html`, `linkcheck`). A good side effect: this also tests the `Makefile`
* Fix "Github" -> "GitHub" typos

Plus some CI config updates:

* Simplify config: we're testing with a single Python version, not a matrix
* Remove the `push: main` restriction. This allows contributors to test their branches _before_ making a PR
* Bump to v3 of actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: Tests
 
 on: [pull_request, push, workflow_dispatch]
 
+env:
+  FORCE_COLOR: 1
+
 jobs:
   test:
     name: test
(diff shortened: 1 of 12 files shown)


pablogsal (approved):

pablogsal:
Thanks a lot for the PR 🎉

PR #841, opened 2022-04-17
Do not recommend to gh- prefix in commit message

methane:
If we add `gh-` prefix to each commits, many spammy notification send to the issue.
See this ecample:

image

We should recommend `gh-` prefix only in pull request title (and merge commit title).

### DIFF
--- a/pullrequest.rst
+++ b/pullrequest.rst
@@ -268,7 +268,7 @@ and for each pull request there may be several commits.  In particular:
 
 Commit messages should follow the following structure::
 
-   gh-42: Make the spam module more spammy (GH-NNNN)
+   Make the spam module more spammy
 
    The spam module sporadically came up short on spam. This change
    raises the amount of spam in the module by making it more spammy.
@@ -346,7 +346,7 @@ Now you want to
 <https://help.github.com/articles/creating-a-pull-request-from-a-fork/>`_.
 If this is a pull request in response to a pre-existing issue on the
 `issue tracker`_, please make sure to reference the issue number using
-``gh-NNNN`` in the pull request title or message.
+``gh-NNNNN: `` in the pull request title.
 
 If this is a pull request for an unreported issue (assuming you already
 performed a search on the issue tracker for a pre-existing issue), create a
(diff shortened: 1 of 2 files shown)


pullrequest.rst
```diff
@@ -346,7 +346,7 @@ Now you want to
 <https://help.github.com/articles/creating-a-pull-request-from-a-fork/>`_.
 If this is a pull request in response to a pre-existing issue on the
 `issue tracker`_, please make sure to reference the issue number using
-``gh-NNNN`` in the pull request title or message.
+``gh-NNNNN: `` in the pull request title.
```
ezio-melotti, line 349: Currently GitHub ignores the `gh-NNNNN: ` in the title, so a `gh-NNNNN` in the message helps adding an event in the issue timeline.  We should recommend both (at least until GitHub fixes title mentions:
```suggestion
``gh-NNNNN: `` in the pull request title and `gh-NNNNN` in the issue body.
```

ezio-melotti:
Since PRs are squash-merged, I agree that we don't need to include the issue number in every commit.  Before the merge the commits will be already contextualized within the branch/PR (so the issue number is known).  After the squash-merge the commit message will link back to the PR, and from the PR it's easy to find the issue.  

Should we recommend to add the `gh-NNNNN: ` to the squash-merge commit message even though the PR is already linked?

pullrequest.rst
methane: Since we have `gh-NNNNN` in title, how about just `#NNNNN` in body?

GitHub doesn't contain pull request body in squash commit message by default.
So `#NNNNN` won't be contained in the commit message.

If core developer copy&paste the pull request message to squash commit message, they would just trim the `#NNNNN`. They don't need to rewrite it to `gh-NNNNN` form because it is included in the first line of the commit message already.

ezio-melotti: > Since we have gh-NNNNN in title, how about just #NNNNN in body?

That works too.  The goal of adding it in the body is just to reference the issue and create an event there, and both `gh-NNNNN` and `#NNNNN` work.

pullrequest.rst
```diff
@@ -346,7 +346,7 @@ Now you want to
 <https://help.github.com/articles/creating-a-pull-request-from-a-fork/>`_.
 If this is a pull request in response to a pre-existing issue on the
 `issue tracker`_, please make sure to reference the issue number using
-``gh-NNNN`` in the pull request title or message.
+``gh-NNNNN: `` in the pull request title and ``#NNNNN`` in the description.
```
arhadthedev, line 349: It seems that a space inside a code block escapes the following double-backtick:

image

methane: Thank you. I fixed it in #844.

PR #855, opened 2022-05-06
Document auto-closing keywords for PRs.

ezio-melotti:
Fixes now outdated #502 and supersedes #701.


ezio-melotti:
See https://cpython-devguide--855.org.readthedocs.build/gitbootcamp/#creating-a-pull-request

gitbootcamp.rst
```diff
@@ -295,6 +295,15 @@ Creating a Pull Request
 
 6. Press the ``Create pull request`` button.
 
+You should include the issue number (``GH-NNNN``) in the title and in the
+body of the PR.
```
hugovk, line 299: It would be good to get Bedevere updated to automatically add it to the PR body:

https://github.com/python/bedevere/pull/430

hugovk (approved):

gitbootcamp.rst
methane: Bot can just link PR to issue. No need to copy `#NNNNN` to PR body to GitHub link them.

ezio-melotti: The `GH-NNNN` in the PR body is needed to add a reference to the issue, whereas the `GH-NNNN` in the title is used by bedevere.

Note that the linked issues in the sidebar only list PRs that will close the issue (or the issues that will be closed by the PR).  There seems to be no mechanism to simply link PRs and issues other than mentioning the number and having a mention reference added by GitHub among the messages.

gitbootcamp.rst
```diff
@@ -295,6 +295,15 @@ Creating a Pull Request
...
+
+You might also use `special keywords
+<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>`_
+to automatically close the linked issue once the PR is merged, however you
+should avoid them if other PRs need to be merged before the issue can be
+closed (e.g. backports to other branches).
```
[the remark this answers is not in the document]
ezio-melotti: > The devguide is most useful for new contributors

FWIW I regularly check the devguide, and given the number of tools involved in our workflow I wouldn't be surprised if other core devs do to.

>  I think new contributors should generally be steered away from doing so. [adding the keywords]

I agree, and I avoid them myself.  Not sure if other core devs find them useful and use them.  If they don't, I wouldn't mind removing the whole paragraph.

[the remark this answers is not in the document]
AlexWaygood: > FWIW I regularly check the devguide, and given the number of tools involved in our workflow I wouldn't be surprised if other core devs do to.

Oh, so do I! But not as often as I did when I first started contributing :)

gitbootcamp.rst
CAM-Gerlach: > The default squash commit message is copied from the first commit in the PR.
PR title and body are not used in the default commit message at all.

Isn't this only the case for PRs that are submitted with only one commit, and without the new "Use PR title for commit message" option checked? At least, I _thought_ the squashed commit message was getting set correctly from the PR title on PRs with multiple commits, and it was just single-commit PRs that was the problem?

ezio-melotti: Bedevere now adds the issue number -- I tweaked this sentence to only mention the title.

gitbootcamp.rst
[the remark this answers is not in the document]
ezio-melotti: If we don't want to advise the use of these keyword, maybe there's no need to document them, unless we can suggest some situations where it's fine to use them.

getting-started/git-boot-camp.rst
```diff
@@ -283,6 +283,14 @@ Creating a Pull Request
 
 6. Press the ``Create pull request`` button.
 
+You should include the issue number in the title of the PR (``gh-NNNNN: ...``).
```
CAM-Gerlach, line 286:
```suggestion
You should include the issue number in the title of the PR,
in the format ``gh-NNNNN: <PR Title>``.
```

Be more specific regarding the format, since at least on the most of the rest of GitHub itself, the conventional way to refer to GitHub issues numbers is using `#` rather than `gh-`.

getting-started/git-boot-camp.rst
```diff
@@ -283,6 +283,14 @@ Creating a Pull Request
...
+
+You might also use `special keywords
+<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>`_
+to automatically close the linked issue once the PR is merged, however you
```
CAM-Gerlach, line 290:
```suggestion
to automatically close the linked issue once the PR is merged. However, you
```

Fix comma splice, add missing comma and this makes more sense as two separate sentences.

CAM-Gerlach (commented):
A couple textual suggestions

ezio-melotti:
I applied the suggested changes, and we now have 3 options:
1. merge as is
2. clarify when they should/shouldn't be used
3. avoid documenting them

CAM-Gerlach (approved):
Sorry for the long delay—missed this somehow.

The text is LGTM, for what it's worth. It might be helpful to have additional clarity on when they should/shouldn't be used, but that's above my pay grade, as is the overall decision on this PR, of course.

getting-started/git-boot-camp.rst
```diff
@@ -283,6 +283,15 @@ Creating a Pull Request
...
 6. Press the ``Create pull request`` button.
 
+You should include the issue number in the title of the PR,
+in the format ``gh-NNNNN: <PR Title>``.
+
+You might also use `special keywords
+<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>`_
+to automatically close the linked issue once the PR is merged. However, you
+should avoid them if other PRs need to be merged before the issue can be
+closed (e.g. backports to other branches).
```
AlexWaygood, line 293: For a new contributor, this might raise more questions than it answers. Can we link to a reference somewhere else in the devguide that tells contributors which PRs are generally backported, and which are not?

CAM-Gerlach: To note, there's also other benefits to linking a PR beyond just auto-close—most notably, it adds an easy to find reference to the PR from the issue and the issue from the PR a number of handy places in the GitHub UI, which make navigation a fair bit quicker than manually/semi-automatically trying to figure out the corresponding PR (or issue).

terryjreedy: Nearly all bug fixes, doc fixes, and IDLE changes are backported, and nearly all feature changes are not.  But even feature issues may require multiple PRs, so the above advice is still incomplete.  And there are PRs without an issue.  By linking autoclose to linking of issue to PR, GH has made issue-PR linking worse than useless.

CAM-Gerlach: Yeah, unfortunately the story on GitHub around relationships of multiple PRs for one issue has never been good, and unlike a number of other pain points that have improved substantially in recent years, this one doesn't seem to have. And unlike UI issues, there's not as much Refined GitHub and tools like it can do about it—it is a big reason that GitHub overlays like ZenHub gained such a following

ezio-melotti: > unfortunately the story on GitHub around relationships of multiple PRs for one issue has never been good

I created this:
* https://github.com/python/bedevere/issues/501

ezio-melotti:
I added a separate entry to the bootcamp about linking, including a rephrased version of the paragraph included in the previous commits.

getting-started/git-boot-camp.rst
```diff
@@ -283,6 +283,28 @@ Creating a Pull Request
...
+will be expanded to show the status and title of the issue/PR.
+
+When you create a PR that includes ``gh-NNNNN`` in the title, `bedevere`_
+will automatically add a link to the issue in the first message.
+
+In addition, pull requests support `special keywords`_ that can be used to
+link to an issue and automatically close it when the PR is merged.
+However, issues often require multiple PRs before they can be closed (e.g.
+backports to other branches), so this features is only useful when
+a single PR is enough to address and close the issue.
```
AlexWaygood, line 303:
```suggestion
backports to other branches), so this features is only useful if
you know for sure that a single PR is enough to address and close the issue.
```

getting-started/git-boot-camp.rst
```diff
@@ -286,11 +286,24 @@ Creating a Pull Request
...
-You might also use `special keywords
-<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>`_
-to automatically close the linked issue once the PR is merged. However, you
-should avoid them if other PRs need to be merged before the issue can be
-closed (e.g. backports to other branches).
+Linking to Issues and Pull Requests
+-----------------------------------
+
+You can link to issues and pull requests using ``gh-NNNNN`` (this form is
+preferred over ``#NNNNN``).  If the reference appears in a list, the link
```
CAM-Gerlach, line 293: For commits and PR titles sure, but why in the comment body? `#NNNN` is the de-facto standard form on GitHub, and is naturally extended to issues on other repos and orgs than just the current one with `org/repo#NNN`, and references to issues/PRs from elsewhere, whereas I know of no such equivalent for `gh-NNN`.

getting-started/git-boot-camp.rst
```diff
@@ -286,11 +286,24 @@ Creating a Pull Request
...
+will be expanded to show the status and title of the issue/PR.
```
CAM-Gerlach, line 294: Is that really what triggers this? I've gotten inconsistent results in my resent testing, with links sometimes expanding and sometimes not in and outside of lists, even in the same spot. I'd be very curious to know.

Also, `gh` (and `#`?) references in commit messages get expanded to `pythongh`—how does that happen?

getting-started/git-boot-camp.rst
arhadthedev: > but why in the comment body

Probably for another potential migration in the future.

Older BPO issues also used hashed issue ids and they already cannot be automatically distinguished from later introduced GitHub PRs. So there's no need to continue such a confusion.

CAM-Gerlach: Thanks, that makes sense I guess. Unfortunately, even as a triager I can't mark my own review comments as resolved (unless/until python/core-workflow#460 is addressed), yet I _can_ mark others' comments as resolved as a PR author (which doesn't make any sense; if anything it should be the opposite), so feel free to resolve this one.

getting-started/git-boot-camp.rst
ezio-melotti: > Oh, is that really what triggers this?

I think so, and it took me a while to figure that out too.  ISTM that references like `#855` are consistently expanded in lists like:
* #855 
* with some text #855
1. #855
2. with some text #855
- [ ] #855
- [ ] with some text #855

But not in paragraphs like this (#855) or quotes like

> #855

Do you have any counterexamples?

> Also, gh (and #?) references in commit messages get expanded to pythongh—how does that happen?

I've seen that and I think it's just a bug in GitHub.

CAM-Gerlach: > with some text https://github.com/python/devguide/pull/855

Ah, I misinterpreted what you said originally to mean just bullets with no other text, when I was seeing them both with and without—silly me.

> Do you have any counterexamples?

Not any others that I can reliably reproduce—I believe the behavior I saw (on the What's New issues) when I was trying to figure this out might have perhaps just been a combination of an issue where the link wasn't getting created as above when the linked issue hadn't been created when I loaded the page, and/or just not being totally sure what caused it when testing a number of things (particularly `#NNN` vs `org/repo#NNN` vs `<full URL>` vs. <comment URL>).

gitbootcamp.rst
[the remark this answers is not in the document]
ezio-melotti: I tried to address this in the last version, and your suggestion makes it even more explicit, so I'm going to mark this as resolved.

AlexWaygood (approved):

gitbootcamp.rst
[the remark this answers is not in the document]
AlexWaygood: Yeah, it looks good to me now :)

getting-started/git-boot-camp.rst
CAM-Gerlach: One side note that came up during some related testing—AFAIK, `GH-`/`gh-` doesn't trigger the issue number lookup and autocompletion, which I find invaluable especially on repos nearing 6 digit issue and PR numbers like CPython.

getting-started/git-boot-camp.rst
omerbensaadon: hey folks, I've captured this feedback! keep it coming :)

CAM-Gerlach: Thanks @omerbensaadon ! We all really appreciate it!

getting-started/git-boot-camp.rst
ezio-melotti: In https://github.com/python/devguide/pull/946#pullrequestreview-1096790696 the ref is not expanded, maybe because it's in a review?

Screenshot showing both the unexpanded list, and the popup on the first ref which confirms the ref is correct:
image

Issue #854, opened 2022-05-06
Update `Makefile`


ezio-melotti:
- [x] `make clean` should also remove the `venv` (see #856)
- [x] `make clean-venv` should be added to only remove the `venv` (see #856)
- [x] `make check` should check for the default role (see #866)
- [x] `make htmlview` should be added (see #826)
- [x] `make venv` shouldn't print when called by other targets (see #867)

ezio-melotti:
This is all done -- closing.

PR #856, opened 2022-05-06
Update `venv` and `clean` make targets

ezio-melotti:
This PR:
* Updates the `venv` target
* Adds a `clean-venv` target
* Updates the `clean` target to call `clean-venv` too

These are copied from [`cpython/Doc/Makefile`](https://github.com/hugovk/cpython/blob/1ed8d035f1edfaec34016b9f8d615df9e9fe9414/Doc/Makefile#L150-L159) with minor changes.

See #854.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,21 +3,23 @@
 
 # You can set these variables from the command line.
 PYTHON        = python3
-SPHINXOPTS    =
-SPHINXBUILD   = ./venv/bin/sphinx-build
-PAPER         =
+VENVDIR       = ./venv
 BUILDDIR      = _build
+SPHINXOPTS    =
+SPHINXBUILD   = $(VENVDIR)/bin/sphinx-build
+PAPER         =
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,13 +40,22 @@ help:
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  check      to run a check for frequent markup errors"
 
-clean:
+clean: clean-venv
 	-rm -rf $(BUILDDIR)/*
 
+clean-venv:
```
hugovk, line 46: Should we include `clean-venv` in the help?

hugovk (approved):

hugovk:
Good to have things working in a similar way here and for CPython docs 👍

Makefile
ezio-melotti: It isn't included in CPython, and I think that's because:
1. if you try to recreate the venv, it already tells you to use `make clean-venv`
2. if you run `make clean` it also cleans the venv

Because of these reasons and for consistency with CPython, I decided to leave it out.

hugovk: Yep, sounds good.

PR #887, opened 2022-06-08
Update GitHub Organization Owners

ewdurbin:
bugs.python.org => GitHub migration is complete, @ezio-melotti no longer needs Owner bit.


ezio-melotti:
FWIW I found that bit handy for a number of administrative tasks, including managing teams, reclaiming mannequins, reporting spam (I blocked a user that was spamming several of our repos just today), setting autolinks and webhooks, etc.

Some of these require org-level permission, while others only require repo-level ones.  Would it be possible to either request the permission again, either for the org or for specific repos (I mostly used on the devguide, core-workflow, bedevere, peps, blurb/blurb_it, cherry_picker repos).  Should I bring this up to with the SC or is someone else handling this?

PR #891, opened 2022-06-09
Updating brew install command to avoid pkg-config warning

izumiberat:
I am using Mac OS Catalina 10.15.7.
To avoid getting the following warning : 
`configure: WARNING: pkg-config is missing. Some dependencies may not be detected correctly.`
`brew install pkg-config` needs to be run.

Issue number: https://github.com/python/cpython/issues/93658

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -385,7 +385,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install pkg-config openssl@1.1 xz gdbm
+    $ brew install pkg-config openssl xz gdbm
 
 and ``configure`` Python versions >= 3.7::
 
--- a/setup.rst
+++ b/setup.rst
@@ -385,7 +385,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz gdbm
+    $ brew install pkg-config openssl xz gdbm
 
 and ``configure`` Python versions >= 3.7::
 
--- a/setup.rst
+++ b/setup.rst
@@ -385,7 +385,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install pkg-config openssl xz gdbm
+    $ brew install pkg-config openssl@1.1 xz gdbm
 
 and ``configure`` Python versions >= 3.7::
 


erlend-aasland (approved):
looks good to me

setup.rst
```diff
@@ -385,7 +385,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz gdbm
+    $ brew install pkg-config openssl@1.1 xz gdbm
```
ezio-melotti, line 388: Which version of `openssl` gets installed if you don't specify the version number?
The error you mentioned in the issue says `Python requires a OpenSSL 1.1.1 or newer` and it seems odd to me that it would install an older version by default and require a pinned version to get a newer release.  I wonder if that error is wrong :thinking:

izumiberat: @ezio-melotti that is a very good point. I uninstalled openssl `brew uninstall openssl`, and verified that I would get the same message when compiling:
> The necessary bits to build these optional modules were not found:
> _hashlib              _ssl   
I then ran `brew install openssl`. This time it worked. The message above didn’t show. I am sure though (I can still see it on the terminal, that installing openssl the first time, and following the instructions, lead to the message above). I believe that even though I uninstalled openssl, the uninstallation didn't remove all the files that were created when installing openssl1.1. I am therefore making this pull request purely about the warning issue and not the ssl module. I updated the related issue accordingly. If you think it is not out of the scope of the documentation, I will create a new issue and pull request, updating the documentation by adding a note saying that user might need to specify the latest version of openssl to be able to build the ssl module successfully. Let me know what you think.

ezio-melotti (approved):

setup.rst
ezio-melotti: Did you check which version of `openssl` gets installed by default (you can use `openssl version` to check)?  FWIW on Linux I have `OpenSSL 3.0.2 15 Mar 2022` (`1.1.1` is from 2018-09-11 and `3.0.0` from 2021-09-07).
You could also try to run `make clean` or `make distclean` and then run `./configure` and `make` again to start fresh and see if anything changes.

ezio-melotti: https://github.com/python/cpython/issues/91357#issuecomment-1105835425 might also be relevant -- maybe you were missing the `libssl-dev` package before? (cc @tiran)

izumiberat: It might be related but I have no quick way to check for sure now that I can't reproduce my issue in the first place! `libssl-dev` appears nowhere in my terminal. If I do a search on my laptop, I can find `libssl.1.1.dylib` and `libssl.3.dylib` but they were created a month ago, so not related. I am using `brew` and not `pip` so the part of the documentation that needs updating would be different in any case.

Once this PR is committed I will create a new issue and PR as I suggested. And yes, I did run `make clean` and `./configure` before `make` for each of my attempts. Thanks for suggesting it.

izumiberat: @ezio-melotti regarding the `openssl version`, it returns `LibreSSL 2.8.3`. 
If I run `brew info openssl `, I get:
openssl@3: stable 3.0.3 (bottled) [keg-only]

> Cryptography and SSL/TLS Toolkit
> https://openssl.org/
> /usr/local/Cellar/openssl@3/3.0.3 (6,441 files, 28.3MB)
>   Poured from bottle on 2022-06-10 at 10:37:27
> From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/openssl@3.rb
> License: Apache-2.0
> ==> Dependencies
> Required: ca-certificates ✔
> ==> Caveats
> A CA file has been bootstrapped using certificates from the system
> keychain. To add additional certificates, place .pem files in
>   /usr/local/etc/openssl@3/certs
> 
> and run
>   /usr/local/opt/openssl@3/bin/c_rehash
> 
> openssl@3 is keg-only, which means it was not symlinked into /usr/local,
> because macOS provides LibreSSL.
> 
> If you need to have openssl@3 first in your PATH, run:
>   echo 'export PATH="/usr/local/opt/openssl@3/bin:$PATH"' >> ~/.zshrc
> 
> For compilers to find openssl@3 you may need to set:
>   export LDFLAGS="-L/usr/local/opt/openssl@3/lib"
>   export CPPFLAGS="-I/usr/local/opt/openssl@3/include"
> 
> For pkg-config to find openssl@3 you may need to set:
>   export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig"
> 
> ==> Analytics
> install: 137,635 (30 days), 449,281 (90 days), 995,856 (365 days)
> install-on-request: 113,094 (30 days), 363,685 (90 days), 791,601 (365 days)
> build-error: 4,447 (30 days)

ezio-melotti (approved):

ezio-melotti:
I'm going to merge this since it fixes the `pkg-config` warning (see also related #750).  We can continue the discussion about `openssl` in a different issue/PR.

PR #893, opened 2022-06-10
Updating brew install command to build _tkinter

izumiberat:
I am using Mac OS Catalina 10.15.7.
This is to avoid getting the following message:
> The necessary bits to build these optional modules were not found:
> _tkinter

The solution was provided by @erlend-aasland.
Issue number: https://github.com/python/cpython/issues/93659

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -387,12 +387,12 @@ with **Homebrew**::
 
     $ brew install pkg-config openssl xz gdbm tcl-tk
 
-For Python 3.11 and newer::
+For Python 3.10 and newer::
 
     $ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
-For Python versions 3.10 through 3.7::
+For Python versions 3.9 through 3.7::
 
     $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
     $ ./configure --with-pydebug \
--- a/setup.rst
+++ b/setup.rst
@@ -389,7 +389,7 @@ with **Homebrew**::
 
 and ``configure`` Python versions >= 3.7::
 
-    PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" 
+    PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
     ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
 or ``configure`` Python versions < 3.7::
--- a/setup.rst
+++ b/setup.rst
@@ -387,16 +387,27 @@ with **Homebrew**::
 
     $ brew install pkg-config openssl xz gdbm tcl-tk
 
-and ``configure`` Python versions >= 3.7::
+For Python 3.11 and newer::
 
-    PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
-    ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
+    $ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
+      ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
-or ``configure`` Python versions < 3.7::
+For Python versions 3.10 through 3.7::
 
+    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
+    $ ./configure --with-pydebug \
+                  --with-openssl=$(brew --prefix openssl) \
+                  --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
+                  --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
+
+For Python versions 3.6 and older::
+
+    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
     $ CPPFLAGS="-I$(brew --prefix openssl)/include" \
       LDFLAGS="-L$(brew --prefix openssl)/lib" \
-      ./configure --with-pydebug
+      ./configure --with-pydebug \
+                  --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
+                  --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
 
 and ``make``::
 
--- a/setup.rst
+++ b/setup.rst
@@ -385,7 +385,7 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz gdbm tcl-tk
+    $ brew install pkg-config openssl xz gdbm tcl-tk
 
 and ``configure`` Python versions >= 3.7::
 
--- a/setup.rst
+++ b/setup.rst
@@ -385,10 +385,11 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz gdbm
+    $ brew install openssl xz gdbm tcl-tk
 
 and ``configure`` Python versions >= 3.7::
 
+    PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" 
     ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
 or ``configure`` Python versions < 3.7::
--- a/setup.rst
+++ b/setup.rst
@@ -400,15 +400,6 @@ For Python versions 3.9 through 3.7::
                   --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
                   --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
 
-For Python versions 3.6 and older::
-
-    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
-    $ CPPFLAGS="-I$(brew --prefix openssl)/include" \
-      LDFLAGS="-L$(brew --prefix openssl)/lib" \
-      ./configure --with-pydebug \
-                  --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
-                  --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
-
 and ``make``::
 
     $ make -s -j2


setup.rst
```diff
@@ -385,10 +385,11 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install openssl xz gdbm
+    $ brew install openssl xz gdbm tcl-tk
 
 and ``configure`` Python versions >= 3.7::
 
+    PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
```
ezio-melotti, line 392: There's a trailing whitespace here.

setup.rst
```diff
@@ -385,10 +385,11 @@ for the header and library files to your ``configure`` command.  For example,
...
```
erlend-aasland, line 392: `configure` as of **Python 3.11** uses pkg-config. For pre 3.11 `configure`, you'll need to use something like this, IIRC:

```console
$ ./configure --with-tcltk-includes=$(brew --prefix tcl-tk)/include --with-tcltk-libs=$(brew --prefix tcl-tk)/lib
```

Please test the above before adding it.

See also discussion on [Discourse as of March 2022](https://discuss.python.org/t/simplifying-tcl-tk-detection/14066?u=erlendaasland)

izumiberat: @erlend-aasland thanks very much.
Two things: I am unclear about what to test and how to test it. 

----
This is what I understood:
- For Python versions >= 3.11

I keep it as it is originally suggested in this PR:
```` 
PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" 
./configure --with-pydebug --with-openssl=$(brew --prefix openssl) 
````
- For 3.11 > Python versions >= 3.7:

The line you suggested in the above comment (to be tested):
````
$ ./configure --with-tcltk-includes=$(brew --prefix tcl-tk)/include --with-tcltk-libs=$(brew --prefix tcl-tk)/lib
````
followed by the original options provided in the documentation:
````
--with-pydebug --with-openssl=$(brew --prefix openssl)
````

- For Python versions < 3.7

Documentation currently says:

````
$ CPPFLAGS="-I$(brew --prefix openssl)/include" \
LDFLAGS="-L$(brew --prefix openssl)/lib" \
./configure --with-pydebug
````
It would need to be updated as follow:
````
$ CPPFLAGS="-I$(brew --prefix openssl)/include -I$(brew --prefix tcl-tk)/include" \
LDFLAGS="-L$(brew --prefix openssl)/lib -L$(brew --prefix  tcl-tk)/lib" \
./configure --with-pydebug
````
The above is my understanding, let me know if this is correct.

----

> Please test the above before adding it.

If I test using 3.10.4, would that be enough for the use case `3.11 > Python versions >= 3.7`? I am currently on 3.10.4. 
Should I test for < 3.7 as well? I am quite new to Python, so I would uninstall my current version of Python and install 3.6 to test it as I am not confident running multiple versions at this time.
Should I test for 3.11?

These are the steps I would follow to test:
`make clean`, followed by the `./configure` line to be tested, and `make -s -j2`.

Comment (erlend-aasland) [reply setup.rst:392]: Note, you don't need to uninstall or install a specific Python version to test these things. You only need to clone the git repo, check out the version you wish to test, then `./configure` and either manually inspect the Makefile, or just build (run `make`).

Comment (ezio-melotti) [reply setup.rst:392]: To make it more readable and easier to copy/paste, the command above can be rewritten as:
```sh
PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
./configure --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
```
Since this is a single command, the `export` shouldn't be necessary anymore.

erlend-aasland: That don't work on my system; that's why I added the explicit `export`:

```console
$ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
    ./configure --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
Package tcl was not found in the pkg-config search path.
Perhaps you should add the directory containing `tcl.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tcl' found
Package tk was not found in the pkg-config search path.
Perhaps you should add the directory containing `tk.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tk' found
Package tcl was not found in the pkg-config search path.
Perhaps you should add the directory containing `tcl.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tcl' found
Package tk was not found in the pkg-config search path.
Perhaps you should add the directory containing `tk.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tk' found

```

ezio-melotti: Thanks for verifying that!
I think that's because of the `$(...)`s in the `configure` args -- even though the `PKG_CONFIG_PATH` should be available while `configure` is executed, it's not yet available there.

This similar example seems to prove my point:
```sh
$ FOO=123 python3 -c 'import os;print(os.environ["FOO"])'
123
$ FOO=123 echo $(python3 -c 'import os;print(os.environ["FOO"])')
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.10/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'FOO'
```

@izumiberat, you can ignore my comment about removing the `export`, but the next command could still be formatted in two lines:
```sh
$ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
$ ./configure --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
```

izumiberat: I tested using `checkout` as suggested by @erlend-aasland for Python 3.4, 3.6, and 3.10 (as well as 3.8 just for fun).

Result : _tkinter gets built (this was the purpose of this PR), but I get the following warning for 3.10:

> configure: WARNING: unrecognized options: --with-tcltk-libs, --with-tcltk-includes

@erlend-aasland when closing issue https://github.com/python/cpython/issues/93659 I had tested that _tkinter got built but I hadn't provided the additional options `--with-tcltk-libs` and  `--with-tcltk-includes` in `configure`. I don't know and haven't researched yet the purpose of defining those options in `configure`. Can we omit them?

----

I copy/paste below part of the content displayed when `configure` and `make` for the different versions I tested.

**Summary**: There are a number of other issues that come up (out of the scope of this PR that I copy/paste below for reference) but it did remove the following message (which is the purpose of this PR):

```
Failed to build these modules:
_tkinter 
```    
For Python 3.4
----
> 9 warnings generated.
> ld: warning: could not create compact unwind for _ffi_call_unix64: does not use RBP or RSP based frame
> 
> Python build finished successfully!
> The necessary bits to build these optional modules were not found:
> _hashlib              _sqlite3              nis                
> ossaudiodev           spwd                  zlib               
> To find the necessary bits, look in setup.py in detect_modules() for the module's name.
> 
> The following modules found by detect_modules() in setup.py, have been
> built by the Makefile instead, as configured by the Setup files:
> atexit                pwd                   time   

For Python 3.6
----
same as above but with in addition:

> configure: WARNING: unrecognized options: --with-openssl

For Python 3.8
----
same as above but with:
_hashlib replaced by -ssl in:
> The necessary bits to build these optional modules were not found:
> _hashlib              _sqlite3              nis                
> ossaudiodev           spwd                  zlib    
 
For Python 3.10
----
`configure: WARNING: unrecognized options: --with-tcltk-libs, --with-tcltk-includes`

This is the full output when `make`:

> Note: Deepfreeze may have added some global objects,
>       so run 'make regen-global-objects' if necessary.
> /Library/Developer/CommandLineTools/usr/bin/ranlib: file: libpython3.12d.a(dynamic_annotations.o) has no symbols
> /Library/Developer/CommandLineTools/usr/bin/ranlib: file: libpython3.12d.a(pymath.o) has no symbols
>  CC='gcc' LDSHARED='gcc -bundle -undefined dynamic_lookup    ' OPT='-g -Og -Wall' 	./python.exe -E ./setup.py -q build
> ld: warning: directory not found for option '-L/usr/lib/termcap'
> renaming build/scripts-3.12/pydoc3 to build/scripts-3.12/pydoc3.12
> renaming build/scripts-3.12/idle3 to build/scripts-3.12/idle3.12
> renaming build/scripts-3.12/2to3 to build/scripts-3.12/2to3-3.12

erlend-aasland: It may be that setup.py automatically detects and uses the system provided Tcl/Tk libraries on macOS. I'm not on my laptop now, so I cannot verify it (yet).

izumiberat: @erlend-aasland when running `python3 -m tkinter`, it displays that I am using Tcl/Tk version 8.6. Looking at setup.py, it looks like, it is indeed installed by default if it doesn't find the options:

>         # Assume we haven't found any of the libraries or include files
>         # The versions with dots are used on Unix, and the versions without
>         # dots on Windows, for detection by cygwin.
>         tcllib = tklib = tcl_includes = tk_includes = None
>         for version in ['8.6', '86', '8.5', '85', '8.4', '84', '8.3', '83',
>                         '8.2', '82', '8.1', '81', '8.0', '80']:
>             tklib = self.compiler.find_library_file(lib_dirs,
>                                                         'tk' + version)
>             tcllib = self.compiler.find_library_file(lib_dirs,
>                                                          'tcl' + version)
>             if tklib and tcllib:
>                 # Exit the loop when we've found the Tcl/Tk libraries
>                 break

Not sure why specifying the options as you provided them lead to the warning:

> configure: WARNING: unrecognized options: --with-tcltk-libs, --with-tcltk-includes

erlend-aasland: > Not sure why specifying the options as you provided them lead to the warning:

Something is wrong on your end. It may be wrong checkout (maybe your were on `main` instead of 3.10?).

```console
$ git checkout -q upstream/3.10
$ ./configure --help | grep tcl
  --with-tcltk-includes='-I...'
  --with-tcltk-libs='-L...'
```

erlend-aasland: @izumiberat, can you please use `<details>` in your posts? The verbosity and length of your comments make it hard to follow the conversation.

<details>
<summary>Here's a short guide</summary>

```
<details>
<summary>Tests with Python 3.10</summary>

I did this:

    $ git checkout ...
    $ ./configure ... && make

And this happened:

    < put output of build script here >

</details>
```

</details>

setup.rst
```diff
@@ -385,17 +385,29 @@ for the header and library files to your ``configure`` command.  For example,
...
+
+For Python versions 3.6 and older::
+
+    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
     $ CPPFLAGS="-I$(brew --prefix openssl)/include" \
       LDFLAGS="-L$(brew --prefix openssl)/lib" \
-      ./configure --with-pydebug
+      ./configure --with-pydebug \
+                  --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
+                  --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
```
erlend-aasland, line 410: @ezio-melotti: can we just remove this? 3.6 is EOL.

setup.rst
```diff
@@ -385,17 +385,29 @@ for the header and library files to your ``configure`` command.  For example,
...
+      ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
-or ``configure`` Python versions < 3.7::
+For Python versions 3.10 through 3.7::
 
+    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
+    $ ./configure --with-pydebug \
+                  --with-openssl=$(brew --prefix openssl) \
+                  --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
+                  --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
```
erlend-aasland, line 401: Tested on 3.10 through 3.7 ✅

setup.rst
izumiberat: @erlend-aasland You were right! I was working on `main`. I just tested again for 3.10 and I am all good. Got it for 3.6 --`with-openssl` option. We are all good then.

setup.rst
bxsx: Can you confirm you don't get this warning?

```
./configure --with-pydebug --with-openssl=$(brew --prefix openssl@1.1) --with-tcltk-libs="$(pkg-config --libs tcl tk)" --with-tcltk-includes="$(pkg-config --cflags tcl tk)" | grep 'unrecognized options'
configure: WARNING: unrecognized options: --with-tcltk-libs, --with-tcltk-includes
```

izumiberat: @bxsx I am not getting the warning anymore. I was on `main` instead of 3.10 when getting the error as suggested by @erlend-aasland. 

This is the full output I am getting when `make -s -j2` after running your `configure` command:
<details>

> /Library/Developer/CommandLineTools/usr/bin/ranlib: file: libpython3.10d.a(dynamic_annotations.o) has no symbols
> /Library/Developer/CommandLineTools/usr/bin/ranlib: file: libpython3.10d.a(pymath.o) has no symbols
>  CC='gcc' LDSHARED='gcc -bundle -undefined dynamic_lookup    ' OPT='-g -O0 -Wall' 	_TCLTK_INCLUDES='-I/usr/local/Cellar/tcl-tk/8.6.12_1/include' _TCLTK_LIBS='-L/usr/local/Cellar/tcl-tk/8.6.12_1/lib -ltk8.6 -ltkstub8.6 -ltcl8.6 -ltclstub8.6' 	./python.exe -E ./setup.py -q build
> ld: warning: directory not found for option '-L/usr/lib/termcap'
> /Users/izumiberat/Documents/GitHub/cpython/Modules/_sqlite/module.c:141:10: warning: 
>       'sqlite3_enable_shared_cache' is deprecated: first deprecated in macOS
>       10.7 - Not supported [-Wdeprecated-declarations]
>     rc = sqlite3_enable_shared_cache(do_enable);
>          ^
> /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sqlite3.h:5950:16: note: 
>       'sqlite3_enable_shared_cache' has been explicitly marked deprecated here
> SQLITE_API int sqlite3_enable_shared_cache(int);
>                ^
> 1 warning generated.
> In file included from /Users/izumiberat/Documents/GitHub/cpython/Modules/_tkinter.c:46:
> In file included from /usr/local/Cellar/tcl-tk/8.6.12_1/include/tk.h:99:
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:131:21: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*free_private)();  /* called to free private storage */
>                            ^
>                             void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:334:33: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         struct _XImage *(*create_image)();
>                                        ^
>                                         void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:453:23: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         XID (*resource_alloc)(); /* allocator function */
>                              ^
>                               void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:471:20: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*synchandler)();   /* Synchronization handler */
>                           ^
>                            void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:496:24: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Bool (*event_vec[128])();  /* vector for wire to event */
>                               ^
>                                void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:497:25: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Status (*wire_vec[128])(); /* vector for event to wire */
>                                ^
>                                 void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:509:20: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Bool (**error_vec)();      /* vector for wire to error */
>                           ^
>                            void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:522:25: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*savedsynchandler)(); /* user synchandler when Xlib usurps */
>                                ^
>                                 void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:1053:24: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
> typedef void (*XIMProc)();
>                        ^
>                         void
> 9 warnings generated.
> In file included from /Users/izumiberat/Documents/GitHub/cpython/Modules/tkappinit.c:17:
> In file included from /usr/local/Cellar/tcl-tk/8.6.12_1/include/tk.h:99:
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:131:21: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*free_private)();  /* called to free private storage */
>                            ^
>                             void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:334:33: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         struct _XImage *(*create_image)();
>                                        ^
>                                         void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:453:23: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         XID (*resource_alloc)(); /* allocator function */
>                              ^
>                               void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:471:20: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*synchandler)();   /* Synchronization handler */
>                           ^
>                            void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:496:24: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Bool (*event_vec[128])();  /* vector for wire to event */
>                               ^
>                                void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:497:25: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Status (*wire_vec[128])(); /* vector for event to wire */
>                                ^
>                                 void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:509:20: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         Bool (**error_vec)();      /* vector for wire to error */
>                           ^
>                            void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:522:25: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
>         int (*savedsynchandler)(); /* user synchandler when Xlib usurps */
>                                ^
>                                 void
> /usr/local/Cellar/tcl-tk/8.6.12_1/include/X11/Xlib.h:1053:24: warning: this
>       function declaration is not a prototype [-Wstrict-prototypes]
> typedef void (*XIMProc)();
>                        ^
>                         void
> 9 warnings generated.
> 
> Python build finished successfully!
> The necessary bits to build these optional modules were not found:
> ossaudiodev           spwd                                     
> To find the necessary bits, look in setup.py in detect_modules() for the module's name.
> 
> 
> The following modules found by detect_modules() in setup.py, have been
> built by the Makefile instead, as configured by the Setup files:
> _abc                  pwd                   time               
> 
> renaming build/scripts-3.10/pydoc3 to build/scripts-3.10/pydoc3.10
> renaming build/scripts-3.10/idle3 to build/scripts-3.10/idle3.10
> renaming build/scripts-3.10/2to3 to build/scripts-3.10/2to3-3.10

</details>

setup.rst
bxsx: > @izumiberat 
> when running `python3 -m tkinter`, it displays that I am using Tcl/Tk version 8.6. Looking at setup.py, it looks like, it is indeed installed by default if it doesn't find the options:

What version of `tcl-tk` are you installing via Homebrew?

In my case:
```bash
$ brew info tcl-tk | head -n1
tcl-tk: stable 8.6.12 (bottled) [keg-only]
```

This seems to match the version from `python3 -m tkinter`.
When I uninstall it via `brew uninstall tcl-tk` I am no longer able to find the library [1].

So I think everything is OK here.

> @erlend-aasland 
> It may be that setup.py automatically detects and uses the system provided Tcl/Tk libraries on macOS.

In that case, there wouldn't be this issue and PR in the first place, would be?

[1] Yes, `brew  --prefix tcl-tk` still outputs a path but it does regardless the package is installed or not. So, the path doesn't exist.

setup.rst
bxsx: > @bxsx I am not getting the warning anymore. I was on `main` instead of 3.10 when getting the error as suggested by @erlend-aasland.

Oh, you are both right! Sorry, that was my fault. I confirm it works as expected!

setup.rst
bxsx: Followed up on my last 👆comment:

I tried to run `configure` with `PKG_CONFIG_PATH` in-line set on 3.7-3.10 and I am getting interesting results when running `PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" ./configure --with-pydebug --with-openssl=$(brew --prefix openssl@1.1) --with-tcltk-libs="$(pkg-config --libs tcl tk)" --with-tcltk-includes="$(pkg-config --cflags tcl tk)"`:
- 3.7-3.8 - unable to find Tcl/Tk lib, no tkinter module available (expected result)
- 3.9-3.10 - errors with `pkg-config` (as in @erlend-aasland output), however `tkinter` is compiled. When I run `./python -m tkinter` I get a warning: `DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning.` and the popup window is a bit broken (black text on black background, only buttons are visible)

Now, when I correctly  configure 3.9/3.10 (`export PKG_CONFIG_PATH`):
- I don't get the deprecation output
- Pop-up window displays correctly with black text on a gray (default) background

After inspecting the compiler output:
```console
gcc -Wno-unused-result -Wsign-compare -g -O0 -Wall -std=c99 -Wextra -Wno-unused-result -Wno-unused-parameter -Wno-missing-field-initializers -Wstrict-prototypes -Werror=implicit-function-declaration -fvisibility=hidden -I./Include/internal -DWITH_APPINIT=1 -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Tcl.framework/Headers -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Tk.framework/Headers -I./Include -I. -I/Users/bxsx/+dev/opensource/cpython/Include -I/Users/bxsx/+dev/opensource/cpython -c /Users/bxsx/+dev/opensource/cpython/Modules/_tkinter.c -o build/temp.macosx-12.4-arm64-3.10-pydebug/Users/bxsx/+dev/opensource/cpython/Modules/_tkinter.o -F /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks -Wno-strict-prototypes
```

@izumiberat I can confirm that in such a situation the default macOS `tkinter` lib is being used.

setup.rst
```diff
@@ -385,17 +385,29 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install pkg-config openssl xz gdbm
+    $ brew install pkg-config openssl xz gdbm tcl-tk
 
-and ``configure`` Python versions >= 3.7::
+For Python 3.11 and newer::
```
bxsx, line 390:
```suggestion
For Python 3.10 and newer::
```

setup.rst
```diff
@@ -385,17 +385,29 @@ for the header and library files to your ``configure`` command.  For example,
...
 
-    ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
+    $ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
+      ./configure --with-pydebug --with-openssl=$(brew --prefix openssl)
 
-or ``configure`` Python versions < 3.7::
+For Python versions 3.10 through 3.7::
```
bxsx, line 395:
```suggestion
For Python versions 3.9 through 3.7::
```

bxsx (commented):
Both instructions for 3.11+ and 3.7-3.9 work in 3.10. I suggest using the latest (and simplified) version for 3.10.

setup.rst
bxsx: @erlend-aasland I would expect a comment on why you rejected it.

setup.rst
bxsx: @erlend-aasland I would expect a comment on why you rejected it.

erlend-aasland:
> Both instructions for 3.11+ and 3.7-3.9 work in 3.10. I suggest using the latest (and simplified) version for 3.10.

No. Please read up on the previous discussion. We changed configure and setup.py in 3.11. You **can not** use the same instructions on 3.10 (and before) and in 3.11 and newer.

Also, if you build in-tree (as I expect you do), you need to **clean everything** when you switch between revisions. This is extremely important. You can clean your source tree by executing `git clean -fdx`. Here's a short recipe on how to switch between revisions and compile:

```console
$ git clone https://github.com/python/cpython.git cpython.git
$ cd cpython.git
$ ./configure && make
$ git checkout origin/3.10
$ git clean -fdx  # <= this is important! make clean is _not_ enough
$ ./configure && make
```

bxsx:
@erlend-aasland 

I did the following:
```console
$ rm -rf cpython
$ git clone https://github.com/python/cpython
$ cd cpython
$ git checkout origin/3.10
$ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" ./configure --with-pydebug --with-openssl=$(brew --prefix openssl@1.1)
$ make -j
$ ./python.exe --version
```

Selected output:
```
CC='gcc' LDSHARED='gcc -bundle -undefined dynamic_lookup    ' OPT='-g -O0 -Wall' 	_TCLTK_INCLUDES='-I/opt/homebrew/Cellar/tcl-tk/8.6.12_1/include' _TCLTK_LIBS='-L/opt/homebrew/Cellar/tcl-tk/8.6.12_1/lib -ltk8.6 -ltkstub8.6 -ltcl8.6 -ltclstub8.6' 	./python.exe -E ./setup.py  build
gcc -Wno-unused-result -Wsign-compare -g -O0 -Wall -std=c99 -Wextra -Wno-unused-result -Wno-unused-parameter -Wno-missing-field-initializers -Wstrict-prototypes -Werror=implicit-function-declaration -fvisibility=hidden -I./Include/internal -DWITH_APPINIT=1 -I./Include -I. -I/Users/bxsx/+dev/opensource/cpython/Include -I/Users/bxsx/+dev/opensource/cpython -c /Users/bxsx/+dev/opensource/cpython/Modules/_tkinter.c -o build/temp.macosx-12.4-arm64-3.10-pydebug/Users/bxsx/+dev/opensource/cpython/Modules/_tkinter.o -I/opt/homebrew/Cellar/tcl-tk/8.6.12_1/include
gcc -Wno-unused-result -Wsign-compare -g -O0 -Wall -std=c99 -Wextra -Wno-unused-result -Wno-unused-parameter -Wno-missing-field-initializers -Wstrict-prototypes -Werror=implicit-function-declaration -fvisibility=hidden -I./Include/internal -DWITH_APPINIT=1 -I./Include -I. -I/Users/bxsx/+dev/opensource/cpython/Include -I/Users/bxsx/+dev/opensource/cpython -c /Users/bxsx/+dev/opensource/cpython/Modules/tkappinit.c -o build/temp.macosx-12.4-arm64-3.10-pydebug/Users/bxsx/+dev/opensource/cpython/Modules/tkappinit.o -I/opt/homebrew/Cellar/tcl-tk/8.6.12_1/include
gcc -bundle -undefined dynamic_lookup build/temp.macosx-12.4-arm64-3.10-pydebug/Users/bxsx/+dev/opensource/cpython/Modules/_tkinter.o build/temp.macosx-12.4-arm64-3.10-pydebug/Users/bxsx/+dev/opensource/cpython/Modules/tkappinit.o -L/usr/local/lib -o build/lib.macosx-12.4-arm64-3.10-pydebug/_tkinter.cpython-310d-darwin.so -L/opt/homebrew/Cellar/tcl-tk/8.6.12_1/lib -ltk8.6 -ltkstub8.6 -ltcl8.6 -ltclstub8.6
Python 3.10.5+
```

Running `./python.exe -m tkinter` behaves **the same** way as in 3.11:
- No deprecation warning (this happens with system TclTk libs)
- Pop-up displayed correctly
(background https://github.com/python/devguide/pull/893#discussion_r895743236)

Therefore, I still believe that the instructions for 3.11+ works in 3.10.

erlend-aasland:
Ah, yes, there were basic support for pkg-config in `configure` in 3.10. My bad. However, using `--with-tcl-*` also works in 3.10. Both approaches works just fine; both get the job done (which is the point of importance). I don't want to touch the PR more than needed now. The PR is correct as it stands now. I'll wait for Ezio's final thumbs up before merging.

bxsx:
I think the point is to have a nice and compact way to get the job done.

```console
$ PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" configure --with-pydebug
```
vs
```console
$ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"
$ ./configure --with-pydebug \
```

They are not the same. The latter includes 2 commands instead of 1 command to run + forces `PKG_CONFIG_PATH` variable to be exported. So the developer experience is much different.
We will also be shading the latter in the future, so it's good to promote the new way to configure the build as the instructions will be updated anyway (like with 3.9 EOL).

I would agree it doesn't matter if it takes a lot of effort, but the fix is already here: https://github.com/python/devguide/pull/893#pullrequestreview-1004370758

erlend-aasland:
> I think the point is to have a nice and compact way to get the job done.

I've updated the instructions to include 3.10. Thanks.

ezio-melotti (approved):
The PR looks good to me.  I don't think the 3.6 section needs to be included, so you can remove that.

PR #896, opened 2022-06-14
Hello Ukrainian translation.

JulienPalard:
Just telling the Ukrainian translation may start.

(I think it's usefull to maintain this list, so if someone land here with the question "is my language being translated yet?" he has good pointers.)

### DIFF
--- a/documenting.rst
+++ b/documenting.rst
@@ -1658,6 +1658,8 @@ in production; others are works in progress.
 | Turkish (tr)    | `Ege Akman (@egeakman)        | `GitHub <github_tr_>`_     |
 |                 | <email_egeakman_>`_           |                            |
 +-----------------+-------------------------------+----------------------------+
+| Ukrainian       |                               | `mail <mail_uk_>`_         |
++-----------------+-------------------------------+----------------------------+
 
 .. _article_pt_br: https://rgth.co/blog/python-ptbr-cenario-atual/
 .. _gh_cocoatomo: https://github.com/cocoatomo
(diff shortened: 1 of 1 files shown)


documenting.rst
```diff
@@ -1698,6 +1700,7 @@ in production; others are works in progress.
 .. _mail_lt: https://mail.python.org/pipermail/doc-sig/2019-July/004138.html
 .. _mail_pl: https://mail.python.org/pipermail/doc-sig/2019-April/004106.html
 .. _mail_ru: https://mail.python.org/pipermail/doc-sig/2019-May/004131.html
+.. _mail_uk: https://mail.python.org/archives/list/translation@python.org/thread/OCSM3TWNKO5FAJ3WOLP7U7CLCHCSJILP/
```
zware, line 1703: This appears to be a private mailing list, which is less than ideal for a public link.  If it's what we've got it's what we've got, though.

ezio-melotti: I also wonder if it would be better to wait until there is at least a GitHub repo, which most projects seem to have even if they are doing the actual translations on Transifex.  Some of these projects list only an email from years ago with no followup.

JulienPalard: >  Some of these projects list only an email from years ago with no followup.

Which may be good (or may be cleaned up after a few years). I mean if one wants to start a translation but not alone, easily finding an email of someone having the same idea is motivating.

JulienPalard: > This appears to be a private mailing list

Oops, I didn't even noticed!

JulienPalard:
Closing as the only mail is in a private mailing list :(

PR #897, opened 2022-06-15
Translations spring clean.

JulienPalard:
As mentionned in https://github.com/python/devguide/pull/896 some links were outdated.

Also keeping a link to an old initial email for a now mature translation with a repo and a presence on docs.python.org does not make sense.

### DIFF
--- a/documenting.rst
+++ b/documenting.rst
@@ -1615,51 +1615,39 @@ in production; others are works in progress.
 +-----------------+-------------------------------+----------------------------+
 | Indonesian (id) | `Oon Arfiandwi <gh_oonid_>`_  | `GitHub <github_id_>`_     |
 +-----------------+-------------------------------+----------------------------+
-| Italian (it)    |                               | `mail <mail_it_>`_         |
-+-----------------+-------------------------------+----------------------------+
 | Japanese (ja)   | `Kinebuchi Tomohiko           | `GitHub <github_ja_>`_     |
-|                 | (@cocoatomo) <gh_cocoatomo_>`_| `Doc <doc_ja_>`_           |
+|                 | (@cocoatomo) <gh_cocoatomo_>`_|                            |
 +-----------------+-------------------------------+----------------------------+
 | Korean (ko)     |                               | `GitHub <github_ko_>`_     |
-|                 |                               | `Doc <doc_ko_>`_           |
 +-----------------+-------------------------------+----------------------------+
 | Marathi (mr)    | `Sanket Garade                | `GitHub <github_mr_>`_     |
 |                 | <email_garade_>`_             |                            |
-+-----------------+-------------------------------+----------------------------+
-| Lithuanian (lt) |                               | `mail <mail_lt_>`_         |
 +-----------------+-------------------------------+----------------------------+
 | Persian (fa)    | `Komeil Parseh (@mmdbalkhi)   | `GitHub <github_fa_>`_     |
 |                 | <gh_mmdbalkhi_>`_             |                            |
 +-----------------+-------------------------------+----------------------------+
 | Polish (pl)     |                               | `GitHub <github_pl_>`_     |
 |                 |                               | `Translations <tx_pl_>`_   |
-|                 |                               | `Doc <doc_pl_>`_           |
-|                 |                               | `mail <mail_pl_>`_         |
-+-----------------+-------------------------------+----------------------------+
-| Portuguese (pt) | Gustavo Toffo                 |                            |
 +-----------------+-------------------------------+----------------------------+
 | Portuguese      | Marco Rougeth                 | `GitHub <github_pt_br_>`_  |
 | as spoken       |                               | `Wiki <wiki_pt_br_>`_      |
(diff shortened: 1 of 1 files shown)


documenting.rst
```diff
@@ -1672,11 +1660,6 @@ in production; others are works in progress.
 .. _email_egeakman: mailto:egeakmanegeakman@hotmail.com
 .. _email_garade: mailto:garade@pm.me
 .. _chat_pt_br: https://t.me/pybr_i18n
-.. _doc_ja: https://docs.python.org/ja/
-.. _doc_ko: https://docs.python.org/ko/
-.. _doc_pl: https://docs.python.org/pl/
-.. _doc_zh_cn: https://docs.python.org/zh-cn/
-.. _doc_zh_tw: https://docs.python.org/zh-tw/
```
ezio-melotti, line 1679: Why removing these?

JulienPalard: They look too obvious to be mentionned.

ezio-melotti: I found them somewhat handy for two reasons:
1. they show which translations have been already published;
2. they provide an easy way to check the current published translation (instead of having to open d.p.o in a separate tab and selecting the language);

I think they might even deserve their own column.

ezio-melotti: Good idea!

ezio-melotti:
This has been moved to https://devguide.python.org/documentation/translating/ -- @JulienPalard do you want to update the PR, including the changes suggested in the review?

ezio-melotti:
There are also a few other improvements that can be done:
* some translators have links to their email, others to their GitHub page, others don't have any links -- these could be cleaned up and made more consistent (e.g. for those that have a GitHub link, add the link to the `@username`)
* the transifex link is repeated since multiple languages use it -- this could either be deduplicated, or links to the specific languages could be added if possible
* the links could be comma-separated, or possibly more specific columns could be created

gvanrossum:
Is this PR still needed? @JulienPalard

safwansamsudeen:
If @JulienPalard isn't assigned to this anymore, can I work on it (if you feel it's required)?

CAM-Gerlach:
Sure, @safwansamsudeen . Do you have a clear idea of what needs to be done here? It should be mostly mechanical in nature, but you'll need to (manually, but its mostly mechanical) integrate these changes with the latest version of the translations table (now in [its own separate file](https://github.com/python/devguide/blob/main/documentation/translating.rst)), as well as make the requested changes (or at least the first two, the third might take a little more effort and could be done separately):

* Move the Python docs links to the text in the first column instead of at the end and linking them consistently, e.g.

  ```rst
  +-----------------+-------------------------------+----------------------------+
  | `Spanish (es)`  | Raúl Cumplido                 | `GitHub <github_es_>`_     |
  +-----------------+-------------------------------+----------------------------+
  
  .. _Spanish (es): https://docs.python.org/es/
  ```
  
* Separate the remaining links on the right column by commas, and ensure the titles are consistent
  
* Ensure there are consistent GitHub usernames and links for each of the translators in the middle column

I suggest you start in a fresh branch as that's probably simpler for you at this point; @JulienPalard can be added as a co-author to the commit metadata if he's okay with that.

We're happy to help further if you need additional guidance!

safwansamsudeen:
Many of the languages don't seem to have the docs online. I checked Arabic, Bengali, and Hindi, and all of them - when I follow the pattern `https://docs.python.org/code/` - raise a 404.

What should I do about this?

CAM-Gerlach:
> How do I know which languages are published to see whether or not I should add the link?

You can tell what languages are published because they appear under the language selector on the [docs homepage](https://docs.python.org/3/index.html). The ones I see visible are these:

image

Also, you can find the links to the GitHub repos for the various language under [a filtered version of the Python org repository list](https://github.com/orgs/python/repositories?q=python-docs&type=all&language=&sort=).

> EDIT: also, how do I edit the table easily - there seems to require a lot of effort managing the spaces and | aligning.

Yeah, it can be quite tedious. I find it much easier in an editor (like Notepad++ on Windows, and many others) that offer a column mode and/or multi-line cursor features.

You could also convert it to a [list table](https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table) which avoids this problem and is generally much easier to work with, but you'd want to make a separate PR converting the table to list-table syntax first with no changes, have that merged, and then manually re-make these changes in another PR on top of that, which would be a good deal more time and effort on your part.

ezio-melotti:
Several editors have an edit mode that makes editing tables easier by allowing you to edit and move columns, e.g. https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_column-box-selection
In this case, adding links to the existing table should be pretty straightforward, so it might be easier to stick with it rather than rewriting it as a list table.

safwansamsudeen:
@CAM-Gerlach thanks for your help! I think converting into a list table would involve much more - IMO unproductive - effort. @ezio-melotti's solution sounds good, only I use IntelliJ, but I'll see.

Issue #890, opened 2022-06-09
Document `git switch` instead of `git checkout`


ezio-melotti:
Some uses of the `checkout` command have been replaced by the `switch` command.  The `switch` command has been introduced in [git 2.23](https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt#L61) (released in 2019-08-16) and by now it should be widespread enough that we can update the docs using `switch`, possibly with a note about the old command.

PR #905, opened 2022-06-23
gh-890: Use git switch to switch to branch

za:
instead of using git checkout

https://github.com/python/devguide/issues/890

### DIFF
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -103,10 +103,9 @@ Creating and Switching Branches
 .. important::
    Never commit directly to the ``main`` branch.
 
-Create a new branch and switch to it::
+Create a new branch from ``main`` and switch to it::
 
-   # creates a new branch off main and switch to it
-   git checkout -b <branch-name> main
+   git switch -c <branch-name> main
 
 This is equivalent to::
 
@@ -133,7 +132,7 @@ To switch to a different branch::
 Other releases are just branches in the repository.  For example, to work
 on the 2.7 release from the ``upstream`` remote::
 
-   git checkout -b 2.7 upstream/2.7
+   git switch -c 2.7 upstream/2.7
 
 .. _deleting_branches:
 
--- a/gitbootcamp.rst
+++ b/gitbootcamp.rst
@@ -112,7 +112,10 @@ This is equivalent to::
 
    # create a new branch from main, without checking it out
    git branch <branch-name> main
+
    # check out the branch
+   git switch <branch-name>
+   # or using old command
    git checkout <branch-name>
 
 To find the branch you are currently on::
@@ -128,6 +131,8 @@ To list all the branches, including the remote branches::
 
 To switch to a different branch::
 
+   git switch <another-branch-name>
+   # or using old command
    git checkout <another-branch-name>
 
 Other releases are just branches in the repository.  For example, to work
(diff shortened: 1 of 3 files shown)


terryjreedy:
To me, a non-expert get user, this creates confusion.  Is there any difference?  Any advantage either way?  Need both alternatives be given every time?

za:
@terryjreedy There's no difference. `git switch` and `git restore` was introduced in git 2.23. I think, we don't need both alternatives given every time.

encukou:
Right. IMO it would be best to remove `git checkout` from the bootcamp entirely. New users don't need to learn about it at all.

Background: `git checkout` does two very different things depending on how you call it (switching branches and restoring changed files). Git developers replaced it by two separate commands (and took the opportunity to also change all the options, so they make more sense). They're keeping `checkout` around for existing scripts and finger memories.

gitbootcamp.rst
```diff
@@ -112,7 +112,10 @@ This is equivalent to::
 
    # create a new branch from main, without checking it out
    git branch <branch-name> main
+
    # check out the branch
+   git switch <branch-name>
+   # or using old command
    git checkout <branch-name>
```
ezio-melotti, line 119:
```suggestion
   # create a new branch from main
   git branch <branch-name> main
   # switch to the new branch
   git switch <branch-name>
```

gitbootcamp.rst
```diff
@@ -128,6 +131,8 @@ To list all the branches, including the remote branches::
 
 To switch to a different branch::
 
+   git switch <another-branch-name>
+   # or using old command
    git checkout <another-branch-name>
```
ezio-melotti, line 136:
```suggestion
```

pullrequest.rst
```diff
@@ -110,7 +110,11 @@ You should have already :ref:`set up your system <setup>`,
   to your branch, making more changes, committing them, and pushing them to
   automatically update your PR::
 
+   # switch to your branch
+   git switch <branch-name>
+   # or using old command
    git checkout <branch-name>
+
```
ezio-melotti, line 117:
```suggestion
   git switch <branch-name>
```

pullrequest.rst
```diff
@@ -129,7 +133,10 @@ You should have already :ref:`set up your system <setup>`,
   will show a warning to this end and you may be asked to address this. Merge
   the changes from the main branch while resolving the conflicts locally::
 
+   git switch <branch-name>
+   # or using old command
    git checkout <branch-name>
+
```
ezio-melotti, line 139:
```suggestion
```

setup.rst
```diff
@@ -97,7 +97,7 @@ in the ``cpython`` directory and two remotes that refer to your own GitHub fork
 If you want a working copy of an already-released version of Python,
 i.e., a version in :ref:`maintenance mode <maintbranch>`, you can checkout
 a release branch. For instance, to checkout a working copy of Python 3.8,
-do ``git checkout 3.8``.
+do ``git switch 3.8`` or using the old comand ``git checkout 3.8``.
```
ezio-melotti, line 100:
```suggestion
do ``git switch 3.8``.
```

ezio-melotti (changes requested):

ezio-melotti (approved):

erlend-aasland (approved):

Issue #14, opened 2016-07-22
include rendered output in addition to markup


brettcannon:
Moved from https://github.com/python/cpython/issues/58426

FormulaCarbon:
Is this still needed? It looks like someone has already fixed it.

hauntsaninja:
Thanks for mentioning, looks like it's been fixed for a while :-)

PR #916, opened 2022-07-13
Add a quick reference for rst markup.

ezio-melotti:
This PR adds a quick reference table at the top of the rst reference page.  The goal of this table is to show some commonly used markup and to include links to the relevant sections.  I also added some internal references and changed a title.

This was proposed a decade ago on IRC, and then captured in https://github.com/python/cpython/issues/58426#issuecomment-1093572689 [^1]

[^1]: note that the issue itself (now transferred to #14) is probably obsolete, and this proposal doesn't aim to solve that.

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -16,22 +16,22 @@ Quick Reference
 This table summarizes which markup should be used for some commonly used
 elements:
 
-===================== =========================================== ====================
-Element               Markup                                      See also
-===================== =========================================== ====================
-arguments/parameters  ``*arg*``                                   :ref:`inline-markup`
-vars/literals/code    ````var````, ````42````, ````len(s) - 1```` :ref:`inline-markup`
-True/False/None       ````True````, ````False````, ````None````   :ref:`inline-markup`
-functions             ``:func:`sorted```                          :ref:`roles`
-func definitions      ``.. function:: func(args)``                :ref:`directives`
-reference labels      ``.. _label-name:``                         :ref:`doc-ref-role`
-internal ref          ``:ref:`label-name```                       :ref:`doc-ref-role`
-external links        ```Link text <https://target>`_``           :ref:`hyperlinks`
-roles w/ custom text  ``:role:`custom text <target>```            :ref:`roles`
-roles w/o link        ``:role:`!target```                         :ref:`roles`
-roles w/o first part  ``:role:`~hidden.hidden.visible```          :ref:`roles`
-comments              ``.. a comment``                            :ref:`comments`
-===================== =========================================== ====================
+======================= =========================================== ====================
+Element                 Markup                                      See also
+======================= =========================================== ====================
+arguments/parameters    ``*arg*``                                   :ref:`inline-markup`
+variables/literals/code ````foo````, ````42````, ````len(s) - 1```` :ref:`inline-markup`
+True/False/None         ````True````, ````False````, ````None````   :ref:`inline-markup`
+functions definitions   ``.. function:: print(*args)``              :ref:`directives`
+functions references    ``:func:`print```                           :ref:`roles`
+reference labels        ``.. _label-name:``                         :ref:`doc-ref-role`
+internal references     ``:ref:`label-name```                       :ref:`doc-ref-role`
+external links          ```Link text <https://example.com>`_``      :ref:`hyperlinks`
+roles w/ custom text    ``:role:`custom text <target>```            :ref:`roles`
+roles w/ only last part ``:role:`~hidden.hidden.visible```          :ref:`roles`
+roles w/o link          ``:role:`!target```                         :ref:`roles`
+comments                ``.. a comment``                            :ref:`comments`
+======================= =========================================== ====================
 
 
 .. _rst-primer:
@@ -263,10 +263,8 @@ You can also explicitly number the footnotes for better context.
 Comments
 --------
 
-Every explicit markup block which isn't a valid markup construct (like the
-footnotes above) is regarded as a comment.
-
-Comments are generally preceeded by ``..``::
+Every explicit markup block (starting with :literal:`.. \ `) which isn't a
+:ref:`valid markup construct <directives>` is regarded as a comment::
 
    .. This is a comment
 
@@ -678,8 +676,12 @@ Roles
 As said before, Sphinx uses interpreted text roles to insert semantic markup in
 documents.
 
-Names of local variables, such as function/method arguments, are an exception,
-they should be marked simply with ``*var*``.
+There are a couple of cases where roles can be omitted in favor of
+simpler markup:
+
+* Function and method arguments can be marked with ``*arg*``.
+* ``True``, ``False``, ``None``, and variables can be marked with
+  :literal:`\`\`...```.
 
 For all other roles, you have to write ``:rolename:`content```.
 
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -671,11 +671,11 @@ include files should be placed in the ``Doc/includes`` subdirectory.
 Roles
 -----
 
-As :ref:`previously mentioned <inline-markup>`,  Sphinx uses
+As :ref:`previously mentioned <inline-markup>`, Sphinx uses
 interpreted text roles of the form ``:rolename:`content```
 to insert semantic markup in documents.
 
-In the CPython documentation, there are a few common cases
+In the CPython documentation, there are a couple common cases
 where simpler markup should be used:
 
 * ``*arg*`` (rendered as *arg*) for function and method arguments.
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -69,8 +69,6 @@ The standard reST inline markup is quite simple: use
 * two asterisks: ``**text**`` for strong emphasis (boldface), and
 * backquotes: ````text```` for code samples, variables, and literals.
 
-Italic is also used for function and method arguments (``*arg*``).
-
 If asterisks or backquotes appear in running text and could be confused with
 inline markup delimiters, they have to be escaped with a backslash.
 
@@ -673,17 +671,15 @@ include files should be placed in the ``Doc/includes`` subdirectory.
 Roles
 -----
 
-As said before, Sphinx uses interpreted text roles to insert semantic markup in
-documents.
+As :ref:`previously mentioned <inline-markup>`,  Sphinx uses
+interpreted text roles of the form ``:rolename:`content```
+to insert semantic markup in documents.
 
-There are a couple of cases where roles can be omitted in favor of
-simpler markup:
+In the CPython documentation, there are a few common cases
+where simpler markup should be used:
 
-* Function and method arguments can be marked with ``*arg*``.
-* ``True``, ``False``, ``None``, and variables can be marked with
-  :literal:`\`\`...```.
-
-For all other roles, you have to write ``:rolename:`content```.
+* ``*arg*`` (rendered as *arg*) for function and method arguments.
+* ````True````/````False````/````None```` for ``True``/``False``/``None``.
 
 There are some additional facilities that make cross-referencing roles more
 versatile:
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -10,6 +10,29 @@ This document describes the custom reStructuredText markup introduced by Sphinx
 to support Python documentation and how it should be used.
 
 
+Quick Reference
+===============
+
+This table summarizes which markup should be used for some commonly used
+elements:
+
+================ ========================================= ====================
+Element          Markup                                    See also
+================ ========================================= ====================
+arguments        ``*arg*``                                 :ref:`inline-markup`
+variables        ````var````                               :ref:`inline-markup`
+literals         ````0````, ````[]````                     :ref:`inline-markup`
+True/False/None  ````True````, ````False````, ````None```` :ref:`inline-markup`
+code snippets    ````print('hello world')````              :ref:`inline-markup`
+functions        ``:func:`sorted```                        :ref:`roles`
+func definitions ``.. function:: func(args)``              :ref:`directives`
+reference labels ``.. _label-name:``                       :ref:`doc-ref-role`
+internal ref     ``:ref:`label-name```                     :ref:`doc-ref-role`
+external links   ```Link text <https://target>`_``         :ref:`hyperlinks`
+comments         ``.. a comment``                          :ref:`comments`
+================ ========================================= ====================
+
+
 .. _rst-primer:
 
 reStructuredText Primer
@@ -34,6 +57,7 @@ chunks of text separated by one or more blank lines.  As in Python, indentation
 is significant in reST, so all lines of the same paragraph must be left-aligned
 to the same level of indentation.
 
+.. _inline-markup:
 
 Inline markup
 -------------
@@ -42,7 +66,9 @@ The standard reST inline markup is quite simple: use
 
 * one asterisk: ``*text*`` for emphasis (italics),
 * two asterisks: ``**text**`` for strong emphasis (boldface), and
-* backquotes: ````text```` for code samples.
+* backquotes: ````text```` for code samples, variables, and literals.
+
+Italic is also used for function and method arguments (``*arg*``).
 
 If asterisks or backquotes appear in running text and could be confused with
 inline markup delimiters, they have to be escaped with a backslash.
@@ -132,6 +158,7 @@ The handling of the ``::`` marker is smart:
 That way, the second sentence in the above example's first paragraph would be
 rendered as "The next paragraph is a code sample:".
 
+.. _hyperlinks:
 
 Hyperlinks
 ----------
@@ -185,6 +212,7 @@ indentation.  (There needs to be a blank line between explicit markup and normal
 paragraphs.  This may all sound a bit complicated, but it is intuitive enough
 when you write it.)
 
+.. _directives:
 
 Directives
 ----------
@@ -229,12 +257,17 @@ body at the bottom of the document after a "Footnotes" rubric heading, like so::
 
 You can also explicitly number the footnotes for better context.
 
+.. _comments:
 
 Comments
 --------
 
 Every explicit markup block which isn't a valid markup construct (like the
 footnotes above) is regarded as a comment.
+
+Comments are generally preceeded by ``..``::
+
+   .. This is a comment
 
 
 Source encoding
@@ -636,9 +669,10 @@ The file name is relative to the current file's path.  Documentation-specific
 include files should be placed in the ``Doc/includes`` subdirectory.
 
 .. _rest-inline-markup:
+.. _roles:
 
-Inline markup
--------------
+Roles
+-----
 
 As said before, Sphinx uses interpreted text roles to insert semantic markup in
 documents.
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -16,21 +16,22 @@ Quick Reference
 This table summarizes which markup should be used for some commonly used
 elements:
 
-================ ========================================= ====================
-Element          Markup                                    See also
-================ ========================================= ====================
-arguments        ``*arg*``                                 :ref:`inline-markup`
-variables        ````var````                               :ref:`inline-markup`
-literals         ````0````, ````[]````                     :ref:`inline-markup`
-True/False/None  ````True````, ````False````, ````None```` :ref:`inline-markup`
-code snippets    ````print('hello world')````              :ref:`inline-markup`
-functions        ``:func:`sorted```                        :ref:`roles`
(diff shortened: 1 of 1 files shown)


documentation/markup.rst
```diff
@@ -10,6 +10,29 @@ This document describes the custom reStructuredText markup introduced by Sphinx
...
+literals         ````0````, ````[]````                     :ref:`inline-markup`
+True/False/None  ````True````, ````False````, ````None```` :ref:`inline-markup`
+code snippets    ````print('hello world')````              :ref:`inline-markup`
+functions        ``:func:`sorted```                        :ref:`roles`
+func definitions ``.. function:: func(args)``              :ref:`directives`
+reference labels ``.. _label-name:``                       :ref:`doc-ref-role`
+internal ref     ``:ref:`label-name```                     :ref:`doc-ref-role`
+external links   ```Link text <https://target>`_``         :ref:`hyperlinks`
+comments         ``.. a comment``                          :ref:`comments`
+================ ========================================= ====================
```
ezio-melotti, line 33: Some bikeshedding about the table:
* "Element" and "See also" could be replaced with other terms if you have better ideas.
* I tried to add an example of a string (` ````'string'```` `), but that doesn't seem to be possible without using ugly workarounds, so I just skipped it.
* I initially included more roles/directives (e.g. for methods/classes), but then decided to remove them after adding the "See also" column with the link to the relative sections.

CAM-Gerlach: "Element" is good, IMO, and See Also is fine (though it *could* be replaced with something like "Reference" or "Section", I don't have a strong opinion).

IMO, I'd think it would be more useful to add methods and classes (and perhaps more) instead of repeating literals, variables _and_ code snippits (which my other suggestion consolidates into one)—those not only point to the same section, but actually are the exact same syntax, and it also makes the contrast with them and function parameters more clear, and the inclusion of `True`/`False`/`None` more purposeful (since otherwise, the reader would have no basis on which to think they would be formatted differently than other code/literals, and thus be confused at their explicit inclusion).

documentation/markup.rst
```diff
@@ -42,7 +66,9 @@ The standard reST inline markup is quite simple: use
 
 * one asterisk: ``*text*`` for emphasis (italics),
 * two asterisks: ``**text**`` for strong emphasis (boldface), and
-* backquotes: ````text```` for code samples.
+* backquotes: ````text```` for code samples, variables, and literals.
+
+Italic is also used for function and method arguments (``*arg*``).
```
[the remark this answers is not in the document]
CAM-Gerlach: The one thing to think about here is that this section currently is nearly word for word identical to the corresponding one in the [Sphinx primer](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup), and these would be the only CPython-specific additions. Depending on what we decide to do about that, this could be either desirable or not, as on one hand only including it below would allow us to elide the section completely or at least just make it a link to the Sphinx version and keep everything in one place, but on the other it would give us _something_ to say here if we keep the section.

Also, to note, the mention in the roles section is not consistent with here, and directly conflicts with what is stated in the table, as it says:

> Names of local variables, such as function/method arguments, are an exception,
they should be marked simply with ``*var*``.

When the table above says that variables should be formatted as inline literals:

> variables ->  ``` ``var`` ```

The second (as implemented here) seems simpler, more sensible and more consistent with what is actually done in CPython's docs, so the other one should be modified accordingly (assuming we keep it around).

documentation/markup.rst
```diff
@@ -10,6 +10,29 @@ This document describes the custom reStructuredText markup introduced by Sphinx
...
+Quick Reference
+===============
+
+This table summarizes which markup should be used for some commonly used
+elements:
+
+================ ========================================= ====================
+Element          Markup                                    See also
+================ ========================================= ====================
+arguments        ``*arg*``                                 :ref:`inline-markup`
```
CAM-Gerlach, line 22:
```suggestion
parameters       ``*arg*``                                 :ref:`inline-markup`
```

Shouldn't this be parameters *of* the function, not arguments *to* the function?

documentation/markup.rst
```diff
@@ -10,6 +10,29 @@ This document describes the custom reStructuredText markup introduced by Sphinx
...
+variables        ````var````                               :ref:`inline-markup`
+literals         ````0````, ````[]````                     :ref:`inline-markup`
+True/False/None  ````True````, ````False````, ````None```` :ref:`inline-markup`
+code snippets    ````print('hello world')````              :ref:`inline-markup`
```
CAM-Gerlach, line 26: Since variables, literals and code snippits are all formatted the same and link to the same section, IMO we should just consolidate them; for example (squished a bit to fit within the existing column width constraints):

```suggestion
var/literal/code ````spam````, ````0````, ````1 + 1````    :ref:`inline-markup`
True/False/None  ````True````, ````False````, ````None```` :ref:`inline-markup`
```

CAM-Gerlach (commented):
IMO, this PR is definitely a great improvement from a reader quick-reference perspective.

I do think, like for PEP 12, we shouldn't duplicate the Sphinx reST primer nearly word for word in most places, as opposed to just linking to it (via Intersphinx) and focusing on the guidance and conventions unique and specific to the CPython docs. I've opened python/docs-community# to discuss that issue further.

documentation/markup.rst
erlend-aasland: In light of CAM's `!` hint, I would propose that we instead promote these versions for `None`, `True`, and `False`:

```
:const:`!None`
:const:`!True`
:const:`!False`
```

documentation/markup.rst
ezio-melotti: I guess it could be both, because if the function defines a *foo* parameter, it also accepts a *foo* argument.  Using "parameters" might lead people to think that they should use the markup in the function signature.  Perhaps we could specify both as "arguments/parameters"?

documentation/markup.rst
ezio-melotti: Regarding the consolidation proposed by @CAM-Gerlach, the reasons for splitting those up were:
* having more space for names in the first column
* having more space for additional examples
* making them slightly easier to find

Regarding `True`/`False`/`None`, I think we should stick to ` ``...`` `:
* it's already fairly used consistently throughout the docs
* it's easier to type/remember (especially considering how common they are)

erlend-aasland: Fair enough :)

CAM-Gerlach: Thanks for elaborating on the reasoning here.

> having more space for names in the first column

Yeah, but if we made the above ``arguments/parameters`` as you suggested, ``variable/literal/code`` would only be one character wider, so the cost would then be minimal.

> having more space for additional examples

Well, I'm not sure we really need more than one of each, since they each represent the three basic classes of ``code``-formatted objects; I'm not sure the additional examples add much additional clarity.

> making them slightly easier to find

Perhaps a little bit, but the fact that it takes up three lines in the table instead of one makes it more to navigate overall and thus everything else harder to find as quickly.

IMO, I still think if we are going to add two more lines, I'd think adding `:class:` and `:meth:` would be a lot more valuable than the marginal advantages repeating variables (which are not even really used much outside of function params/args), literals and code as separate lines would bring. But what do you think @erlend-aasland ?

> Regarding `True`/`False`/`None`, I think we should stick to `` ``...`` ``:

Agreed :+1:

documentation/markup.rst
CAM-Gerlach: > I guess it could be both, because if the function defines a foo parameter, it also accepts a foo argument. 

Yeah, I guess so—unless it is positional-only, where the name only matters for the parameter.

> Using "parameters" might lead people to think that they should use the markup in the function signature.

Well, it seems rather unlikely at least to me that a Python documented would adopt this misconception, particularly if they looked at any existing function signature in the docs, but its not impossible I suppose.

> Perhaps we could specify both as "arguments/parameters"?

That sounds good, we'd just need to widen the table column—which would also give more room for my combined suggestion.

documentation/markup.rst
ezio-melotti: > IMO, I'd think it would be more useful to add methods and classes (and perhaps more) 

The problem is that fitting `func`/`meth`/`cls` on one line is doable if we widen the table, but for the corresponding directive there's not enough space to add an example for each.  If we add all of them on separate lines, we will end up with a total of 6 total lines.  Listing all 3 together under the first column and only providing one example (e.g. using `:func:` and `.. function::` seems misleading.

documentation/markup.rst
erlend-aasland: > But what do you think @erlend-aasland ?

Sorry, I don't have the bandwidth to participate here for the next few days; I trust your judgement here, CAM.

ezio-melotti:
Here is a new iteration:
image

* added `parameters` next to `arguments`
* merged `vars/literals/code`
* added examples of roles with custom text, without links, without the first part[^1].

I didn't add methods/classes as discussed in a previous comment.  The goal of this table is to provide a quick reminder for the syntax (e.g. how do I create a link? or a coment? was it `.. directive:` or `.. directive::`?) -- additional roles (like `meth` or `class`) are just a click away.

[^1]: the descriptions in the first column are not too accurate, but IMHO functional enough.  People that already know that e.g. there's a way to only show the last component of an object will recognize the entry and see the syntax in the table; people that are not familiar with the feature or find the description confusing can always click the link and find a more exhaustive description.

documentation/markup.rst
CAM-Gerlach: Oh, I originally meant adding functions, classes and methods as separate rows instead of variables, literals and code as separate rows. But I forgot about also adding references to them, which would mean 6 total rows which is a lot. So I agree with what you ended up doing :+1:

documentation/markup.rst
```diff
@@ -10,6 +10,30 @@ This document describes the custom reStructuredText markup introduced by Sphinx
...
+Element               Markup                                      See also
+===================== =========================================== ====================
+arguments/parameters  ``*arg*``                                   :ref:`inline-markup`
+vars/literals/code    ````var````, ````42````, ````len(s) - 1```` :ref:`inline-markup`
+True/False/None       ````True````, ````False````, ````None````   :ref:`inline-markup`
+functions             ``:func:`sorted```                          :ref:`roles`
+func definitions      ``.. function:: func(args)``                :ref:`directives`
+reference labels      ``.. _label-name:``                         :ref:`doc-ref-role`
+internal ref          ``:ref:`label-name```                       :ref:`doc-ref-role`
+external links        ```Link text <https://target>`_``           :ref:`hyperlinks`
```
CAM-Gerlach, line 29:
```suggestion
external links        ```Link text <https://example.com>`_``      :ref:`hyperlinks`
```

Use a proper URL and the canonical example, since you have the space

documentation/markup.rst
```diff
@@ -10,6 +10,30 @@ This document describes the custom reStructuredText markup introduced by Sphinx
...
+roles w/ custom text  ``:role:`custom text <target>```            :ref:`roles`
+roles w/o link        ``:role:`!target```                         :ref:`roles`
+roles w/o first part  ``:role:`~hidden.hidden.visible```          :ref:`roles`
```
CAM-Gerlach, line 32:
```suggestion
roles w/only last part  ``:role:`~hidden.hidden.visible```          :ref:`roles`
```

Be more specific/correct, since you'll have the space here if you make the above change

documentation/markup.rst
```diff
@@ -229,13 +258,18 @@ body at the bottom of the document after a "Footnotes" rubric heading, like so::
...
 
+.. _comments:
 
 Comments
 --------
 
 Every explicit markup block which isn't a valid markup construct (like the
 footnotes above) is regarded as a comment.
 
+Comments are generally preceeded by ``..``::
```
CAM-Gerlach, line 269:
```suggestion
Every explicit markup block (starting with ``.. ``)
which isn't a :ref:`valid markup construct <directives>`
is regarded as a comment::
```

* Add the mention of "starting with `..`" up here instead
* Avoid double paren and relying on the specific structure of the text
* Link the directive entry explaining the valid markup constructs (and explicit markup blocks)
* Simply and avoid redundancy
* Fix typo ("preceeded")

Since `..` _is_ the start of an explicit markup block as stated in the previous line, comments *must* be therefore be preceded by such, not just most of the time. Since this is therefore mostly redundant, I just combined the two lines

documentation/markup.rst
```diff
@@ -636,9 +670,10 @@ The file name is relative to the current file's path.  Documentation-specific
...
 .. _rest-inline-markup:
+.. _roles:
 
-Inline markup
--------------
+Roles
+-----
 
 As said before, Sphinx uses interpreted text roles to insert semantic markup in
 documents.
```
CAM-Gerlach, line 679: Don't forget to update the line below this, as discussed.

CAM-Gerlach (commented):
Looks good overall with your changes :+1: , I just had some textual improvements, minor nits and bikeshedding

ezio-melotti:
This is the latest iteration of the table:
image

I also had to use a couple of rst tricks for `.. ` and ` ``...`` `, but I was able to get it to work using `` :literal:`...` ``:
image
image

documentation/markup.rst
```diff
@@ -678,8 +676,12 @@ Roles
...
 
-Names of local variables, such as function/method arguments, are an exception,
-they should be marked simply with ``*var*``.
+There are a couple of cases where roles can be omitted in favor of
+simpler markup:
+
+* Function and method arguments can be marked with ``*arg*``.
+* ``True``, ``False``, ``None``, and variables can be marked with
+  :literal:`\`\`...```.
```
CAM-Gerlach, line 684:
```suggestion

* Function and method arguments should be written as ``*arg*``
  (rendering as *arg*)
* ``True``, ``False``, and ``None`` should be written with :literal:`\`\`...```
  (rather than ``:const:``
```

* Show rendered form for *arg*
* Use clearer and more idiomatic wording/phrasing
* Make this at least a SHOULD, not just a MAY
* Explicitly specify the alternative that is being discouraged here
* Fix grammar

Also, after a lot of thought, I suggest eliding the line above specifically mentioning how arguments in the CPython documentation are written, to separate the general reST material in the previous top-level section (with a few CPython-specific bits that can be factored out) from the guidance specific to CPython, and avoiding duplication of the same information

documentation/markup.rst
```diff
@@ -678,8 +676,12 @@ Roles
...
 
 For all other roles, you have to write ``:rolename:`content```.
```
CAM-Gerlach, line 686:
```suggestion
```

I've moved the meaningful content of this (the syntax example) up above where it fits better, and I suggest just eliding this; it duplicates the previous linked section that discusses the general reST/Sphinx syntax, whereas this one describes the CPython-specific semantics, and its not "all" other roles (there are others mentioned in the aforementioned section).

CAM-Gerlach (commented):
I suggested some re-writes to the roles section to avoid parts that are contradictory, duplicative and potentially confusing. Otherwise, this looks pretty good overall.

documentation/markup.rst
```diff
@@ -678,8 +676,12 @@ Roles
 As said before, Sphinx uses interpreted text roles to insert semantic markup in
 documents.
 
-Names of local variables, such as function/method arguments, are an exception,
-they should be marked simply with ``*var*``.
+There are a couple of cases where roles can be omitted in favor of
+simpler markup:
```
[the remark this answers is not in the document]
ezio-melotti: > In addition, there are a few places in the CPython documentation where simpler markup is used:

To me this sounds like "a simpler markup is used in some part of the CPython docs", whereas in theory these exceptions are/should be applied to all the CPython docs.  It also sounds like a neutral statement, whereas I wanted to tell the reader that they should do this too (the following two sentences are more explicit about this).  I agree that we should point out that this is a CPython-specific convention.

> avoid [...] contradicting the following statement implying these are roles (as they are)

Did you mean "aren't roles"?  `*...*` is not a role afaik, wheareas `` `...` `` technically is, even though from the context it can  be inferred that this is talking about `` :role:`...` `` roles.

Maybe something like `In the CPython documentation, simpler markup is conventionally used in a few common cases:` is better?

documentation/markup.rst
ezio-melotti: > * Show rendered form for arg
> * Use clearer and more idiomatic wording/phrasing

ok.

> * Make this at least a SHOULD, not just a MAY

Agreed :+1: 

> * Explicitly specify the alternative that is being discouraged here

I would leave this out, both because there isn't a clear alternative (some people use `:const:`, some `:keyword:`) and because mentioning doesn't seem useful.

> separate the general reST material in the previous top-level section (with a few CPython-specific bits that can be factored out) from the guidance specific to CPython

Do you think this is a separate issue, or should be tackled in this PR?

documentation/markup.rst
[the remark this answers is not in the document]
CAM-Gerlach: > To me this sounds like "a simpler markup is used in some part of the CPython docs", whereas in theory these exceptions are/should be applied to all the CPython docs. It also sounds like a neutral statement, whereas I wanted to tell the reader that they should do this too (the following two sentences are more explicit about this). 

Good points. Here's a revision taking both of them and your suggestion into account:

```suggestion
In the CPython documentation, there are a few common cases
where simpler markup should be used:
```

> Did you mean "aren't roles"? `*...*` is not a role afaik, wheareas `` `...` `` technically is, even though from the context it can be inferred that this is talking about `` :role:`...` `` roles.

They are indeed roles; I didn't initially think they were either until I checked [the relevant reST docs](https://docutils.sourceforge.io/docs/ref/rst/roles.html#emphasis) as part of writing the above, and both `:emphasis:` and `:literal:` are indeed roles, with `*..*` and ``` ``...`` ``` as equivalent forms (just like `` `...` `` is for the default `title-reference` role).

documentation/markup.rst
```diff
@@ -42,7 +67,9 @@ The standard reST inline markup is quite simple: use
 
 * one asterisk: ``*text*`` for emphasis (italics),
 * two asterisks: ``**text**`` for strong emphasis (boldface), and
-* backquotes: ````text```` for code samples.
+* backquotes: ````text```` for code samples, variables, and literals.
+
+Italic is also used for function and method arguments (``*arg*``).
```
CAM-Gerlach, line 72:
```suggestion
```

I suggest re-eliding this from here, to avoid duplication and conflating general reST/Sphinx markdown with CPython-specific conventions

documentation/markup.rst
CAM-Gerlach: > I would leave this out, both because there isn't a clear alternative (some people use :const:, some :keyword:) and because mentioning doesn't seem useful.

Okay, sure. I updated my above suggestion accordingly.

> Do you think this is a separate issue, or should be tackled in this PR?

Separate, since its a substantially bigger change that should be discussed first and IMO out of scope here (the sentence in particular was added with this PR, I just wasn't viewing the right commit to be able to change it). I brought it up to put that suggestion into context with my big-picture view on this.

ezio-melotti: Any reason for omitting variables from the list of elements that should use ` ``...`` `?

CAM-Gerlach: Separation of concerns; they are already listed above in the general Markdown inline syntax section, along with a number of other similar elements that should use ``...``, so it doesn't make sense to repeat them here without repeating everything. This section specifically discusses the additional specific markup usage particular to CPython, rather than being a general list of places ``...`` should be used

documentation/markup.rst
```diff
@@ -636,17 +666,20 @@ The file name is relative to the current file's path.  Documentation-specific
...
+.. _roles:
 
-Inline markup
--------------
+Roles
+-----
 
-As said before, Sphinx uses interpreted text roles to insert semantic markup in
-documents.
+As :ref:`previously mentioned <inline-markup>`,  Sphinx uses
```
CAM-Gerlach, line 674:
```suggestion
As :ref:`previously mentioned <inline-markup>`, Sphinx uses
```

An extra space was inserted when de-logicalizing the line breaks

documentation/markup.rst
```diff
@@ -636,17 +666,20 @@ The file name is relative to the current file's path.  Documentation-specific
...
+interpreted text roles of the form ``:rolename:`content```
+to insert semantic markup in documents.
 
-Names of local variables, such as function/method arguments, are an exception,
-they should be marked simply with ``*var*``.
+In the CPython documentation, there are a few common cases
```
CAM-Gerlach, line 678:
```suggestion
In the CPython documentation, there are a couple common cases
```

Since there are only two listed, I should have suggested this instead

CAM-Gerlach (approved):
Two tiny tweaks, otherwise LGTM :+1:

ezio-melotti:
Merged. thanks for the review and suggestions!

PR #925, opened 2022-07-30
Add Erlend Aasland as sqlite3 expert

erlend-aasland:


core-developers/experts.rst
```diff
@@ -209,7 +209,7 @@ sndhdr
 socket                gpshead
 socketserver
 spwd
-sqlite3               ghaering^
+sqlite3               ghaering^, erlend-aasland*
```
erlend-aasland, line 212: Perhaps we should remove Gerhard. AFAIK, he isn't even maintaining the original pysqlite repo anymore.

Issue #821, opened 2022-03-25
Update label names in the triaging section


ezio-melotti:
The labels used in the `python/cpython` repo have been updated In python/core-workflow#432 and python/bedevere#419.

https://devguide.python.org/triaging/ should be updated accordingly.

https://discuss.python.org/t/github-issues-migration-label-mapping/14212 has more information about the labels.

pgadige:
@Mariatta I'm working on this one. May I share a draft PR?

Mariatta:
Sure @pgadige go for it.

ezio-melotti:
@pgadige, are you still working on this?

pgadige:
@ezio-melotti Yes, I'm working on it. I was MIA for a few days due to a personal urgency.

ezio-melotti:
Now that the migration is done, labels are documented in three places:
* https://devguide.python.org/gh-labels/
* https://devguide.python.org/triaging/
* https://github.com/python/cpython/labels

We are also still updating labels and their names as we adapt to the new workflow, which means that whenever we add/remove/update a label, we also have to update the devguide.

It might be easier to just link to the labels page on `python/cpython`, however not all labels there have descriptions.  This is intentional since some of them are self-explanatory and adding descriptions clutters the "labels" dropdown.

Maybe the best approach would be to document groups of labels and labels that trigger bots or have special meanings and then link to https://github.com/python/cpython/labels for the full list.

za:
Hi @pgadige allow me to initiate the PR. @ezio-melotti I've drafted a PR. Am I in the right direction? I haven't filled the description though.

za:
I've added the description.

za:
I've added the description.

AA-Turner:
Labels are now consolidated into one place in the devguide (https://devguide.python.org/triage/labels/), but I just cut-and-pasted, so there is rampant duplication and likely a lot of the content is outdated (but no less outdated than previously).

A

PR #930, opened 2022-08-06
Overhaul the labels page to reflect current GH labeling practice

ezio-melotti:
This PR updates the `triage/labels.rst` page, that contained outdated references to removed/renamed labels and previous fields used by bpo.

The PR is still a draft.  The first commit documents the existing labels.  Next, most of the remaining text in the page will either be deleted or moved to other sections, since it mostly covers bpo fields.  Finally, I might add more links to the relevant sections to avoid duplicating information.

This is part of https://github.com/orgs/python/projects/25 and will fix the following issues:

Fixes #296 
Fixes #518
Fixes #821
Fixes #868
Fixes #929

The structure and grouping of the labels is also based on python/core-workflow#450

PR #932, opened 2022-08-08
Add a Sphinx role to link to GitHub labels

hugovk:
This adds a `:gh-label:` role to link to a GitHub label on https://github.com/python/cpython.

For example:

* ``` `:gh-label:`docs` ``` -> `https://github.com/python/cpython/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3Adocs`

* ``` :gh-label:`skip news` ``` -> `https://github.com/python/cpython/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3A%22skip+news%22`

This makes it easy to click through to see recent examples of the actual label in use.

It also helps https://github.com/python/devguide/issues/821 because we can click through and find which labels have since been deleted (e.g. `invalid`) or renamed (`library`).

### DIFF
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -10,10 +10,7 @@ from docutils import nodes
 def setup(app):
     app.add_role(
         "gh-label",
-        autolink(
-            "https://github.com/python/cpython/issues"
-            "?q=is%%3Aopen+sort%%3Aupdated-desc+label%%3A%s"
-        ),
+        autolink("https://github.com/python/cpython/labels/%s"),
     )
     # Parallel safety:
     # https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata
@@ -23,7 +20,7 @@ def setup(app):
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
         if " " in text:
-            url_text = urllib.parse.quote_plus(f'"{text}"')
+            url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
         url = pattern % (url_text,)
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -1,0 +1,33 @@
+"""Sphinx extension to add custom roles.
+
+Based on https://protips.readthedocs.io/link-roles.html
+"""
+import urllib.parse
+
+from docutils import nodes
+
+
+def setup(app):
+    app.add_role(
+        "gh-label",
+        autolink(
+            "https://github.com/python/cpython/issues"
+            "?q=is%%3Aopen+sort%%3Aupdated-desc+label%%3A%s"
+        ),
+    )
+    # Parallel safety:
+    # https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata
+    return {"parallel_read_safe": True, "parallel_write_safe": True}
+
+
+def autolink(pattern):
+    def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+        if " " in text:
+            url_text = urllib.parse.quote_plus(f'"{text}"')
+        else:
+            url_text = text
+        url = pattern % (url_text,)
+        node = nodes.reference(rawtext, text, refuri=url, **options)
+        return [node], []
+
+    return role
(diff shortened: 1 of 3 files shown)


erlend-aasland (approved):
Neat!

ezio-melotti:
> This makes it easy to click through to see recent examples of the actual label in use.

This might also be helpful in a couple of other places in the devguide that link e.g. to doc or easy issues, but I wonder if it's a bit overkill to create a role just for it 🤔 

> It also helps #821 because we can click through and find which labels have since been deleted (e.g. invalid) or renamed (library).

This will only work by clicking through the links manually, right?  Even invalid labels will return a valid search page with no results, so we won't be able to detect them automatically.  Also, FWIW, I'm fixing that section in #930.

hugovk:
> This might also be helpful in a couple of other places in the devguide that link e.g. to doc or easy issues, but I wonder if it's a bit overkill to create a role just for it 🤔

My thinking behind creating a role was that we can define all the params stuff in a single place, and it'll then also be consistent. Using GitHub's defaults here:

`https://github.com/python/cpython/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3Afoo`

It also deals with escaping and quoting labels that contains spaces.


---

> This will only work by clicking through the links manually, right? 

Yep.

> Even invalid labels will return a valid search page with no results, so we won't be able to detect them automatically. 

Correct, just for manual checking.

> Also, FWIW, I'm fixing that section in #930.

Great, I knew you were working on it somewhere, thank you!

terryjreedy:
Any particular reason to sort by last update rather than the default of opening time?  I don't know if it makes any difference on the load on GH.  `docs` takes several seconds to return about 1200.  Seems like overkill for examples.

hugovk:
I used the default options GitHub gave me after clicking one of the labels in the issue list, at https://github.com/python/devguide/issues

Here's timings in a fresh Incognito window with local cache disabled, 3 loads for each.

The URLs used in this PR:

* `https://github.com/python/cpython/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3Adocs`
* 5.81s
* 6.25s
* 6.55s
* **Average: 6.20s**

image


The same but removing the `sort%3Aupdated-desc` parameter:

* `https://github.com/python/cpython/issues?q=is%3Aopen+label%3Adocs`
* Resolved to `https://github.com/python/cpython/labels/docs` in the URL bar
* 4.71s
* 8.03s
* 5.98s
* **Average: 6.24s**

image

About the same?

But perhaps using the `https://github.com/python/cpython/labels/docs` form is the neatest?

erlend-aasland:
> But perhaps using the `https://github.com/python/cpython/labels/docs` form is the neatest?

Definitely; I'd go for that.

hugovk:
Updated!

ezio-melotti:
> This might also be helpful in a couple of other places in the devguide that link e.g. to doc or easy issues

I searched through the devguide and I can confirm these are the only two places where they are currently used, but they link to issues only so we might not be able to use the new role if it links to [https://github.com/python/cpython/labels/{LABEL-NAME}](https://github.com/python/cpython/labels/%7BLABEL-NAME%7D):
https://github.com/python/devguide/blob/e807042f9b513092840d9acf1d8a0ffb7c9f3e05/documentation/help-documenting.rst#L66
https://github.com/python/devguide/blob/e807042f9b513092840d9acf1d8a0ffb7c9f3e05/getting-started/fixing-issues.rst#L25

In #930 I can add links to all the labels (the currently proposed link would work fine there), and I was also planning to add a page with several useful search queries (there it won't work).  In addition in #930 I have another link that won't be able to use the role: <https://github.com/python/cpython/labels?q=expert>.

IOW, it looks like this role will be only useful in the page that lists the label, unless we make it more flexible.

hugovk:
Yep, that's right.

Originally this PR used ugly links like `https://github.com/python/cpython/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3Adocs`, but now has nicer ones like `https://github.com/python/cpython/labels/docs`

So it's less useful than before, but I think it's more readable to see `:gh-label:`docs` than a long link.

But I'm happy to close this, or replace with normal links, or wait for normal links in #930 :)

PR #936, opened 2022-08-17
Document how to add C API tests

encukou:
Here are some developer docs for C API tests (https://github.com/python/cpython/issues/93649 and https://github.com/python/cpython/issues/95991).

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -113,9 +113,9 @@ To add a new set of tests (or extract a set out of the monolithic
 - The file should define a module as usual, except:
 
   - Instead of ``<Python.h>``, include ``"parts.h"``.
-  - Instead of ``PyInit_modname``, define ``_PyTestCapi_Init_yourfeature``
-    that *takes* the ``_testcapi`` module and adds functions/classes to it.
-    (Use ``PyModule_AddFunctions`` to add functions).
+  - Instead of ``PyInit_modname``, define a ``_PyTestCapi_Init_yourfeature``
+    function that *takes* the ``_testcapi`` module and adds functions/classes
+    to it. (You can use ``PyModule_AddFunctions`` to add functions.)
 
 - Add the ``_PyTestCapi_Init_*`` function to ``Modules/_testcapi/parts.h``
 
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -108,7 +108,7 @@ files.
 To add a new set of tests (or extract a set out of the monolithic
 :file:`Modules/_testcapimodule.c`):
 
-- Create a C named ``Modules/_testcapi/yourfeature.c``
+- Create a C file named ``Modules/_testcapi/yourfeature.c``
 
 - The file should define a module as usual, except:
 
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -62,6 +62,13 @@ definition in ``Include/internal/`` (or directly in a ``.c`` file):
   a macro aliasing the old name to the new one.
   See :pep:`"Finalizing the API" in PEP 590 <590#finalizing-the-api>` for an example.
 
+Internal API Tests
+------------------
+
+C tests for the internal C API live in ``Modules/_testinternalcapi.c``.
+Functions named ``test_*`` are used as tests directly.
+Python parts of the tests live in various places in ``Lib/test``.
+
 
 .. _public-capi:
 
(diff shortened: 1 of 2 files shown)


developer-workflow/c-api.rst
```diff
@@ -88,6 +95,40 @@ Guidelines for expanding/changing the public API:
...
+:file:`Modules/_testcapimodule.c`):
+
+- Create a C named ``Modules/_testcapi/yourfeature.c``
+
+- The file should define a module as usual, except:
+
+  - Instead of ``<Python.h>``, include ``"parts.h"``.
+  - Instead of ``PyInit_modname``, define ``_PyTestCapi_Init_yourfeature``
+    that *takes* the ``_testcapi`` module and adds functions/classes to it.
+    (Use ``PyModule_AddFunctions`` to add functions).
```
ezio-melotti, line 118:
```suggestion
  - Instead of ``PyInit_modname``, define a ``_PyTestCapi_Init_yourfeature``
    that *takes* the ``_testcapi`` module and adds functions/classes to it
    (use ``PyModule_AddFunctions`` to add functions).
```

developer-workflow/c-api.rst
```diff
@@ -88,6 +95,40 @@ Guidelines for expanding/changing the public API:
...
+
+- Add the ``_PyTestCapi_Init_*`` function to ``Modules/_testcapi/parts.h``
+
+- Call the ``_PyTestCapi_Init_*`` from ``PyInit__testcapi`` in
```
ezio-melotti, line 122: Is this a typo or is it correct?
```suggestion
- Call the ``_PyTestCapi_Init_*`` from ``PyInit_testcapi`` in
```

ezio-melotti (approved):

developer-workflow/c-api.rst
encukou: Thanks!
I'd rather keep the note as a separate sentence, but even then the punctuation needs a fix.

encukou:
```suggestion
  - Instead of ``PyInit_modname``, define a ``_PyTestCapi_Init_yourfeature``
    function that *takes* the ``_testcapi`` module and adds functions/classes
    to it. (You can use ``PyModule_AddFunctions`` to add functions.)
```

developer-workflow/c-api.rst
encukou: You have a sharp eye!
But it's [correc](https://github.com/python/cpython/blob/586fc02be5b3e103bfddd49654034a898a8d6dfc/Modules/_testcapimodule.c#L6426)t: [`PyInit_`](https://docs.python.org/3/extending/building.html#c.PyInit_modulename) is prepended to the module name, which itself starts with an underscore.

testing/run-write-tests.rst
```diff
@@ -134,6 +134,8 @@ When you are adding tests to an existing test file, it is also recommended
 that you study the other tests in that file; it will teach you which precautions
 you have to take to make your tests robust and portable.
 
+For tests of the C API, see Tests sections in :ref:`c-api`.
```
erlend-aasland, line 137: Add an anchor for the Tests section, and add a `:ref:` to it here.

encukou: I would, but there are 3 of them. IMO in this case it's better for readers to find the one they want by name in the TOC.

ezio-melotti: Agreed.  I was going to suggest the same, but as @encukou said there are 3 sections and they are all easily discoverable in the `c-api` page.

developer-workflow/c-api.rst
```diff
@@ -88,6 +95,40 @@ Guidelines for expanding/changing the public API:
   fields, arguments and return values are well defined.
 
 
+C API Tests
+-----------
+
+Tests for the public C API live in the ``_testcapi`` module.
+Functions named ``test_*`` are used as tests directly.
+Tests that need Python code (or are just easier to partially write in Python)
+live in ``Lib/test``, mainly in :file:`Lib/test/test_capi.py`.
```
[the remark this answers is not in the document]
encukou: No, it doesn't :(
I get `ERROR: Unknown interpreted text role "source"`. I also get no hits for `:source:` in the Devguide repo.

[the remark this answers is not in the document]
erlend-aasland: Ah, too bad.

erlend-aasland (approved):

encukou:
Thank you for the reviews!

PR #943, opened 2022-08-28
Prefer upstream for main

jaraco:

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -60,7 +60,7 @@ The CPython repo is hosted on GitHub. To get a copy of the source code you shoul
 :ref:`fork the Python repository on GitHub <fork-cpython>`, :ref:`create a local
 clone of your personal fork, and configure the remotes <clone-your-fork>`.
 
-You will only need to execute these steps once:
+You will only need to execute these steps once per machine:
 
 1. Go to https://github.com/python/cpython.
 
@@ -85,13 +85,18 @@ You will only need to execute these steps once:
 
       $ git config --local branch.main.remote upstream
 
-8. Verify that your setup is correct::
+8. Since one should never attempt to push to ``upstream``, configure
+   ``git`` to push always to ``origin``::
+
+      $ git remote set-url --push upstream git@github.com:<your-username>/cpython.git
+
+9. Verify that your setup is correct::
 
       $ git remote -v
       origin  git@github.com:<your-username>/cpython.git (fetch)
       origin  git@github.com:<your-username>/cpython.git (push)
       upstream        git@github.com:python/cpython.git (fetch)
-      upstream        git@github.com:python/cpython.git (push)
+      upstream        git@github.com:<your-username>/cpython.git (push)
       $ git config branch.main.remote
       upstream
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -79,7 +79,7 @@ You will only need to execute these steps once per machine:
 6. Configure an ``upstream`` remote::
 
       $ cd cpython
-      $ git remote add upstream git@github.com:python/cpython.git
+      $ git remote add upstream https://github.com/python/cpython.git
 
 7. Configure ``git`` to pull ``main`` from the ``upstream`` remote::
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -79,7 +79,7 @@ You will only need to execute these steps once per machine:
 6. Configure an ``upstream`` remote::
 
       $ cd cpython
-      $ git remote add upstream https://github.com/python/cpython.git
+      $ git remote add upstream https://github.com/python/cpython
 
 7. Configure ``git`` to pull ``main`` from the ``upstream`` remote::
 
@@ -95,7 +95,7 @@ You will only need to execute these steps once per machine:
       $ git remote -v
       origin  git@github.com:<your-username>/cpython.git (fetch)
       origin  git@github.com:<your-username>/cpython.git (push)
-      upstream        git@github.com:python/cpython.git (fetch)
+      upstream        https://github.com/python/cpython (fetch)
       upstream        git@github.com:<your-username>/cpython.git (push)
       $ git config branch.main.remote
       upstream
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -81,7 +81,7 @@ You will only need to execute these steps once:
       $ cd cpython
       $ git remote add upstream git@github.com:python/cpython.git
 
-7. Configure ``git`` to pull main from the ``upstream`` remote::
+7. Configure ``git`` to pull ``main`` from the ``upstream`` remote::
 
       $ git config --local branch.main.remote upstream
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -81,13 +81,19 @@ You will only need to execute these steps once:
       $ cd cpython
       $ git remote add upstream git@github.com:python/cpython.git
 
-7. Verify that your setup is correct::
+7. Configure ``git`` to pull main from the ``upstream`` remote::
+
+      $ git config --local branch.main.remote upstream
+
+8. Verify that your setup is correct::
 
       $ git remote -v
       origin  git@github.com:<your-username>/cpython.git (fetch)
       origin  git@github.com:<your-username>/cpython.git (push)
       upstream        git@github.com:python/cpython.git (fetch)
       upstream        git@github.com:python/cpython.git (push)
+      $ git config branch.main.remote
+      upstream
 
 If you did everything correctly, you should now have a copy of the code
 in the ``cpython`` directory and two remotes that refer to your own GitHub fork
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -54,7 +54,7 @@ You will only need to do this once per machine.  From your command line::
 It is also recommended to configure an ``upstream`` remote repository::
 
    cd cpython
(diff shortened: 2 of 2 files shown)


getting-started/setup-building.rst
```diff
@@ -81,13 +81,19 @@ You will only need to execute these steps once:
       $ cd cpython
       $ git remote add upstream git@github.com:python/cpython.git
 
-7. Verify that your setup is correct::
+7. Configure ``git`` to pull main from the ``upstream`` remote::
+
+      $ git config --local branch.main.remote upstream
```
ezio-melotti, line 86: So by doing this `git pull` without extra args will pull from `upstream` instead of `origin`?
```suggestion
7. Configure ``git`` to pull ``main`` from the ``upstream`` remote::

      $ git config --local branch.main.remote upstream
```

getting-started/setup-building.rst
```diff
@@ -81,13 +81,19 @@ You will only need to execute these steps once:
...
+
+8. Verify that your setup is correct::
 
       $ git remote -v
       origin  git@github.com:<your-username>/cpython.git (fetch)
       origin  git@github.com:<your-username>/cpython.git (push)
       upstream        git@github.com:python/cpython.git (fetch)
       upstream        git@github.com:python/cpython.git (push)
+      $ git config branch.main.remote
+      upstream
```
ezio-melotti, line 96: This could be added to the Git bootcamp too:
https://devguide.python.org/getting-started/git-boot-camp/index.html#cloning-a-forked-cpython-repository

hugovk:
Would this mean pushing to `main` would attempt to push to `upstream/main`?

Usually there's no need to push to `main`, except when syncing `upstream/main` to `origin/main`, which I think this change would make redundant? 

But some people may have muscle memory, because it's a very common thing to do with forks (see all these [upvotes](https://stackoverflow.com/q/7244321/724176), and ["Syncing a fork"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) is quite high in GitHub's ["Collaborating with pull requests" docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests)).

Another rare but occasional use case for pushing to main: wanting to test your own changes on CI when the CI is locked to only run for `main` and not for feature branches.

erlend-aasland:
> Would this mean pushing to `main` would attempt to push to `upstream/main`?

The devguide should be updated to show how to configure upstream push as an invalid operation. For example, by supplying an invalid URL for such an operation.

getting-started/setup-building.rst
jaraco: Correct. This configuration is also what happens when one runs `gh repo fork --remote`.

jaraco:
> Would this mean pushing to `main` would attempt to push to `upstream/main`?

Yes, it seems that's the case :(

```
 draft $ git clone gh://python/cpython
Cloning into 'cpython'...
remote: Enumerating objects: 911455, done.
remote: Counting objects: 100% (245/245), done.
remote: Compressing objects: 100% (163/163), done.
remote: Total 911455 (delta 145), reused 143 (delta 82), pack-reused 911210
Receiving objects: 100% (911455/911455), 497.39 MiB | 5.25 MiB/s, done.
Resolving deltas: 100% (722615/722615), done.
 draft $ cd cpython
 cpython main $ git remote -v
origin  https://github.com/python/cpython (fetch)
origin  https://github.com/python/cpython (push)
 cpython main $ gh repo fork --remote
! jaraco/cpython already exists
✓ Added remote origin
 cpython main $ git remote -v
origin  https://github.com/jaraco/cpython.git (fetch)
origin  https://github.com/jaraco/cpython.git (push)
upstream        https://github.com/python/cpython (fetch)
upstream        https://github.com/python/cpython (push)
 cpython main $ git config --local -l
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true
core.ignorecase=true
core.precomposeunicode=true
remote.upstream.url=gh://python/cpython
remote.upstream.gh-resolved=base
remote.upstream.fetch=+refs/heads/*:refs/remotes/upstream/*
branch.main.remote=upstream
branch.main.merge=refs/heads/main
remote.origin.url=https://github.com/jaraco/cpython.git
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
 cpython main $ touch foo
 cpython main $ git add foo
 cpython main $ git commit -a -m "add foo"
[main 4838818fec] add foo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
 cpython main $ git push
Enumerating objects: 4, done.
Counting objects: 100% (4/4), done.
Delta compression using up to 8 threads
Compressing objects: 100% (2/2), done.
Writing objects: 100% (3/3), 925 bytes | 71.00 KiB/s, done.
Total 3 (delta 1), reused 1 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (1/1), completed with 1 local object.
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: error: 7 of 7 required status checks are expected.
To https://github.com/python/cpython
 ! [remote rejected]       main -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/python/cpython'
```

It looks like maybe the upstream 'push' can be configured to also point to the user's fork:

```
 cpython main $ git remote set-url --push upstream gh://jaraco/cpython
 cpython main $ git remote -v
origin  https://github.com/jaraco/cpython.git (fetch)
origin  https://github.com/jaraco/cpython.git (push)
upstream        https://github.com/python/cpython (fetch)
upstream        https://github.com/jaraco/cpython (push)
 cpython main $ git push
Enumerating objects: 4, done.
Counting objects: 100% (4/4), done.
Delta compression using up to 8 threads
Compressing objects: 100% (2/2), done.
Writing objects: 100% (3/3), 925 bytes | 102.00 KiB/s, done.
Total 3 (delta 1), reused 1 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (1/1), completed with 1 local object.
To https://github.com/jaraco/cpython
   e0d54a4a79..4838818fec  main -> main
```

That configuration should also address @erlend-aasland's suggestion. Instead of making the push invalid, it just directs it to the user's repo. That also means that a `git pull; git push` in main will sync main to the fork.

I'll update the docs to reflect this approach.

getting-started/setup-building.rst
jaraco: Probably we should figure out a way to consolidate these two sets of instructions. Describe the steps in one place and link to them from another. My instinct is to provide the instructions in the git boot camp and link them from here. I'll probably do that in a prerequisite change prior to this one to avoid having to keep the two sets of instructions in sync. How does that sound?

jaraco:
> I'll update the docs to reflect this approach.

This is done in 3fdba64.

getting-started/setup-building.rst
ezio-melotti: Having all the commands here is convenient, since people doing a setup don't have to follow links back and forth from this page to the GitHub bootcamp page.  On the other hand, the bootcamp entries are also convenient on their own [^1].

I think the best approach would be to:
* list all upstream-related commands in a single step and code block[^2], with short inline comments (about what the commands do)
* have the same commands in the bootcamp, but with more in-depth discussions (e.g. why they are necessary)
* from here, link to the bootcamp so that people that want more details can find them there
* add a couple of comments about the duplicated commands as a remainder for people updating either page

[^1]: I regular revisit the entry about setting up a remote, since I do it for every new repo I clone, but not often enough that I remember the exact synta
[^2]: this includes steps 6-9.  Step 5 can be separated.

jaraco: I'll work on this today.

jaraco:
FYI, Since I last visited this issue, I've implemented [jaraco.develop.git.configure_fork](https://github.com/jaraco/jaraco.develop/blob/5a023776582cccf3f8a633672f64d324b154bacd/jaraco/develop/git.py#L150-L160), which I use to checkout cpython (and other projects to which I contribute through forks) so that I don't have to remember these steps or refer to any documentation and I get consistent behavior across [hundreds of projects](https://github.com/jaraco/jaraco.develop/blob/main/jaraco/develop/projects.txt).

getting-started/setup-building.rst
jaraco: Done in 769a334. I've opted not to implement the "best approach" above for now, though I agree that would be a nice improvement.

getting-started/setup-building.rst
```diff
@@ -81,13 +81,24 @@ You will only need to execute these steps once:
       $ cd cpython
       $ git remote add upstream git@github.com:python/cpython.git
```
CAM-Gerlach, line 82: For public repos like this, I always set the pull request URL to the HTTPS one (`https://github.com/python/cpython.git`) since it doesn't require authentication and may be slightly less performant. It does normally require an extra step to add the URL, but in this case since the `push` URL is being overridden anyway, I figure we may as well just use it here instead, which will add an additional protection against accidentally pushing to `main` in case users don't do the next step.

```suggestion
      $ git remote add upstream https://github.com/python/cpython.git
```

jaraco: I'm a little wary of adding another bit of variance here. If we change it here, it needs to change in other steps and the copy of these steps in the bootcamp. Also, changing it here implies that we could (should?) also change it to https for fetch for `origin` as well (for symmetry). I'll take the minimum step from what you're proposing.

CAM-Gerlach (approved):
LGTM, thanks @jaraco

ezio-melotti (approved):

merwok:
A reminder that the merge commit message is pre-filled with all commit messages and should be edited to be brief and useful.

PR #882, opened 2022-06-06
gh-861 Document GitHub CLI on gitbootcamp.rst

za:

Issue #861, opened 2022-05-10
Document GitHub CLI


ezio-melotti:
https://cli.github.com/ should be documented in https://devguide.python.org/gitbootcamp/, on its own FAQ and/or integrated with the other FAQs.

Useful commands (shared on Discord):
* `gh auth login` (once)
* `gh pr create`: to create a PR
* `gh pr checkout`: to checkout others' PRs
* `gh config set git_protocol ssh` (once): to be able to push to others' PRs
* `gh config set browser "/Applications/Firefox.app/Contents/MacOS/firefox"`: to set the browser

erlend-aasland:
Note, there is already a short mention of GitHub CLI: https://devguide.python.org/gitbootcamp/#downloading-other-s-patches

PR #946, opened 2022-09-05
Recommend GitHub CLI for easier bootstrapping

jaraco:
As I was working on #943, I realized it might be easier to direct users to use the GitHub CLI instead of all of the manual steps.

Thoughts?

- Consider directing users to the GitHub CLI for easier bootstrapping.

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -25,8 +25,8 @@ and are provided for development and testing purposes only.
 
 .. _vcsetup:
 
-Install ``git``
-===============
+Install ``git`` and ``gh``
+==========================
 
 CPython is developed using `git <https://git-scm.com>`_ for version control. The git
 command line program is named ``git``; this is also used to refer to git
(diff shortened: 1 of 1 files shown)


ezio-melotti (commented):
I don't think `gh` should be required.  It's fine documenting it as an alternative setup (here or in the bootcamp), but a regular setup that doesn't use `gh` should still be documented.

See also,:
* #861
* #882

erlend-aasland:
I think promoting `gh` is a good thing. It is a great tool for our workflow.

hugovk:
[Sphinx Inline Tabs](https://sphinx-inline-tabs.readthedocs.io/en/latest/) could be useful to show the two different options.

It's used on https://packaging.python.org/en/latest/tutorials/packaging-projects/ to select between Unix/macOS and Windows (and also Hatchling, setuptools, Flit and PDM):

* https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives
* https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml

It's nice that you when you select one tab, it flips for the whole page:

image

image

Also used for example at https://pip.pypa.io/en/stable/getting-started/

CAM-Gerlach:
Agreed that showing both options, one with a specific one wrapper tool and the more manual one with the "standard" tooling, would be helpful (I personally still use `hub` myself). And +1 for Sphinx inline tabs; with appropriate organization they allow showing alternatives while avoiding making the document longer and more confusing. We (Spyder) really need to adopt this ourselves in all our own docs that offer both `venv/pip` and `conda` environment and installation methods.

@hugovk is there a built-in way to link the tab choices, so that clicking on one will automatically switch all of them, or is that a custom extension on the packaging site? That definitely is a big boon for usability.

hugovk:
@CAM-Gerlach Yes, it's built-in!

> * Synchronisation: Tabs with the same label all switch with a single click.

And see the demo at https://sphinx-inline-tabs.readthedocs.io/en/latest/

ezio-melotti:
@jaraco, do you want to update the PR to use `sphinx-inline-tabs`?

jaraco:
While I can see the value in using sphinx-inline-tabs, I worry that the maintenance burden is getting out of hand, especially considering that the docs exist in two places, so anything that happens probably needs to be duplicated, so duplicating the techniques would be a multiplier on the existing multiple. It's probably best to provide one technique. I wouldn't be opposed to someone adopting this diff to use sphinx-inline-tabs, but I'm not inclined to do it myself.

PR #952, opened 2022-10-05
Update expected release date of 3.11

Zac-HD:
As per https://www.python.org/downloads/release/python-3110rc2/ - I was looking up EOL dates, but it was a bit jarring to see 3.11 listed as already released 😅


ezio-melotti (approved):

ezio-melotti:
This change is fine.  Once it's actually released we can remove the `*...*`.

PR #953, opened 2022-10-06
Use sphinxext-opengraph to generate OpenGraph metadata

hugovk:
Add [Open Graph metadata](https://ogp.me/) using https://github.com/wpilibsuite/sphinxext-opengraph.


conf.py
```diff
@@ -97,3 +98,7 @@
 }
 
 todo_include_todos = True
+
+# sphinxext-opengraph config
+ogp_site_url = "https://devguide.python.org/"
+ogp_image = "https://devguide.python.org/_static/python-logo.svg"
```
ezio-melotti, line 104: Should we add [more variables](https://github.com/wpilibsuite/sphinxext-opengraph#options)?

hugovk: Re: other variables:

I'll check `ogp_custom_meta_tags` for `<meta name="description">`.

https://github.com/python/docs-community/issues/65#issuecomment-1270446743

hugovk: > I'll check `ogp_custom_meta_tags` for `<meta name="description">`.

We can use `ogp_custom_meta_tags` to add it, but it would be hardcoded and we'd like to generate it from each page.

ezio-melotti: > The recommended image size is 1200 x 630, should we use a Python logo or something else?

Since the image is rectangular (and somewhat big), it would be nice if it had both the Python logo and the text "Devguide".  If/when we do the same for `cpython` (or for the `peps` repo), we can have the same logo but different texts, so that the links are easily distinguishable.

This however would require creating new images, so if you don't want to do that, just a simple logo will do (we can always update the image later).

> I don't think we need any of the other variables.

ok

> We can use `ogp_custom_meta_tags` to add it, but it would be hardcoded and we'd like to generate it from each page.

Right, it would be better if the description was created dynamically.

hugovk: SVG replaced with a PNG of the Python logo + "Devguide".

og-image

I wrote a script to create it (https://github.com/hugovk/pixel-tools/pull/15), so we can re-use that for other repos if we want.

I also squashed out the extra bytes from the PNG using `optipng -o7 -zm1-9 _static/og-image.png` (57,118 -> 9,573 bytes, 16.76% decrease).

When temporarily setting `ogp_image = "https://cpython-devguide--953.org.readthedocs.build/_static/og-image.png"` Facebook's sharing debugger shows the following OG metadata/preview: 

image

https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fcpython-devguide--953.org.readthedocs.build%2F

conf.py
```diff
@@ -104,6 +105,14 @@
 
 todo_include_todos = True
 
+# sphinxext-opengraph config
+ogp_site_url = "https://devguide.python.org/"
```
ezio-melotti, line 109:
```suggestion
ogp_site_url = "https://devguide.python.org/"
ogp_site_name = "Python Developer's Guide"
```
I think this would be useful to add:
> `ogp_site_name`:
>     This is not required. Name of the site. This is displayed above the title.

In this example this would be the small "GitHub" above the title.
image

Compare with this (old) preview, that doesn't mention "Devguide" anywhere:
image

conf.py
```diff
@@ -104,6 +105,15 @@
...
+ogp_site_name = "Python Developer's Guide"
+ogp_image = "https://devguide.python.org/_static/og-image.png"
```
ezio-melotti, line 111: The immage is currently broken in the preview build.  This should fix it and should work once deployed too:
```suggestion
ogp_image = "_static/og-image.png"
```

ezio-melotti (commented):
I tested the latest iteration and got this:
image

The website name works, but the image still doesn't (it loads for a while and then :hankey:).
This is because it currently points to https://devguide.python.org/_static/og-image.png, so the suggestion above should fix it.

conf.py
hugovk: Updated.

With relative, it's still broken in preview builds because it's joined to `ogp_site_url = "https://devguide.python.org/"` and we don't yet have the image in `main`:

```html
<meta property="og:image" content="https://devguide.python.org/_static/og-image.png" />
```

But not a big problem, it'll be there soon :)

For testing, we can temporarily point `ogp_site_url` to the RTD preview URL. Feel free to push to this PR if you like.

ezio-melotti:
I saw that in the source of https://cpython-devguide--953.org.readthedocs.build/ there is a `<meta property="og:url" content="https://devguide.python.org/index/" />`, and https://devguide.python.org/index/ is a 404.

With the `dirhtml` builder `sphinxext-opengraph` is adding the name of the current page to the url:
https://github.com/wpilibsuite/sphinxext-opengraph/blob/9f1a8c890ede83e11d24f20e569220873f5f71f8/sphinxext/opengraph/__init__.py#L89-L97

On the main page (i.e. `/`) this ends up appending an extra `index/` resulting in the 404.
We could add a redirect on our end, but I wonder if `index` should be special-cased upstream. :thinking:

hugovk:
This would be better fixed upstream, as it will break for other sites using `dirhtml`. Please could you report?

hugovk:
Reported at https://github.com/wpilibsuite/sphinxext-opengraph/issues/77 and created PR https://github.com/wpilibsuite/sphinxext-opengraph/pull/78.

hugovk:
> Reported at [wpilibsuite/sphinxext-opengraph#77](https://github.com/wpilibsuite/sphinxext-opengraph/issues/77) and created PR [wpilibsuite/sphinxext-opengraph#78](https://github.com/wpilibsuite/sphinxext-opengraph/pull/78).

The `dirhtml` bug is fixed and released in sphinxext-opengraph 0.7.1! 🚀 

Also, to help https://github.com/python/docs-community/issues/65, my PR https://github.com/wpilibsuite/sphinxext-opengraph/pull/71 to use `og:description` as `description` is also part of 0.7.1 and enabled by default.

---

Here's a demo branch and build with the URL fixed so the image URL works (https://github.com/hugovk/devguide/commit/483ee38b0628c52ff981f60651f2709ea021aa74)

* https://hugovk-devguide.readthedocs.io/en/add-og-metadata2/

Here's a page in the Facebook sharing debugger:

* https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fhugovk-devguide.readthedocs.io%2Fen%2Fadd-og-metadata2%2Fversions%2F

image

Here's a Slack share (demo and current site):

image

And Discord:

image

TheTripleV:
Noticed you have a Discord example. Adding a "theme-color" meta tag lets you customize the color of the vertical bar on the left side of the Discord embed.
Example color config: https://github.com/wpilibsuite/frc-docs/blob/main/source/conf.py#L123
Example url with a color: https://docs.wpilib.org/en/stable/index.html

hugovk:
Thanks for the tip! I updated this and the demo branch to use Python blue:

image

conf.py
hugovk: It's taking a while to get the font. It would be nice to get this merged to see how it works when live. Shall we merge this as-is, or with the image removed?

ezio-melotti: Let's merge it.
I'll ping Deb again about the font and we can update that in a follow-up PR.

hugovk:
I found the fonts in https://github.com/python/pythondotorg/tree/main/static/fonts so have regenerated the image using https://github.com/hugovk/pixel-tools/blob/1618b2a09bd5998899958856b7fef4503f95cba2/og_image.py like:

```sh
python og_image.py --logo tests/python-logo.png --size 210 --font ../pythondotorg/static/fonts/FluxRegular.ttf
...
optipng -o7 -zm1-9 _static/og-image.png # 16.28% decrease
```

og-image

# Demo

https://hugovk-devguide.readthedocs.io/en/add-og-metadata2/testing/run-write-tests/

# Preview in Facebook

image

# Preview in Slack

image

conf.py
hugovk: (Continued in main thread: https://github.com/python/devguide/pull/953#issuecomment-1328700892)

conf.py
```diff
@@ -104,6 +105,16 @@
...
 todo_include_todos = True
 
+# sphinxext-opengraph config
+ogp_site_url = "https://devguide.python.org/"
+ogp_site_name = "Python Developer's Guide"
+ogp_image = "_static/og-image.png"
+ogp_custom_meta_tags = [
+    '<meta property="og:image:width" content="1200">',
+    '<meta property="og:image:height" content="630">',
+    '<meta name="theme-color" content="#3776ab" />',
```
ezio-melotti, line 115: FWIW I looked into using a gradient that matches both colors of the Python logo, but this doesn't seem supported.

ezio-melotti (approved):

hugovk:
Google's SEO checks using their PageSpeed Insights tool:

https://pagespeed.web.dev/report?url=https%3A%2F%2Fdevguide.python.org%2F&form_factor=desktop

# Before: 91

image

# After: 100

image

hugovk:
# Facebook sharing preview

https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fdevguide.python.org%2F

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>
	<td>
</table>

PR #954, opened 2022-10-06
FIX: Broken inline literal.

JulienPalard:
It's currently not rendered due to this missing space:

Capture d’écran du 2022-10-06 15-00-46

Yes this will be automatically caught by the next release of sphinxlint.


hugovk (approved):
Thank you!

PR #961, opened 2022-10-07
Add a Sphinx role to link to GitHub files 

### LINKED ISSUE #960: Add links to files in `internals/parser.rst`
[This comment by @CAM-Gerlach](https://github.com/python/devguide/pull/957#discussion_r989467777) suggested to add links to a few files mentioned in `internals/parser.rst`.

Even though direct links could be added, the page already refers to several other files using the `:file:` role, which [doesn't create any link](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-file).

Maybe a new role could solve that.

ezio-melotti:
This PR adds a role to link to GitHub files (in the `python/cpython` repo) and uses them in `internals/parser.rst`.

* Fixes #960

### DIFF
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -11,12 +11,12 @@ def setup(app):
     # role to link to cpython files
     app.add_role(
         "cpy-file",
-        autolink("https://github.com/python/cpython/blob/main/%s"),
+        autolink("https://github.com/python/cpython/blob/main/{}"),
     )
     # role to link to cpython labels
     app.add_role(
         "gh-label",
-        autolink("https://github.com/python/cpython/labels/%s"),
+        autolink("https://github.com/python/cpython/labels/{}"),
     )
     # Parallel safety:
     # https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata
@@ -30,8 +30,8 @@ def autolink(pattern):
             url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
-        url = pattern % (url_text,)
-        node = nodes.literal('', text.lstrip('!'), **options)
+        url = pattern.format(url_text)
+        node = nodes.literal(rawtext, text.lstrip('!'), **options)
         # don't create a reference if the text starts with !
         if not text.startswith('!'):
             node = nodes.reference('', '', node, refuri=url, **options)
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -25,12 +25,16 @@ def setup(app):
 
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+        """Combine literal + reference (unless the text is prefixed by a !)."""
         if " " in text:
             url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
         url = pattern % (url_text,)
-        node = nodes.reference(rawtext, text, refuri=url, **options)
+        node = nodes.literal('', text.lstrip('!'), **options)
+        # don't create a reference if the text starts with !
+        if not text.startswith('!'):
+            node = nodes.reference('', '', node, refuri=url, **options)
         return [node], []
 
     return role
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -8,6 +8,12 @@ from docutils import nodes
 
 
 def setup(app):
+    # role to link to cpython files
+    app.add_role(
+        "cpy-file",
+        autolink("https://github.com/python/cpython/blob/main/%s"),
+    )
+    # role to link to cpython labels
     app.add_role(
         "gh-label",
         autolink("https://github.com/python/cpython/labels/%s"),
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -24,17 +24,19 @@ def setup(app):
 
 
 def autolink(pattern):
-    def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+    def role(name, rawtext, text, lineno, inliner, _options=None, _content=None):
         """Combine literal + reference (unless the text is prefixed by a !)."""
         if " " in text:
-            url_text = urllib.parse.quote(f"{text}")
+            url_text = urllib.parse.quote(text)
         else:
             url_text = text
         url = pattern.format(url_text)
-        node = nodes.literal(rawtext, text.lstrip('!'), **options)
         # don't create a reference if the text starts with !
-        if not text.startswith('!'):
-            node = nodes.reference('', '', node, refuri=url, **options)
+        if text.startswith('!'):
+            node = nodes.literal(rawtext, text[1:])
+        else:
+            node = nodes.reference(rawtext, '', nodes.literal(rawtext, text),
+			           refuri=url, internal=False)
         return [node], []
 
     return role
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -36,7 +36,7 @@ def autolink(pattern):
             node = nodes.literal(rawtext, text[1:])
         else:
             node = nodes.reference(rawtext, '', nodes.literal(rawtext, text),
-			           refuri=url, internal=False)
+                                   refuri=url, internal=False)
         return [node], []
 
     return role
--- a/internals/parser.rst
+++ b/internals/parser.rst
@@ -539,7 +539,7 @@ parser (the one used by the interpreter) just execute: ::
 
     make regen-pegen
 
-using the :file:`Makefile` in the main directory.  If you are on Windows you can
+using the :cpy-file:`!Makefile` in the main directory.  If you are on Windows you can
 use the Visual Studio project files to regenerate the parser or to execute: ::
 
     ./PCbuild/build.bat --regen
--- a/internals/parser.rst
+++ b/internals/parser.rst
@@ -518,17 +518,17 @@ Pegen
 =====
 
 Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. It is the
-program that can be used to read the python grammar located in :file:`Grammar/Python.gram` and produce the final C
+program that can be used to read the python grammar located in :cpy-file:`Grammar/python.gram` and produce the final C
 parser. It contains the following pieces:
 
 * A parser generator that can read a grammar file and produce a PEG parser written in Python or C that can parse
-  said grammar. The generator is located at :file:`Tools/peg_generator/pegen`.
+  said grammar. The generator is located at :cpy-file:`Tools/peg_generator/pegen`.
 * A PEG meta-grammar that automatically generates a Python parser that is used for the parser generator itself
   (this means that there are no manually-written parsers). The meta-grammar is
-  located at :file:`Tools/peg_generator/pegen/metagrammar.gram`.
+  located at :cpy-file:`Tools/peg_generator/pegen/metagrammar.gram`.
 * A generated parser (using the parser generator) that can directly produce C and Python AST objects.
 
-The source code for Pegen lives at :file:`Tools/peg_generator/pegen` but normally all typical commands to interact
+The source code for Pegen lives at :cpy-file:`Tools/peg_generator/pegen` but normally all typical commands to interact
 with the parser generator are executed from the main makefile.
 
 How to regenerate the parser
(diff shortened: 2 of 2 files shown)


_extensions/custom_roles.py
```diff
@@ -8,6 +8,12 @@
 
 
 def setup(app):
+    # role to link to cpython files
+    app.add_role(
+        "cpy-file",
```
ezio-melotti, line 13: Using `gh-file` will make it more symmetric with `gh-label`, but since the two are not used together, I thought using `cpy-file` would make the fact that they are linking to the `cpython` repo more explicit.

internals/parser.rst
```diff
@@ -518,17 +518,17 @@ Pegen
 =====
 
 Pegen is the parser generator used in CPython to produce the final PEG parser used by the interpreter. It is the
-program that can be used to read the python grammar located in :file:`Grammar/Python.gram` and produce the final C
+program that can be used to read the python grammar located in :cpy-file:`Grammar/python.gram` and produce the final C
```
ezio-melotti, line 521: After creating the link, I (manually) found out that it was supposed to be lowercase.

Upon further testing, I found out that these are already integrated with `linkcheck`:
```
(internals/parser: line  520) broken    https://github.com/python/cpython/blob/main/Grammar/Python.gram - 
404 Client Error: Not Found for url: https://github.com/python/cpython/blob/main/Grammar/Python.gram
```

hugovk (approved):
Looks good!

image

One idea, filenames are often written with a fixed-width font (e.g. in this [style guide](https://bishopfox.com/resources/cybersecurity-style-guide)), shall we have the role do that too?

And document this role at https://cpython-devguide--961.org.readthedocs.build/documentation/markup.html#roles ?

(We should add `gh-label` there too.)

CAM-Gerlach (commented):
This is very helpful; we could add this to the PEPs too.

I do agree with @hugovk that this really should be literal-formatted like the original `:file:` role, though I'm not entirely sure how to achieve that without just re-implementing the link manually.

ezio-melotti:
> One idea, filenames are often written with a fixed-width font (e.g. in this [style guide](https://bishopfox.com/resources/cybersecurity-style-guide)), shall we have the role do that too?

It took me a bit but eventually (and with the help of @AA-Turner and @CAM-Gerlach) I figured out how to make it a literal, in a way that is simple and concise enough.

While I was at it, I also added support for `!` so that we can do e.g. `` :cpy-file:`!Makefile` `` without creating a link (since `Makefile` is generated, and not in the actual repo).

I also thought about adding support for variables parts like `` :cpy-file:`somedir/python3.{x}/somefile.py` ``, but this seems more involved and IMHO not worth it.  If there is a variable part, the file can't be linked anyway and in that case a regular `:file:` can be used (even if it's slightly less semantic in the source, but otherwise unnoticeable since they render the same).

Support for `~` (to strip the dirs and leave the filename) seems YAGNI to me, but it's not too difficult to add if we really wanted to.

This is how the output of the current PR looks like:
image

> And document this role at https://cpython-devguide--961.org.readthedocs.build/documentation/markup.html#roles ?

Those roles are just for the `cpython` repo so I don't think this should be documented.  If we end up adding it to `python/cpython` (and `python/peps` too), then it can be documented.  (Doing that would also mean duplicating (or triplicating) the code, so we might want to come up with a better solution.)

<details>
<summary>Click to see some related findings</summary>

Sphinx defines a bunch of extra nodes, including a `literal_emphasis`, but it doesn't support variables parts and just makes everything italic.  I haven't seen a node that already does what we need:
  https://github.com/sphinx-doc/sphinx/blob/6ed4bbba396eb410c4d00e5f9881aa621cec749a/sphinx/addnodes.py#L512

A new node type that combines `literal` and `reference` could be created and used as well:
  https://github.com/docutils/docutils/blob/c5a6ec93fe81fc44851f65aff37bef8a4cd2d173/docutils/docutils/nodes.py#L1893-L1894

Sphinx also defines a `EmphasizedLiteral` role (which is what is used for the `:file:` role):
  https://github.com/sphinx-doc/sphinx/blob/6ed4bbba396eb410c4d00e5f9881aa621cec749a/sphinx/roles.py#L272

If we wanted to handle this at the *role* level (rather than the *node* level) we could subclassed that and enhance it to support linking.  This shouldn't be *too* difficult, but it's not entirely obvious to me how to do it and I'm happy with the current solution.

</details>

CAM-Gerlach:
Ah, I didn't realize `autolink` was a custom role defined for this repo (I'd seen it mentioned being used on the NumPy repo and elsewhere); my assumptions and a few comments I'd seen when googling had led me to believe otherwise, and we'd have to re-implement it ourselves (in which case I went with the class approach), but don't know why I didn't double check. 

In any case, yeah, its simple to add that given it's already defined here, especially for a first implementation. As for

> If we end up adding it to python/cpython (and python/peps too), then it can be documented. (Doing that would also mean duplicating (or triplicating) the code, so we might want to come up with a better solution.)

I've implemented (but not yet tested) [a common base class](https://gist.github.com/CAM-Gerlach/0a48cc966ce04a54e4b762ec62f6f1ac) that provides a superset of the existing functionality, including everything here plus customizable branches/tags/commits, `~` support, custom link titles (which are automatically not rendered as a literal), emulating the CSS classes of the file role, and some other things, while being fully extensible to files on any GitHub project, or indeed any arbitrary link.

It needs some more refactoring (to move the default org/repo to config settings rather than a custom subclass, and provide a factory function to generate new ones), a couple more features (support for fragments and validation so it works with PEPs/RFCs, support for GitHub issues/PRs), and a few tweaks (using `@` instead of `:` for branches/tags/commits, adding URL escaping). However, it could form a common base class (and subclasses) implemented in a Sphinx extension to provide many of the similar, currently-custom roles used/desired across our repos (GitHub issue/PR, BPO issue, PEPs, and RFCs, alongside source files, which could all specify other repos or even orgs if needed), with a superset of the existing functionality, to replace the patchwork of repo-bespoke code. Going forward, building off this would likely make more sense than copying things around piecemeal between repos, and would be easily reusable for other projects as well.

> If we wanted to handle this at the role level (rather than the node level) we could subclassed that and enhance it to support linking. This shouldn't be too difficult, but it's not entirely obvious to me how to do it and I'm happy with the current solution.

That's quite _possible_, but would require basically re-implementing the role, which is really just the same as `:samp:` with an different CSS class, as it is basically one big monolithic function and almost all the logic is devoted to parsing the sample bits, which aren't useful for a concrete specific file (that this role represents) as opposed to an abstract file that the `file` role corresponds to. Your approach here is far more reasonable, especially as a first implementation.

Longer-term, the above proposal features a common base class with many small, overridable methods, which can be easily subclassed used for other types of links and custom behavior, which avoids having to reimplement everything.

_extensions/custom_roles.py
```diff
@@ -19,12 +25,16 @@ def setup(app):
...
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+        """Combine literal + reference (unless the text is prefixed by a !)."""
         if " " in text:
             url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
         url = pattern % (url_text,)
-        node = nodes.reference(rawtext, text, refuri=url, **options)
+        node = nodes.literal('', text.lstrip('!'), **options)
```
CAM-Gerlach, line 34:
```suggestion
        node = nodes.literal(rawtext, text.lstrip('!'), **options)
```

The source text should be included here, as it was before—IIRC, it is used at least in error messages, and possibly by translations and other extensions

_extensions/custom_roles.py
```diff
@@ -19,12 +25,16 @@ def setup(app):
 
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+        """Combine literal + reference (unless the text is prefixed by a !)."""
         if " " in text:
             url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
         url = pattern % (url_text,)
```
CAM-Gerlach, line 33:
```suggestion
        url = pattern.format(url_text)
```

Maybe worth using modern format syntax rather than old-style printf?

_extensions/custom_roles.py
```diff
@@ -8,6 +8,12 @@
 
 
 def setup(app):
+    # role to link to cpython files
+    app.add_role(
+        "cpy-file",
+        autolink("https://github.com/python/cpython/blob/main/%s"),
```
CAM-Gerlach, line 14:
```suggestion
        autolink("https://github.com/python/cpython/blob/main/{}"),
```

Corresponding change here

CAM-Gerlach (commented):
A few minor comments; otherwise LGTM for now :+1:

hugovk:
> Ah, I didn't realize `autolink` was a custom role defined for this repo (I'd seen it mentioned being used on the NumPy repo and elsewhere);

Yeah, it's one of those handy bits of sample code that gets copied and pasted around a lot!

_extensions/custom_roles.py
```diff
@@ -8,6 +8,12 @@
...
+    )
+    # role to link to cpython labels
     app.add_role(
         "gh-label",
         autolink("https://github.com/python/cpython/labels/%s"),
```
ezio-melotti, line 19:
```suggestion
        autolink("https://github.com/python/cpython/labels/{}"),
```

_extensions/custom_roles.py
ezio-melotti: Good catch!

_extensions/custom_roles.py
```diff
@@ -19,12 +25,16 @@ def setup(app):
 
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
```
AA-Turner, line 27: options and content are only needed when custom in-document roles are used (`.. role:: new(base)`)

```suggestion
    def role(name, rawtext, text, lineno, inliner, _options=None, _content=None):
```

_extensions/custom_roles.py
```diff
@@ -19,12 +25,16 @@ def setup(app):
...
             url_text = urllib.parse.quote(f"{text}")
         else:
             url_text = text
-        url = pattern % (url_text,)
-        node = nodes.reference(rawtext, text, refuri=url, **options)
+        url = pattern.format(url_text)
+        node = nodes.literal(rawtext, text.lstrip('!'), **options)
+        # don't create a reference if the text starts with !
+        if not text.startswith('!'):
+            node = nodes.reference('', '', node, refuri=url, **options)
```
AA-Turner, line 37: The current implementation disallows e.g. ``:cpy-file:`!!spam!.txt` `` (I guess this is never likely to actually be needed). Refactored to skip options per above, and pass internal=False/rawtext to the reference node.
```suggestion
        # don't create a reference if the text starts with !
        if text.startswith('!'):
            node = nodes.literal(rawtext, text[1:])
        else:
            node = nodes.reference(rawtext, '', nodes.literal(rawtext, text),
			                       refuri=url, internal=False)
```

_extensions/custom_roles.py
```diff
@@ -19,12 +25,16 @@ def setup(app):
 
 def autolink(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+        """Combine literal + reference (unless the text is prefixed by a !)."""
         if " " in text:
             url_text = urllib.parse.quote(f"{text}")
```
AA-Turner, line 30: This is already a string, so the f-string call is unneeded
```suggestion
            url_text = urllib.parse.quote(text)
```

AA-Turner (commented):
Docutils things:

A

_extensions/custom_roles.py
```diff
@@ -8,23 +8,35 @@
...
             url_text = text
-        url = pattern % (url_text,)
-        node = nodes.reference(rawtext, text, refuri=url, **options)
+        url = pattern.format(url_text)
+        # don't create a reference if the text starts with !
+        if text.startswith('!'):
+            node = nodes.literal(rawtext, text[1:])
+        else:
+            node = nodes.reference(rawtext, '', nodes.literal(rawtext, text),
+			           refuri=url, internal=False)
```
ezio-melotti, line 39:
```suggestion
            node = nodes.literal(rawtext, text[1:])
        else:
            node = nodes.reference(rawtext, '', nodes.literal(rawtext, text),
                                   refuri=url, internal=False)
```

AA-Turner: oops!

CAM-Gerlach (commented):
Seems to work well overall now. While more features can be added in the future, or it can be replaced with a more powerful, flexible and extensible alternative, IMO the main limitation here that could be worth addressing now is that there is currently no support for explicit titles/overriding the default title; i.e. `:cpy-file:`Python's PEG generator <Tools/peg_generator/>` displays:

[`Python's PEG generator <Tools/peg_generator/>`](https://example.com)

But I'm not sure how easy that would be to add aside from manual parsing logic; my longer-term approach relies on the `ReferenceRole` superclass to handle that. @AA-Turner thoughts?

ezio-melotti:
I'm going to merge this and follow up with another PR to use it with other files.  We can then improve on it with other PRs and possibly port it to `cpython` and the `peps` repos.

CAM-Gerlach (approved):
SGTM, thanks @ezio-melotti

ezio-melotti:
I created a new PR to use the new role throughout the Devguide:
* #984

PR #981, opened 2022-11-02
Fix copying of multiline snippets

hauntsaninja:
For instance, the macOS build command

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -107,3 +107,5 @@ todo_include_todos = True
 # Strip the dollar prompt when copying code
 # https://sphinx-copybutton.readthedocs.io/en/latest/use.html#strip-and-configure-input-prompts-for-code-cells
 copybutton_prompt_text = "$"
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Thanks, it's working now:

# Before

Clicking the copy button for "For Python 3.10 and newer" at https://devguide.python.org/getting-started/setup-building/#macos-and-os-x gives:

```sh
 CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
```

# After

But with the fix at https://cpython-devguide--981.org.readthedocs.build/getting-started/setup-building/#macos-and-os-x it gives:

```sh
 CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
  LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
  PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
  ./configure --with-pydebug \
              --with-openssl=$(brew --prefix openssl)
```

PR #983, opened 2022-11-10
Explain that the contributors should base changes against `main`

ambv:
This isn't a particularly common issue but every now and then somebody will waste their time by preparing a change against a maintenance branch instead of working with `main` from the get-go.

This change serves as a reminder to contributors to think about the `main` branch first.

Key changes:
Screen Shot 2022-11-10 at 17 30 43

Screen Shot 2022-11-10 at 17 30 57

### DIFF
--- a/getting-started/pull-request-lifecycle.rst
+++ b/getting-started/pull-request-lifecycle.rst
@@ -86,7 +86,7 @@ You should have already :ref:`set up your system <setup>`,
      git checkout -b <branch-name> upstream/main
 
   .. note::
-     Almost all changes to Python, including bug fixes, should first be 
+     Almost all changes to Python, including bug fixes, should first be
      made against the ``main`` branch. See :ref:`here <good-prs>` for
      more details.
 
--- a/getting-started/pull-request-lifecycle.rst
+++ b/getting-started/pull-request-lifecycle.rst
@@ -29,7 +29,8 @@ Here is a quick overview of how you can contribute to CPython:
 
 #. `Create an issue`_ that describes your change [*]_
 
-#. :ref:`Create a new branch in Git <pullrequest-steps>`
+#. :ref:`Create a new branch in Git <pullrequest-steps>` from the
+   ``main`` branch
 
 #. Work on changes (e.g. fix a bug or add a new feature)
 
@@ -79,9 +80,15 @@ You should have already :ref:`set up your system <setup>`,
 
      git fetch upstream
 
-* Create a new branch in your local clone::
+* Create a new branch in your local clone from the ``main`` branch
+  in the upstream repository::
 
      git checkout -b <branch-name> upstream/main
+
+  * NOTE: Don't work on old branches of Python. Even if you're working on
+    a fix for a bug you discovered in an older version of Python, we
+    want to fix it in ``main`` first and then
+    :ref:`port your fix back <branch-merge>` later.
 
 * Make changes to the code, and use ``git status`` and ``git diff`` to see them.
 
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Looks good, we could consider putting one of the early mentions of `main` in bold text.

getting-started/pull-request-lifecycle.rst
```diff
@@ -180,7 +187,14 @@ Making Good PRs
 When creating a pull request for submission, there are several things that you
 should do to help ensure that your pull request is accepted.
 
-First, make sure to follow Python's style guidelines. For Python code you
+First, make your change against the right version of Python. Even if you're
+fixing a bug, your fix needs to go to the ``main`` branch first. After the
+fix is merged, it will be :ref:`ported back <branch-merge>` to older
+maintenance releases as well. That way we ensure all affected versions are
+handled. Unless you're sure that your change does not apply to ``main``,
+don't base your fixes on old versions of Python.
```
ezio-melotti, line 195: The tone seems to assume that they are going to do it against the wrong version.  I would use a more neutral tone and state that almost all changes (including bug and security fixes) should be done against `main`, but in certain cases they should be done on older branches (maybe list a couple of examples (IIRC they are already mentioned somewhere)).

Providing the background for this (e.g. by explaining that we backport fixes like you did) and cases where the PR should be done on old branches, will help reassure the reader about which versions will receive the fix.

getting-started/pull-request-lifecycle.rst
```diff
@@ -79,10 +80,16 @@ You should have already :ref:`set up your system <setup>`,
...
-* Create a new branch in your local clone::
+* Create a new branch in your local clone from the ``main`` branch
+  in the upstream repository::
 
      git checkout -b <branch-name> upstream/main
 
+  * NOTE: Don't work on old branches of Python. Even if you're working on
+    a fix for a bug you discovered in an older version of Python, we
+    want to fix it in ``main`` first and then
+    :ref:`port your fix back <branch-merge>` later.
```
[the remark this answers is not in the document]
CAM-Gerlach: And if it _is_ a note and is called out with an ad-hoc ``NOTE:`` plain text, we should just make it an actual ``.. note::`.

getting-started/pull-request-lifecycle.rst
```diff
@@ -180,7 +187,14 @@ Making Good PRs
...
 
-First, make sure to follow Python's style guidelines. For Python code you
+First, make your change against the right version of Python. Even if you're
+fixing a bug, your fix needs to go to the ``main`` branch first. After the
+fix is merged, it will be :ref:`ported back <branch-merge>` to older
+maintenance releases as well. That way we ensure all affected versions are
+handled. Unless you're sure that your change does not apply to ``main``,
+don't base your fixes on old versions of Python.
+
+Second, make sure to follow Python's style guidelines. For Python code you
```
CAM-Gerlach, line 197: Sidenote: To avoid noisy, tedious and easy to mess up manual renumbering like this, a long list of prose-numbered paragraphs, and where there isn't an inherent order, it would be better to reformat this as an unordered (bulleted) list instead (or at least use an auto-numbered ordered list). But perhaps out of scope here.

ambv: I remember we already deferred doing this once to a separate PR so I'll just do it here then.

ambv:
Thanks for your insightful comments. Addressed in the update. Current state:
Screen Shot 2022-11-11 at 11 01 37

Screen Shot 2022-11-11 at 11 02 12

JelleZijlstra (approved):

PR #984, opened 2022-11-14
Use `:cpy-file:` throughout the Devguide

ezio-melotti:
This PR replaces the use of `:file:` with `:cpy-file:` (introduced in #961) in most places.  Since the generated links are now checked by `make linkcheck`, I also went through and corrected some of those.

There are a few broken links that I haven't touched yet:
> (developer-workflow/grammar: line   40) broken    https://github.com/python/cpython/blob/main/Include/Python-ast.h
> (developer-workflow/grammar: line   33) broken    https://github.com/python/cpython/blob/main/Include/token.h
> (internals/compiler: line  516) broken    https://github.com/python/cpython/blob/main/Include/code.h
> (internals/compiler: line  488) broken    https://github.com/python/cpython/blob/main/Python/peephole.c

* `Include/Python-ast.h` was removed by @vstinner
  *  https://github.com/python/cpython/pull/24933
* `Include/token.h` was removed by @vstinner
  *  https://github.com/python/cpython/pull/92652
* `Include/code.h` was removed by @vstinner
  *  https://github.com/python/cpython/pull/32385
* `Python/peephole.c` was removed by @markshannon
  *  https://github.com/python/cpython/pull/21517

The respective sections in `developer-workflow/grammar.rst` and `internals/compiler.rst` might need to be revisited.
@vstinner and @markshannon: can you advise on whether these are quick fixes that we can include in this PR or if they should be handled separately?

I also found another issue with this section: https://devguide.python.org/internals/compiler/#important-files
In addition to the fact that the listed files don't use any markup, the markup used for most of the files seem incorrect (it's a blockquote due to the indentation).  A nested list with full paths for all the files and `:cpy-file:` should be a better alternative.

The `:cpy-file:` role could also be improved to support `~`, so that in that list (and in a few other places) we can hide the full path and just display the file name.


core-developers/experts.rst
```diff
@@ -17,7 +17,7 @@ while any other issues can and should be decided by any committer.
 
 Developers can choose to follow labels, so if a label that they are
 following is added to an issue or pull request, they will be notified
-automatically.  The :file:`CODEOWNERS` file is also used to indicate
+automatically.  The :cpy-file:`.github/CODEOWNERS` file is also used to indicate
```
ezio-melotti, line 20: If we add `~` support to `:cpy-file:` we could hide the dir.

developer-workflow/c-api.rst
```diff
@@ -101,12 +101,12 @@ C API Tests
 Tests for the public C API live in the ``_testcapi`` module.
 Functions named ``test_*`` are used as tests directly.
 Tests that need Python code (or are just easier to partially write in Python)
-live in ``Lib/test``, mainly in :file:`Lib/test/test_capi.py`.
+live in ``Lib/test``, mainly in :cpy-file:`Lib/test/test_capi`.
```
ezio-melotti, line 104: `test_capi` is now a dir.

internals/compiler.rst
```diff
@@ -142,7 +143,7 @@ In general, unless you are working on the critical core of the compiler, memory
 management can be completely ignored.  But if you are working at either the
 very beginning of the compiler or the end, you need to care about how the arena
 works.  All code relating to the arena is in either
-:file:`Include/Internal/pycore_pyarena.h` or :file:`Python/pyarena.c`.
+:cpy-file:`Include/internal/pycore_pyarena.h` or :cpy-file:`Python/pyarena.c`.
```
ezio-melotti, line 146: `internal` is lowercase.  There were a few similar errors in this file.

internals/compiler.rst
```diff
@@ -532,12 +535,12 @@ Important Files
 
     asdl_c.py
```
ezio-melotti, line 536: All the files in this section should be linkified.  I left a comment about it in the PR.

developer-workflow/c-api.rst
hugovk, line 104: We could also link things that aren't already `:file:`, like `Lib/test`:

```suggestion
live in :cpy-file:`Lib/test`, mainly in :cpy-file:`Lib/test/test_capi`.
```

There's a lot more that could be linked on this page too, especially the first few sections.

But let's keep the scope of this PR to replacing the old `:file:` with the new `:cpy-file:`.

hugovk (approved):
Gave a few a spot check, looks good.

developer-workflow/c-api.rst
ezio-melotti: Agreed on keeping the PR focused.  We should also try to create links only when/where they are useful, but there are definitely more places where `:cpy-file:` can be used.

CAM-Gerlach: That SGTM

PR #985, opened 2022-11-14
Build with Python 3.11 on CI and RTD

hugovk:
Build the docs on GitHub Actions and Read the Docs using the new Python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
           cache: pip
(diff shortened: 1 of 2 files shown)


AA-Turner (approved):
An alternate is to use "3" for both to use the latest stable Python, but this looks good for explicitness and simplicity.

A

hugovk:
There is no `3-dev`, so `3` it is.

Demo: https://github.com/hugovk/test/actions/runs/3464785635

hugovk:
RTD also supports `3` for the latest version, so updated that to match:

* https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python
* https://readthedocs.org/projects/cpython-devguide/builds/18635085/

ezio-melotti (approved):

PR #988, opened 2022-11-20
Generate release cycle chart and CSV

hugovk:
For https://github.com/python/docs-community/issues/67.

Add a script to read in a JSON file of data about Python releases: PEP no, branch status, release manager and date, and EOL.

The script generates:

* Two CSV files
  * One for active branches
  * One for EOL branches

* A [Mermaid Gantt chart file](https://mermaid-js.github.io/mermaid/#/gantt)

The CSV files are used for the tables (https://github.com/python/devguide/pull/884) and the Mermaid .mmd file is used for the release cycle chart:

image

image

Rather than a Sphinx extension, this is a standalone script that can be run via `make version` 	or `python generate-release-cycle.py`. I've also taken advantage of standard Makefile logic so the CSV and Mermaid files are regenerated when the JSON file is modified. They're also regenerated when running `make html`, and only if the JSON file has been modified.

# Questions

Q1: Is this Makefile method of generating the files acceptable? If so, I'll need a hand with the Windows `make.bat` (I don't have Windows).

Q2: Do we want the Python script in a subdirectory?


include/branches.csv
```diff
@@ -1,7 +1,7 @@
-Branch,Schedule,Status,First release,End-of-life,Release manager
-main,:pep:`693`,features,*2023-10-02*,*2028-10*,Thomas Wouters
-3.11,:pep:`664`,bugfix,2022-10-24,*2027-10*,Pablo Galindo Salgado
-3.10,:pep:`619`,bugfix,2021-10-04,*2026-10*,Pablo Galindo Salgado
-3.9,:pep:`596`,security,2020-10-05,*2025-10*,Łukasz Langa
-3.8,:pep:`569`,security,2019-10-14,*2024-10*,Łukasz Langa
-3.7,:pep:`537`,security,2018-06-27,*2023-06-27*,Ned Deily
+Branch,Schedule,Status,First release,End-of-life,Release manager
```
CAM-Gerlach, line 1: I'm guessing maybe the EoL changed here? Can you double-check its (now) `\n`?

CAM-Gerlach (commented):
IMO, I'm personally not a huge fan of making building the devguide much more dependent upon ye olde not-so-cross-platform `make`; I'd much prefer a Sphinx extension that would work reliably cross platform with any of the standard Sphinx invocations. It's also more reusable that way, if we ever want to use it on other repos.

That said, that's just my personal take; I don't mean to tell a Core Developer like you what do do :stuck_out_tongue: And to be fair, back in the day I basically did this same architecture for generating a production version of our Spyder-Docs site with mutliversion, redirects, etc, before I knew anything about writing a Sphinx extension (and doing so would have been more difficult for that use case).

Windows is my primary platform, but I've never for the life of me had to write a batch file so I wouldn't know where to start there (I have a Windows build of `make` for when I need it, but I rarely do—ironically, the one time I do is for building the Spyder-Docs site mentioned above in production mode with multiversion).

Makefile
```diff
@@ -186,3 +187,17 @@ check: ensure-venv
...
+
+# TODO make.bat
+include/branches.csv: include/release-cycle.json
+	$(PYTHON) generate-release-cycle.py
+
+include/end-of-life.csv: include/release-cycle.json
+	$(PYTHON) generate-release-cycle.py
+
+include/release-cycle.mmd: include/release-cycle.json
+	$(PYTHON) generate-release-cycle.py
```
merwok, line 199: So this command recreates all three files (guessing from the absence of arguments), but will be called three times when the source file changes?

generate-release-cycle.py
```diff
@@ -0,0 +1,131 @@
...
+    "security": "done,",
+    "end-of-life": "crit,",
+}
+
+CSV_HEADER = (
+    "Branch",
+    "Schedule",
+    "Status",
+    "First release",
+    "End-of-life",
```
merwok, line 34: Hyphens are used in adjective, for a column heading that’s a noun, we’d have `End of life`

generate-release-cycle.py
```diff
@@ -0,0 +1,131 @@
...
+    "Release manager",
+)
+
+
+def cycle(branch: str) -> str:
+    """Convert branch name to version number"""
+    return "3.12" if branch == "main" else branch
```
merwok, line 41: Wonder if we could have the info about current branch in the CSV somehow (bool column?) to avoid needing to remember to change the script.

CAM-Gerlach: Or separate the version name from the branch name for the active branches, and have the key be the former rather than the latter? We could at least make that a script-level constant at the top.

Makefile
hugovk: Yes and no: make is smart and will only call the script once. This is the sort of stuff [make](https://www.gnu.org/software/make/manual/make.html#Overview) is built for, it: 

> determines automatically which pieces of a large program need to be recompiled, and issues the commands to recompile them.

To demonstrate by temporarily adding something to the script:

```console
$ git diff
diff --git a/generate-release-cycle.py b/generate-release-cycle.py
index 16dd23b..35bb73c 100644
--- a/generate-release-cycle.py
+++ b/generate-release-cycle.py
@@ -122,6 +122,7 @@ class Versions:
 
 
 def main() -> None:
+    print("main")
     versions = Versions()
     versions.save_csv()
     versions.save_mermaid()
$ touch include/release-cycle.json
$ make versions
python3 generate-release-cycle.py
main
Release cycle data generated.
```

`main` is printed only once.

generate-release-cycle.py
hugovk: Have `"3.12"` in the JSON, and for the CSV replace the highest version with `main`?

generate-release-cycle.py
```diff
@@ -0,0 +1,131 @@
...
+        with open("include/release-cycle.json") as f:
+            self.versions = json.load(f)
+
+    def save_csv(self) -> None:
+        """Output CSV files"""
+        now_str = str(dt.datetime.utcnow())
+
+        with (
+            open("include/branches.csv", "w", newline="") as file_branches,
+            open("include/end-of-life.csv", "w", newline="") as file_eol,
```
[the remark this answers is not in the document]
hugovk: Thanks, we already have non-ASCII.

generate-release-cycle.py
```diff
@@ -0,0 +1,131 @@
...
+                    version["status"],
+                    csv_date(version["releaseDate"], now_str),
+                    csv_date(version["eol"], now_str),
+                    version["releaseManager"],
+                )
+
+                if version["status"] == "end-of-life":
+                    csv_eol.writerow(row)
+                else:
+                    csv_branches.writerow(row)
```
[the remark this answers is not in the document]
hugovk: Hmm, we'd still need to manually munge the keys (e.g. `"releaseManager"` -> `"Release manager"`) and also values (e.g. `664` -> ```":pep:`693`"```).

We could change the keys in the JSON to match what we want in the output, but it's best to avoid spaces in keys (so JavaScript can reference using dot notation).

include/branches.csv
hugovk: Got confused by end-of-life/end-of-line there for a moment!

It was defaulting to `\r\n`, now updated to `\n`:

https://docs.python.org/3/library/csv.html?highlight=csv#csv.Dialect.lineterminator

generate-release-cycle.py
merwok: Sure, that should work too

versions.rst
```diff
@@ -32,6 +32,12 @@ Unsupported Versions
    :file: include/end-of-life.csv
 
 
+Python Release Cycle
+====================
+
+.. mermaid:: include/release-cycle.mmd
+
+
```
ezio-melotti, line 40: I would put this above the two tables.

include/branches.csv
CAM-Gerlach: Oh, haha—should have anticipated that!

Yeah, that triggers my memory on how to do that, thanks—I'd remembered doing it before, but its been a while since I've used `csv` instead of Pandas and browsing through the docs, I couldn't find it—just admonitions to only use `newline=""` in the `open` call.

generate-release-cycle.py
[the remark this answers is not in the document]
CAM-Gerlach: Yeah, though I still feel it would be better to just do the munging in the structured `dict` (via f-strings, lookup tables, etc) and then output it via the standard method, rather than rolling out own.

generate-release-cycle.py
CAM-Gerlach:
```suggestion
    "End of life",
```

generate-release-cycle.py
CAM-Gerlach: That could, though for programmatic use (i.e. what the JSON is primarily for) its likely to be helpful to have the branch name and version name as separate fields, since one or the other might be needed for a particular use.

merwok:
On a temporary branch for this I pointed out that the label `features` should be `feature`, to match the term `feature release` (like the status `bugfix` matches `bugfix release`).  You could say that it’s not in the scope of this particular PR but I would say that it’s the opportunity to do this minor cleanup 🙂

generate-release-cycle.py
hugovk: So something like?

```json
[
  {
    "branch": "main",
    "cycle": "3.12",
    ...
  },
  {
    "branch": "3.11",
    "cycle": "3.11",
    ...
  },
  {
    "branch": "3.12",
    "cycle": "3.12",
    ...
  },
...
```

generate-release-cycle.py
hugovk: Updated!

versions.rst
hugovk: Updated!

Shall we also ditch the subheading:

```rst
Python Release Cycle
====================
```

so it's directly under the intro "The main branch is currently..." paragraph?

hugovk:
> On a temporary branch for this I pointed out that the label `features` should be `feature`, to match the term `feature release` (like the status `bugfix` matches `bugfix release`). You could say that it’s not in the scope of this particular PR but I would say that it’s the opportunity to do this minor cleanup 🙂

Updated!

---

Will check the other review comments a bit later, thanks all!

generate-release-cycle.py
CAM-Gerlach: Yeah, that's what I was thinking, but just as a dict instead of a list of dicts with `cycle` also as the key, like we did for the `peps.json`, so its much simpler and more efficient to access a specific release cycle without having to iterate and check through the whole thing.

hugovk: Updated!

generate-release-cycle.py
```diff
@@ -0,0 +1,136 @@
...
+        date_str = f"{date_str}-01"
+    return date_str
+
+
+class Versions:
+    """For converting JSON to CSV and Mermaid"""
+
+    def __init__(self) -> None:
+        with open("include/release-cycle.json") as f:
+            self.versions = json.load(f)
```
CAM-Gerlach, line 65:
```suggestion
        with open("include/release-cycle.json", encoding="UTF-8") as in_file:
            self.versions = json.load(in_file)
```

Add missing encoding and use a more descriptive filename

generate-release-cycle.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+
+            csv_branches.writerow(CSV_HEADER)
+            csv_eol.writerow(CSV_HEADER)
+
+            sorted_versions = sorted(
+                self.versions.items(),
+                key=lambda k: int(k[0].replace(".", "")),
+                reverse=True,
+            )
```
CAM-Gerlach, line 93: This won't actually work quite right, as it will sort all `X.YY` versions ahead of all `X.Y` versions. My refactored version, https://github.com/CAM-Gerlach/devguide/commit/c2cbf6b0326d6b63c3a35934529c8247f83d5785 , replaces the `key` line with `lambda k: [int(i) for i in k.split(".")]` which performs the same sorting as with standard version tuples.

generate-release-cycle.py
```diff
@@ -0,0 +1,136 @@
...
+}
+
+CSV_HEADER = (
+    "Branch",
+    "Schedule",
+    "Status",
+    "First release",
+    "End of life",
+    "Release manager",
+)
```
CAM-Gerlach, line 36: These are duplicated down below; my suggested refactoring https://github.com/CAM-Gerlach/devguide/commit/c2cbf6b0326d6b63c3a35934529c8247f83d5785 fixes this

generate-release-cycle.py
```diff
@@ -0,0 +1,136 @@
...
+        with open(
+            "include/release-cycle.mmd", "w", encoding="UTF-8", newline="\n"
+        ) as f:
+            f.writelines(out)
+
+
+def main() -> None:
+    versions = Versions()
+    versions.save_csv()
+    versions.save_mermaid()
```
CAM-Gerlach, line 132: Trivial bikesheddy nitpick, but wouldn't `write_*` be more idiomatic?

CAM-Gerlach (commented):
Per our previous discussion, I've worked on a suggested revision: https://github.com/CAM-Gerlach/devguide/commit/c2cbf6b0326d6b63c3a35934529c8247f83d5785 factoring the dict preprocessing and the actual CSV writing into separate, modular and easily-composable steps, which IMO:

* is a cleaner design (and could be separated into smaller, more atomic functions)
* is a fair bit shorter/simpler
* avoids the code smell of opening two files, creating two CSV writers, having two write calls, etc.
* is more efficient (writing each CSV in one go and only having one file open for a shorter time)
* eliminates the redundant and potentially bug-magnet duplicate CSV_HEADER and listing of keys in the final dict

I also fixed the version sorting issue there.

In addition, I have a few comments/suggestions below, and a couple overall:

* [x] ~One thing that's missing here is a clear indication of the expected/actual cut-over point between bugfix and security-only releases. Could we add an additional column with this information, and indicate it as such on the chart? This is standard for most pages like this, and the [Wikipedia page](https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions) has it. Otherwise, it's not clearly indicated anywhere here when the cutover is planned/has occurred.~ Deferred to a followup
* [ ] Could we maybe try to keep the JSON key names and order more in sync with the table? I don't really see a compelling reason to not just use the same keys and order, minus formatting (spaces, caps)
* [x] Wouldn't it be more idiomatic to name the Python file with `_`s? Then it could be imported if needed. Otherwise, doesn't really make sense to have a `main()` function and an `if __name__ == "__main__"` block since it isn't possible to import it without `importlib` hackery.
* [x] ~Maybe its just me, but I'd certainly prefer if this was a Sphinx extension integrated into the build as opposed to a standalone script invoked manually or via a non-cross-platform Makefile...~

generate-release-cycle.py
hugovk: I don't mind either way, fine to change it 👍

hugovk:
> Per our previous discussion, I've worked on a suggested revision: [CAM-Gerlach@c2cbf6b](https://github.com/CAM-Gerlach/devguide/commit/c2cbf6b0326d6b63c3a35934529c8247f83d5785) factoring the dict preprocessing and the actual CSV writing into separate, modular and easily-composable steps, which IMO:
> 
> * is a cleaner design (and could be separated into smaller, more atomic functions)
> * is a fair bit shorter/simpler
> * avoids the code smell of opening two files, creating two CSV writers, having two write calls, etc.
> * is more efficient (writing each CSV in one go and only having one file open for a shorter time)
> * eliminates the redundant and potentially bug-magnet duplicate CSV_HEADER and listing of keys in the final dict
> 
> I also fixed the version sorting issue there.

Thanks, I've integrated that stuff 👍 

> In addition, I have a few comments/suggestions below, and a couple overall:
> 
> * [ ]  One thing that's missing here is a clear indication of the expected/actual cut-over point between bugfix and security-only releases. Could we add an additional column with this information, and indicate it as such on the chart? 

Is it okay to deal with that separately after this PR, to reduce the scope? At least we don't have that information in the current tables.

> This is standard for most pages like this, and the [Wikipedia page](https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions) has it. Otherwise, it's not clearly indicated anywhere here when the cutover is planned/has occurred.

The table does, but the chart doesn't. (Aside: it's a bit strange the Wikipedia table has "End of full support" before "End of security fixes".)


> * [ ]  Could we maybe try to keep the JSON key names and order more in sync with the table? I don't really see a compelling reason to not just use the same keys and order, minus formatting (spaces, caps)

Can do!

I think we should keep `"pep": 693` instead of `"schedule": 693`?

So like this?

```diff
diff --git a/include/release-cycle.json b/include/release-cycle.json
--- a/include/release-cycle.json
+++ b/include/release-cycle.json
 {
   "3.12": {
     "branch": "main",
     "pep": 693,
     "status": "feature",
-    "release_manager": "Thomas Wouters",
-    "release_date": "2023-10-02",
-    "eol": "2028-10"
+    "first_release": "2023-10-02",
+    "end_of_life": "2028-10",
+    "release_manager": "Thomas Wouters"
   },
```


> * [ ]  Wouldn't it be more idiomatic to name the Python file with `_`s? Then it could be imported if needed. Otherwise, doesn't really make sense to have a `main()` function and an `if __name__ == "__main__"` block since it isn't possible to import it without `importlib` hackery.

Yep, renamed! Shall we keep the script in the root or move it to some subdirectory?

> * [ ]  Maybe its just me, but I'd certainly prefer if this was a Sphinx extension integrated into the build as opposed to a standalone script invoked manually or via a non-cross-platform Makefile...

I slightly prefer `Makefile`, these files won't need updating much, and it would be nice to avoid having to run this for every single `make pages` iteration when working on something else. Also no need to write/maintain Sphinx extension. If we use a non-cross-platform `Makefile`, we'll also update the non-cross-platform `make.bat` for Windows users ;)

But then again there's not much overhead in reading one file and writing two.

What are others' thoughts?

merwok:
A script with a makefile (and if possible windows script equivalent, but not a blocker) seems simplest and best to me.

generate-release-cycle.py
CAM-Gerlach: > That's the intention:
> 
> ```python
> >>> sorted(["30", "31", "310", "311"], reverse=True)
> ['311', '310', '31', '30']
> ```

I'm a little confused what you mean here—the sorting algorithm being used here is sorting these as integers, not as strings. In any case, this algorithm does not give the correct result for any two-digit minor versions 2 or higher, which are not included in the above set:

```python
>>> sorted(["30", "31", "32", "310", "311"], reverse=True)
['32', '311', '310', '31', '30']
```

whereas tuple sorting sorts this correctly:

```python
>>> sorted([(3, 0), (3, 1), (3, 2), (3, 10), (3, 11)], reverse=True)
[(3, 11), (3, 10), (3, 2), (3, 1), (3, 0)]
```

Moreover, the actual sorting here is done by casting the above strings to integers, and while that does happen to work on the currently-released Python versions, it only does so by accident because there happens to be only one major version that has two-digit minor versions, and _that_ in turn happens to be the latest version. Look what happens when you add a `4.0`:

```python
>>> sorted([30, 31, 310, 311, 40], reverse=True)
[311, 310, 40, 31, 30]
```

Whereas again, tuple sorting handles this correctly:

```python
>>> sorted([(3, 0), (3, 1), (3, 2), (3, 10), (3, 11), (4, 0)], reverse=True)
[(4, 0), (3, 11), (3, 10), (3, 2), (3, 1), (3, 0)]
```

CAM-Gerlach:
> Is it okay to deal with that separately after this PR, to reduce the scope? At least we don't have that information in the current tables.

Unacceptable I say, unacceptable! :laughing: Yeah, that seems quite reasonable to me, though its not really up to me either way, heh.

> The table does, but the chart doesn't. (Aside: it's a bit strange the Wikipedia table has "End of full support" before "End of security fixes".)

How so, sorry? Since it occurs between initial release and final security EoL, it would seem to me to make the most sense to put it between the two, unless there's something I'm missing...

> I think we should keep "pep": 693 instead of "schedule": 693?
>
> So like this?

Sure, that seems to make the most sense to me; in general, I don't have any strong preference what they are, just that they match if practicable. In particular, I realize I was unclear before—I didn't mean to imply that the JSON should be changed to match the names and order of the table, but rather just that one should match the other either way.

> Shall we keep the script in the root or move it to some subdirectory?

It seems cleaner to me to move it to an appropriate subdirectory (e.g. `_tools`, `_scripts`, etc), though a little less visible/convenient, but I have no strong opinion myself.

> I slightly prefer Makefile, these files won't need updating much, and it would be nice to avoid having to run this for every single make pages iteration when working on something else. 

Seems fair, so long as the script itself works on Windows and I can just execute it directly if needed. In that case, we of course need to run it in CI to verify the generated output matches is in sync with what is checked in, and that it works without errors (at least for any PRs/pushes touching the script, source or destination files, though since its likely very cheap to run, perhaps simpler to just do it unconditionally).

generate_release_cycle.py
```diff
@@ -0,0 +1,110 @@
+"""
+Read in a JSON and generate two CSVs and a Mermaid file
+"""
```
CAM-Gerlach, line 3:
```suggestion
"""Read in a JSON and generate two CSVs and a Mermaid file."""
```

Two CSVs, a JSON and a Mermaid walk into a bar...

(Trivial nit, but PEP 257, the docstring standard, requires docstring summary lines to end with a period (as they are complete sentences), and says that for one line docstrings, the starting and ending triple-quotes should be on the same line.)

generate-release-cycle.py
hugovk: Ah yes, I had missed the 4.0 case, thanks!

3.x->4.0 will be a good source of bugs just for version handling, https://github.com/asottile/flake8-2020 can help with some. And a common fix is to compare tuples not strings :)

CAM-Gerlach: Maybe you're right and we do need that `flake8-2020` in the pre-commit config after all :laughing:

hugovk:
Updated:

* Rename/reorder data keys
* Move under `_tools` and add to `make.bat`
* Test on CI with Windows and Linux
* PEP 257 docstrings

Re: CI

I made a new workflow as the existing workflow doesn't pass on Windows.

Test steps are:

1. Run the script to update the files.
2. Run `git status --porcelain` to show a list of changed files -> should be empty list.
3. To make the CI fail for changes, run `test $(git status --porcelain | wc -l) = 0`

This works for Linux but not Windows, I think because of line endings. I also added `git config core.eol lf` and `git config core.autocrlf true` to try and appease it, but no luck.

@CAM-Gerlach Any Windows tips here?

CAM-Gerlach:
`git config core.autocrlf true` will check the files out with CRLF, and the script (assuming it is working correctly) will check them in with LF. If that is indeed the issue here, then `git config core.autocrlf input` should fix it, which will normalize the line endings on commit but not on checkout. Using that locally on my Windows machine and following the steps in the CI workflow, everything works as expected and the test passes.

.github/workflows/release-cycle.yml
```diff
@@ -0,0 +1,34 @@
...
+      matrix:
+        os: [windows-latest, ubuntu-latest]
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Silence Windows warnings
+        run: |
+          git config core.eol lf
+          git config core.autocrlf true
```
CAM-Gerlach, line 22:
```suggestion
          git config core.autocrlf input
```

This *should* fix the test, to my knowledge and also per local testing

.github/workflows/release-cycle.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3"
+
+      - name: Generate release cycle output
+        run: python _tools/generate_release_cycle.py
```
CAM-Gerlach, line 29: For a more robust test run, I suggest this standard invocation:

```suggestion
        run: python -I -bb -X dev -X warn_default_encoding -W error _tools/generate_release_cycle.py
```

This will run in isolated mode, catch btyes-string issues, turn on warnings (e.g. deprecationwarnings, resourcewarnings, etc) and make them errors as well as a few other checks, and also warn on the implicit default encoding (which would have caught the various `open` calls missing `encoding` parameters).

.github/workflows/release-cycle.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Check for differences
+        run: |
+          git status --porcelain
```
CAM-Gerlach, line 33:
```suggestion
          git status
          git diff
```

Maybe output the full status, as well as the actual diff? This would make debugging issues (like this one) much easier, no?

CAM-Gerlach (commented):
Some comments on the new release cycle test workflow, otherwise seems LGTM

.github/workflows/release-cycle.yml
merwok: Will contributors / release managers using windows run into the same problem, and be required to have this same config?

CAM-Gerlach: Good question; to make a long story short, the answer is in fact no; this particular issue is essentially a "false positive" in the test we're using on CI. Read on for the detailed explanation...

----

The problem here was that because `core.autocrlf` was set to `true`, Git converted the LF EOLs to CRLF when the files were checked out on Windows. Then, when the script ran, it output the files with a consistent LF rather than the platform default, per my previous review suggestion. This causes the files in the working directory to appear as if they were modified. However, if `core.autocrlf` is set to `true`, Git will conform them back to LF when they are staged to the index, so the EOL won't actually change in the index or the repo. To note, if the test had done `git stage .` first, the test would have (correctly) passed.

Therefore, if Windows users have `core.autocrlf` set to `true`, the files will show as modified in the working tree, but once they do `git add .`, they will be conformed and no changes will be shown or committed (we _could_ prevent the minor display nit by reading the existing files, checking the EoL and outputing with that, but IMO the extra complexity that increases the chance of an _actual_ bug isn't worth it). If they have it set to `input` or `false`, the files will be checked out as LF, so there will be no changes at any point, since the script always outputs LF.

To note, there _would_ be an actual problem if the script used the default platform-specific newline output, and a Windows user had `core.autocrlf` set to `false` instead of `true` or `input`, but that was fixed in an earlier round of review by always using `\n` on output.

hugovk: Still fails with `git config core.autocrlf input`:
```
HEAD detached at pull/988/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   include/branches.csv
	modified:   include/end-of-life.csv
	modified:   include/release-cycle.mmd

no changes added to commit (use "git add" and/or "git commit -a")
```
https://github.com/python/devguide/actions/runs/3580813722/jobs/6023265680

> The problem here was that because core.autocrlf was set to true, Git converted the LF EOLs to CRLF when the files were checked out on Windows. 

Aha, we need to set the config _before_ the checkout! And with `--global` because we don't have a Git repo yet.

[`a87279d` (#988)](https://github.com/python/devguide/pull/988/commits/a87279dc382c040a828e7a044e52b930b6b9c35f)

https://github.com/hugovk/devguide/actions/runs/3580902248/jobs/6023437583

:tada: Thanks!

.github/workflows/release-cycle.yml
```diff
@@ -0,0 +1,35 @@
...
+          python-version: "3"
+
+      - name: Generate release cycle output
+        run: python -I -bb -X dev -X warn_default_encoding -W error _tools/generate_release_cycle.py
+
+      - name: Check for differences
+        run: |
+          git status
+          git diff
+          test $(git status --porcelain | wc -l) = 0
```
CAM-Gerlach, line 35:
```suggestion
          git add .
          git status
          git diff --staged
          test $(git status --porcelain | wc -l) = 0
```

_This_ should work, as I thought up in my reply to @merwok , since if files are checked out with `CRLF` due to `core.autocrlf true`, once staged to the index there will be no actual changes, since the LF EOL matches what is stored in the repository. (And if this somehow isn't the case, we perhaps _do_ have a potential broader problem we need to address).

hugovk: I think we don't need this after https://github.com/python/devguide/pull/988#discussion_r1035574945?

.github/workflows/release-cycle.yml
```diff
@@ -0,0 +1,35 @@
...
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest]
+
+    steps:
+      - name: Silence Windows warnings
+        run: |
+          git config --global core.eol lf
+          git config --global core.autocrlf input
+
```
CAM-Gerlach, line 21:
```suggestion
```

I didn't finish my review before you noticed the same issue, that the config was not getting set _before_ checkout, but instead of setting custom config, we can just do what local users will and actually check the staged index against the repo `HEAD` instead of the working tree against the index/repo `HEAD`. That will verify, per @merwok 's question, that users won't actually face a problem, and is also simpler and less finicky.

.github/workflows/release-cycle.yml
CAM-Gerlach: I had also figured out the cause, but determined per the discussion above that this is both a simpler and more robust solution, as it verifies that local users won't see a difference at least after staging.

.github/workflows/release-cycle.yml
CAM-Gerlach: Yeah, I'd realized that as well once the CI had failed, but was in the process of recommending a simpler and more robust solution (staging the files first) based on this discussion.

.github/workflows/release-cycle.yml
hugovk: Much cleaner, thank you!

https://github.com/python/devguide/actions/runs/3580947474/jobs/6023523022

CAM-Gerlach (approved):
LGTM now, thanks @hugovk !

PR #990, opened 2022-11-22
Add missing step to C API tests instructions

encukou:
Build definitions need to be updated -- see e.g.
https://github.com/python/cpython/pull/99549
https://github.com/python/cpython/pull/95819
https://github.com/python/cpython/pull/99494

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -122,6 +122,11 @@ To add a new set of tests (or extract a set out of the monolithic
 - Call the ``_PyTestCapi_Init_*`` from ``PyInit__testcapi`` in
   ``Modules/_testcapimodule.c``.
 
+- Add the new C file to :cpy-file:`Modules/Setup.stdlib.in`,
+  :cpy-file:`PCbuild/_testcapi.vcxproj` and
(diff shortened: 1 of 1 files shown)


developer-workflow/c-api.rst
```diff
@@ -122,6 +122,11 @@ To add a new set of tests (or extract a set out of the monolithic
 - Call the ``_PyTestCapi_Init_*`` from ``PyInit__testcapi`` in
   ``Modules/_testcapimodule.c``.
 
+- Add the new C file to :cpy-file:`Modules/Setup.stdlib.in`,
+  :cpy-file:`PCbuild/_testcapi.vcxproj` and
+  :cpy-file:`PCbuild/_testcapi.vcxproj.filters`,
```
erlend-aasland, line 127: Perhaps we should mention that these are `noeol` files?

encukou: Perhaps. But I'm not sure what that means in practice, on other people's computers...

erlend-aasland: I guess we can just leave it like this; Git warns you if you try to add newline at the end of these files anyways (see `.gitattributes` IIRC).

erlend-aasland (approved):

PR #992, opened 2022-11-26
Add Python favicon

hugovk:
Use the Python logo as favicon instead of the default [Read the Docs](https://readthedocs.org/) one.

Sphinx docs for `html_favicon`:

* https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon

# Preview

https://cpython-devguide--992.org.readthedocs.build/

Before and after, in Chrome, Firefox and Safari on macOS:

image


Mariatta (approved):
Seems to work on the preview.

CAM-Gerlach:
If we can only go with a single favicon (which rules out SVG thanks to Safari being backwards as usual), PNG support is [essentially universal](https://caniuse.com/link-icon-png), so I recommend a 192 px PNG.

hugovk:
Updated to 192px PNG!

Sphinx [says](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon) "It should be a Windows-style icon file (`.ico`), which is 16x16 or 32x32 pixels large" but it doesn't seem to matter.

Also "The image file will be copied to the `_static` directory of the output HTML, but only if the file does not already exist there", so let's put it there already. Keeps the root tidier too.

---

If we wanted to cover a lot of the possible options (to see them, go to https://realfavicongenerator.net/ and "Demo with a random image"), we could look at https://pypi.org/project/sphinx-favicon/

CAM-Gerlach:
> If we wanted to cover a lot of the possible options (to see them, go to https://realfavicongenerator.net/ and "Demo with a random image"), we could look at https://pypi.org/project/sphinx-favicon/

Yeah, I thought about that, I just figured it wasn't worth it for now just for the devguide

CAM-Gerlach (approved):
LGTM, thanks @hugovk !

ezio-melotti (approved):

PR #994, opened 2022-11-29
cpy-file: Go direct to the GitHub page and avoid a redirect

hugovk:
# Steps to reproduce

1. Visit https://devguide.python.org/internals/parser/
2. Find and click one of the links that use the `:cpy-file:` role, for example:

> The generator is located at [`Tools/peg_generator/pegen`](https://github.com/python/cpython/blob/main/Tools/peg_generator/pegen).

# Actual result

Takes you to:

`https://github.com/python/cpython/blob/main/Tools/peg_generator/pegen`

And then redirects to:

`https://github.com/python/cpython/tree/main/Tools/peg_generator/pegen`

# Expected result

Go straight to:

`https://github.com/python/cpython/tree/main/Tools/peg_generator/pegen`

# Notes

This is a recent GitHub change.

One redirect isn't a big deal, but we might as well save a couple of hundred milliseconds and be direct.

### DIFF
--- a/_extensions/custom_roles.py
+++ b/_extensions/custom_roles.py
@@ -11,7 +11,7 @@ def setup(app):
     # role to link to cpython files
     app.add_role(
         "cpy-file",
-        autolink("https://github.com/python/cpython/blob/main/{}"),
+        autolink("https://github.com/python/cpython/tree/main/{}"),
     )
     # role to link to cpython labels
     app.add_role(


encukou:
Isn't that only for directories? `https://github.com/python/cpython/tree/main/Tools/peg_generator/pegen/ast_dump.py` redirects the other way.

zware:
How do you tell the difference between a file (`blob`) and a directory (`tree`) from just the name?  AFAICT, here you're trading a redirect in one case for a redirect in another.  To have no redirects, you'd need a separate `:cpy-dir:` role.

hugovk:
Ah yes, I didn't spot it's only for directories... Thanks for pointing it out!

I don't think it's worth making a second role just to save a redirect.

If we keep a single role, we should pick whichever of file/directory is more common, and I think it's a file?

So we can probably just close this.

ezio-melotti (commented):
Agree on closing.

PR #995, opened 2022-12-04
Update name to macOS

hugovk:
Apple's operating system has been called [macOS since 2016](https://en.wikipedia.org/wiki/MacOS#Release_history). Before that it was OS X, and before that it was Mac OS X.

In most cases we're referring to the recent versions, so let's use the current name.

---

We have [one section](https://devguide.python.org/getting-started/setup-building/#macos) specifically with two sets of instructions:

* older set: Mac OS X (versions 10.0-10.7) and OS X 10.8
* newer set: OS X 10.9-10.11 and macOS (versions 10.12+)

It's good to differentiate the names here. But whilst we're reviewing this, are the instructions for the older set still useful?

* Mac OS X versions 10.0-10.7 were released between 2001-2012
* OS X 10.8 was released between 2012-2015

Similarly, do we still need to mention OS X 10.9+ in the newer set?

* OS X versions 10.9-10.11 were released between 2013-2018

Mac OS X and OS X are all long out of support.

---

Finally, I noticed the link from https://devguide.python.org/#key-resources to the daily installer goes to https://buildbot.python.org/daily-dmg/, but it's an empty directory listing.

Should there be an installer there or shall we remove the link?

---

cc @ronaldoussoren @ned-deily

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -209,7 +209,7 @@ If you decide to :ref:`build-dependencies`, you will need to re-run both
 Once CPython is done building you will then have a working build
 that can be run in-place; ``./python`` on most machines (and what is used in
 all examples), ``./python.exe`` wherever a case-insensitive filesystem is used
-(e.g. on OS X by default), in order to avoid conflicts with the ``Python``
+(e.g. on macOS by default), in order to avoid conflicts with the ``Python``
 directory. There is normally no need to install your built copy
 of Python! The interpreter will realize where it is being run from
 and thus use the files found in the working copy.  If you are worried
@@ -326,7 +326,7 @@ Install dependencies
 ====================
 
 This section explains how to install additional extensions (e.g. ``zlib``)
-on :ref:`Linux <deps-on-linux>` and :ref:`macOs/OS X <macOS>`.  On Windows,
+on :ref:`Linux <deps-on-linux>` and :ref:`macOS`.  On Windows,
 extensions are already included and built automatically.
 
 .. _deps-on-linux:
@@ -385,7 +385,7 @@ their dependencies::
          lzma lzma-dev tk-dev uuid-dev zlib1g-dev
 
 
-.. _MacOS:
+.. _macOS:
 
 macOS and OS X
 --------------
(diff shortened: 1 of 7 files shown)


ezio-melotti (approved):

getting-started/setup-building.rst
```diff
@@ -385,7 +385,7 @@ their dependencies::
          lzma lzma-dev tk-dev uuid-dev zlib1g-dev
 
 
-.. _MacOS:
+.. _macOS:
```
arhadthedev, line 388: Can we keep both the old and the new anchor for potentially existing hyperlinks? AFAIK URL paths and anchors are case sensitive.

Mariatta (approved):

index.rst
```diff
@@ -190,7 +190,7 @@ Key Resources
 * Source code
     * `Browse online <https://github.com/python/cpython/>`_
     * `Snapshot of the *main* branch <https://github.com/python/cpython/archive/main.zip>`_
-    * `Daily OS X installer <https://buildbot.python.org/daily-dmg/>`_
+    * `Daily macOS installer <https://buildbot.python.org/daily-dmg/>`_
```
ned-deily, line 193: We should just delete this. That buildbot hasn't existed for years.

ned-deily (changes requested):

index.rst
hugovk: Thanks, deleted!

ned-deily (approved):

hugovk:
> We have [one section](https://devguide.python.org/getting-started/setup-building/#macos) specifically with two sets of instructions:
> 
> * older set: Mac OS X (versions 10.0-10.7) and OS X 10.8
> * newer set: OS X 10.9-10.11 and macOS (versions 10.12+)
> 
> It's good to differentiate the names here. But whilst we're reviewing this, are the instructions for the older set still useful?
> 
> * Mac OS X versions 10.0-10.7 were released between 2001-2012
> * OS X 10.8 was released between 2012-2015
> 
> Similarly, do we still need to mention OS X 10.9+ in the newer set?
> 
> * OS X versions 10.9-10.11 were released between 2013-2018
> 
> Mac OS X and OS X are all long out of support.

Please see PR https://github.com/python/devguide/pull/1006 to remove these old bits.

PR #1001, opened 2022-12-07
Remove reference to unused Zulip chat

hugovk:
For the second point of https://github.com/python/core-workflow/issues/457.

Zulip is no longer used and there's very little activity:

image

The little activity that remains is unanswered questions, such as:

image

There are better places to ask for help, let's remove Zulip references from the devguide.

Ping also @Mariatta as it referred to your Office Hour, but I see there was no activity for a few years so I think we're fine to remove it, unless you'd like to replace it with something else.

### DIFF
--- a/getting-started/getting-help.rst
+++ b/getting-started/getting-help.rst
@@ -73,23 +73,6 @@ whereas ``#python`` is for questions concerning development *with* Python.
 
 .. _Libera.Chat: https://libera.chat/
 
-Zulip
------
-
-An alternative to IRC is our own `Zulip`_ instance.  There are different streams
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

CAM-Gerlach (approved):
LGTM, thanks @hugovk

ezio-melotti (approved):
IIRC @Mariatta is still doing office hours, so maybe just remove the Zulip mention (unless she says otherwise).

hugovk:
I checked with Mariatta via DM, she's no longer doing office hours so we can remove it.

PR #1006, opened 2022-12-12
Remove old Mac OS X (10.0-10.7) and OS X (10.8-10.11) info

hugovk:
Follow on from https://github.com/python/devguide/pull/995.

---

Apple currently supports macOS 11 (released Nov 2020), 12 (Oct 2021) and 13 (Oct 2022).

The older "Mac OS X" and "OS X" are all long out of support.

* https://en.wikipedia.org/wiki/MacOS#Release_history

---

We have [one section](https://devguide.python.org/getting-started/setup-building/#macos) specifically with two sets of instructions:

* older set: Mac OS X (versions 10.0-10.7) and OS X (10.8)
* newer set: OS X (10.9-10.11) and macOS (versions 10.12+)

It's good to differentiate the names here, but let's remove the instructions for the older set:

* Mac OS X versions 10.0-10.7 were released between 2001-2012
* OS X 10.8 was released between 2012-2015

And also for OS X in the newer set:

* OS X versions 10.9-10.11 were released between 2013-2018

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -145,7 +145,7 @@ checks that one should not skip.
 
 .. _unix-compiling:
 
-UNIX
+Unix
 ----
 
 The core CPython interpreter only needs a C compiler to be built,
@@ -334,7 +334,7 @@ extensions are already included and built automatically.
 Linux
 -----
 
-For UNIX based systems, we try to use system libraries whenever available.
+For Unix-based systems, we try to use system libraries whenever available.
 This means optional components will only build if the relevant system headers
 are available. The best way to obtain the appropriate headers will vary by
 distribution, but the appropriate commands for some popular distributions
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -385,12 +385,13 @@ their dependencies::
          lzma lzma-dev tk-dev uuid-dev zlib1g-dev
 
 
+.. _macOS and OS X:
 .. _macOS:
 
-macOS and OS X
---------------
+macOS
+-----
 
-For **macOS systems** (versions 10.12+) and **OS X 10.9 and later**,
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -385,12 +385,13 @@ their dependencies::
          lzma lzma-dev tk-dev uuid-dev zlib1g-dev
 
 
+.. _macOS and OS X:
 .. _macOS:
```
CAM-Gerlach, line 389: Nice trick, and I can confirm it works :+1:

getting-started/setup-building.rst
```diff
@@ -466,10 +457,10 @@ and ``make``::
     $ make -s -j2
 
 There will sometimes be optional modules added for a new release which
-won't yet be identified in the OS level build dependencies. In those cases,
+won't yet be identified in the OS-level build dependencies. In those cases,
 just ask for assistance on the core-mentorship list.
 
-Explaining how to build optional dependencies on a UNIX based system without
+Explaining how to build optional dependencies on a Unix-based system without
```
CAM-Gerlach, line 463: It's referred to the uppercase UNIX the other two times it is mentioned on the page, so IMO it would be better to either change all of them or none of them in this PR, rather than be inconsistent.

CAM-Gerlach (commented):
A few minor comments, other wise LGTM, thanks

getting-started/setup-building.rst
hugovk: Yep, updated the other two to Unix.

Unix is in our style guide: https://devguide.python.org/documentation/style-guide/#capitalization

CAM-Gerlach (approved):
LGTM now, FWIW

ned-deily (approved):
LGTM, thanks!

PR #1011, opened 2022-12-24
Add translation (l10n) infrastructure to devguide

rffontenelle:
See #1012

### DIFF
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,31 @@
+name: Translations
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  translations:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3
+          cache: pip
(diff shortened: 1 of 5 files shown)


CAM-Gerlach:
See [my comment in #1012](https://github.com/python/devguide/issues/1012#issuecomment-1364586917) for some overall questions about this.

My web browser hangs trying to load the full 20k+ line file in GitHub, but it looks like there are at least a couple Sphinx issues in the translated version:

```
locales/pt_BR/LC_MESSAGES/devguide.po:9140: superfluous backtick in front of role (backtick-before-role)
locales/pt_BR/LC_MESSAGES/devguide.po:9140: default role used (hint: for inline literals, use double backticks) (default-role)
```

It seems we need to enable or add support for the appropriate theme options to add a translation menu or similar in the theme, and add support on our hosting serve (RTD?), as I'm not seeing the translated versions anywhere in the rendered preview.

Also, note that the theme text is unfortunately not yet itself translated, see pradyunsg/furo#434 for more.

rffontenelle:
Thanks @CAM-Gerlach. #1014 should fix the root cause of the Sphinx issue.

rffontenelle:
Closing as commented in the linked issue.

PR #1015, opened 2022-12-26
Move 'moved to' link to the end

hugovk:
These "Moved to [another-page#another-anchor]" links are a convenience to avoid linkrot:

image

---

Like at https://devguide.python.org/documentation/start-documenting/#style-guide, let's put them at end of the page.

So when reading through a page, the reader is not disturbed by a notice that some content has moved.

---

Before: https://devguide.python.org/#status-of-python-branches

After: https://cpython-devguide--1015.org.readthedocs.build/#status-of-python-branches

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -89,12 +89,6 @@ contributing to Python:
 * PEPs_ (Python Enhancement Proposals)
 * :ref:`gitbootcamp`
 
-Status of Python branches
--------------------------
-
-Moved to :ref:`versions`
-
-
 .. _contributing:
 
 Contributing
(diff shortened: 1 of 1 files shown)


ezio-melotti (commented):
I wonder if it would be better to put it before the full table of contents, since it looks a bit out of place having it after. :thinking:

CAM-Gerlach:
Agreed with @ezio-melotti , I guess; otherwise seems LGTM

ezio-melotti (approved):

hugovk:
Moved!

CAM-Gerlach (approved):
LGTM, thanks @hugovk

PR #1016, opened 2022-12-29
Add Doc link to Turkish section

egeakman:
Add docs.python.org link to the Turkish section in the table.


documentation/translating.rst
```diff
@@ -98,6 +99,7 @@ in production; others are works in progress.
 .. _doc_ja: https://docs.python.org/ja/
 .. _doc_ko: https://docs.python.org/ko/
 .. _doc_pl: https://docs.python.org/pl/
+.. _doc_tr: https://docs.python.org/tr
```
ezio-melotti, line 102:
```suggestion
.. _doc_tr: https://docs.python.org/tr/
```
I also noticed that this page doesn't redirect to `/tr/3/` like the other 3 before it, so this should be fixed too.

egeakman: I think it will be getting redirected once https://github.com/python/psf-salt/pull/289 gets merged.

PR #1021, opened 2023-01-05
Add hauntsaninja as tomllib maintainer

hauntsaninja:


AlexWaygood (approved):

hauntsaninja:
https://github.com/python/cpython/pull/100779

PR #1023, opened 2023-01-06
Add barneygale to experts index

barneygale:
As @AlexWaygood [suggested](https://github.com/python/cpython/pull/100782#issuecomment-1373563789)!


barneygale:
Ack, wrong git email address attached to the commit. Stand by...

Issue #1027, opened 2023-01-14
Garbage collector generation threshold 


zikcheng:
In post https://github.com/python/devguide/blob/main/internals/garbage-collector.rst, generation threshold is described as follwing:

> Generations are collected when the number of objects that they contain reaches some predefined threshold, which is unique for each generation and is lower the older the generations are.

It sounds like generation 0 starts when there are 700 objects in it, generation 1 starts when there are 10 objects in it and generation 2 starts when there are 10 objects in it.

This is very different from the fact(at least for generation 1) which https://docs.python.org/3/library/gc.html#gc.set_threshold describes:

> If generation 0 has been examined more than threshold1 times since generation 1 has been examined, then generation 1 is examined as well. With the third generation, things are a bit more complicated, see [Collecting the oldest generation](https://devguide.python.org/garbage_collector/#collecting-the-oldest-generation) for more information.

Maybe I am too literal, but I think it's really misleading.

pablogsal:
Thanks, @zikcheng for opening the issue!

Indeed, the current way this is described is confusing and incorrect. I will update the text so it matches the documentation.

Thanks for noticing the error!

PR #1030, opened 2023-01-14
Correct the description of generation thresholds in the gc docs

pablogsal:
Closes: #1027

### DIFF
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -324,7 +324,7 @@ allocations and deallocations since the last collection. When the number of
 allocations minus the number of deallocations exceeds ``threshold_0``,
 collection starts. Initially only generation 0 is examined. If generation 0 has
 been examined more than ``threshold_`` times since generation 1 has been
-examined, then generation 1 is examined as well. With the third generation,
+examined, then generation 1 is examined as well. With generation 2,
 things are a bit more complicated, see :ref:`gc-oldest-generation` for
 more information. These thresholds can be examined using the
 :func:`gc.get_threshold` function:
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -325,7 +325,7 @@ allocations minus the number of deallocations exceeds ``threshold_0``,
 collection starts. Initially only generation 0 is examined. If generation 0 has
 been examined more than ``threshold_`` times since generation 1 has been
 examined, then generation 1 is examined as well. With the third generation,
-things are a bit more complicated, see `Collecting the oldest generation`_ for
+things are a bit more complicated, see :ref:`gc-oldest-generation` for
 more information. These thresholds can be examined using the
 :func:`gc.get_threshold` function:
 
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -374,6 +374,7 @@ specifically in a generation by calling ``gc.collect(generation=NUM)``.
     [..., <__main__.MyObj object at 0x7fbcc12a3400>, ...]
 
 
+.. _gc-oldest-generation:
 
 Collecting the oldest generation
 --------------------------------
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -323,7 +323,7 @@ In order to decide when to run, the collector keeps track of the number of objec
 allocations and deallocations since the last collection. When the number of
 allocations minus the number of deallocations exceeds ``threshold_0``,
 collection starts. Initially only generation 0 is examined. If generation 0 has
-been examined more than ``threshold_`` times since generation 1 has been
+been examined more than ``threshold_1`` times since generation 1 has been
 examined, then generation 1 is examined as well. With generation 2,
 things are a bit more complicated, see :ref:`gc-oldest-generation` for
 more information. These thresholds can be examined using the
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -319,10 +319,15 @@ the same object survives another GC round in this new generation (generation 1)
 it will be moved to the last generation (generation 2) where it will be
 surveyed the least often.
 
-Generations are collected when the number of objects that they contain reaches some
-predefined threshold, which is unique for each generation and is lower the older
-the generations are. These thresholds can be examined using the  ``gc.get_threshold``
-function:
+In order to decide when to run, the collector keeps track of the number object
+allocations and deallocations since the last collection. When the number of
+allocations minus the number of deallocations exceeds ``threshold_0``,
+collection starts. Initially only generation 0 is examined. If generation 0 has
+been examined more than ``threshold_`` times since generation 1 has been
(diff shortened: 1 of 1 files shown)


internals/garbage-collector.rst
```diff
@@ -319,10 +319,15 @@ the same object survives another GC round in this new generation (generation 1)
...
-function:
+In order to decide when to run, the collector keeps track of the number object
+allocations and deallocations since the last collection. When the number of
+allocations minus the number of deallocations exceeds ``threshold_0``,
+collection starts. Initially only generation 0 is examined. If generation 0 has
+been examined more than ``threshold_`` times since generation 1 has been
+examined, then generation 1 is examined as well. With the third generation,
+things are a bit more complicated, see `Collecting the oldest generation`_ for
+more information. These thresholds can be examined using the
+``gc.get_threshold`` function:
```
CAM-Gerlach, line 330:
```suggestion
:func:`gc.get_threshold` function:
```

You can directly cross-reference this, since Intersphinx is enabled here for the CPython docs

internals/garbage-collector.rst
```diff
@@ -319,10 +319,15 @@ the same object survives another GC round in this new generation (generation 1)
...
-Generations are collected when the number of objects that they contain reaches some
-predefined threshold, which is unique for each generation and is lower the older
-the generations are. These thresholds can be examined using the  ``gc.get_threshold``
-function:
+In order to decide when to run, the collector keeps track of the number object
+allocations and deallocations since the last collection. When the number of
+allocations minus the number of deallocations exceeds ``threshold_0``,
+collection starts. Initially only generation 0 is examined. If generation 0 has
+been examined more than ``threshold_`` times since generation 1 has been
+examined, then generation 1 is examined as well. With the third generation,
```
CAM-Gerlach, line 327: Does "the third generation" refer to "generation 2" here? Or what happens to the second generation? At least to me, this seems like this could be a bit confusing...

internals/garbage-collector.rst
```diff
@@ -319,10 +319,15 @@ the same object survives another GC round in this new generation (generation 1)
...
+things are a bit more complicated, see `Collecting the oldest generation`_ for
```
CAM-Gerlach, line 328:
```suggestion
things are a bit more complicated, see :ref:`gc-oldest-generation` for
```

Per best practice and standard docs/devguide convention, `:ref:` is used for referencing sections instead of old-style implicit hyperlinks.

You'll also need to add a ref target label before the section heading:

```rst
.. _gc-oldest-generation:

Collecting the oldest generation
--------------------------------
```

CAM-Gerlach (commented):
Seems like a big improvement overall, thanks!

Assuming you apply the suggestions (e.g. with [`Files changes`](https://github.com/python/devguide/pull/1030/files) -> `Add to batch` -> `Commit`), you'll need to add a new ref target before the desired section heading:

```rst
.. _gc-oldest-generation:

Collecting the oldest generation
--------------------------------
```

internals/garbage-collector.rst
pablogsal: We can say "generation 2", but I would say is clear from the previous number that the generation 0 is the first and generation 1 is the 2nd

internals/garbage-collector.rst
pablogsal: I am afraid I am not familiar with this practice, what's the rationale for this?

CAM-Gerlach: This is the standard way to reference sections in Sphinx and the CPython docs See the [quickref](https://devguide.python.org/documentation/markup/#quick-reference) and the [internal links section](https://devguide.python.org/documentation/markup/#internal-links) of the reST Markup page in the devguide itself, and the [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref) for full details.

Unlike legacy docutils-style implicit internal hyperlinks, they work reliably instead of silently breaking if section titles are changed at all, either the source or the target is moved to another page, or two sections have the same title, for instance—and in case they do ever break or are incorrect, they will raise a warning on build and appear unlinked in the rendered text instead of failing silently for both. They also automatically insert and update the section title, and they work consistently across all builders. Finally, they are standard convention in the CPython docs and particularly the devguide: here, `:ref:` is used 216 times, compared to only 6 (in 4 in 1 file, plus 1 one-offs) implicit section hyperlinks.

internals/garbage-collector.rst
CAM-Gerlach: Perhaps I'm just a n00b, but I sure was more than a little confused by the abrupt and unexplained jump from 0-indexed cardinal numbers to 1-indexed ordinal numbers. Absent a strong reason otherwise, it would seem clearly preferable to be consistent and avoid the obvious potential for confusion and ambiguity here, particularly in a technical explanation aimed at readers not already well-versed —as the old saying goes, there are two hard problems in programming: cache invalidation, naming things and off by one errors...

As such, I'd suggest

```suggestion
examined, then generation 1 is examined as well. With generation 2,
```

internals/garbage-collector.rst
pablogsal: Great, thanks for the explanation :+1:

CAM-Gerlach: Of course, glad it was helpful!

internals/garbage-collector.rst
```diff
@@ -323,7 +323,7 @@ In order to decide when to run, the collector keeps track of the number of objec
 allocations and deallocations since the last collection. When the number of
 allocations minus the number of deallocations exceeds ``threshold_0``,
 collection starts. Initially only generation 0 is examined. If generation 0 has
-been examined more than ``threshold_`` times since generation 1 has been
+been examined more than ``threshold_1`` times since generation 1 has been
```
CAM-Gerlach, line 326: I was wondering about this one...I should have spoken up, sorry :)

internals/garbage-collector.rst
```diff
@@ -319,10 +319,15 @@ the same object survives another GC round in this new generation (generation 1)
...
-predefined threshold, which is unique for each generation and is lower the older
-the generations are. These thresholds can be examined using the  ``gc.get_threshold``
-function:
+In order to decide when to run, the collector keeps track of the number of object
+allocations and deallocations since the last collection. When the number of
+allocations minus the number of deallocations exceeds ``threshold_0``,
+collection starts. Initially only generation 0 is examined. If generation 0 has
+been examined more than ``threshold_1`` times since generation 1 has been
+examined, then generation 1 is examined as well. With generation 2,
+things are a bit more complicated, see :ref:`gc-oldest-generation` for
```
CAM-Gerlach, line 328:
```suggestion
things are a bit more complicated; see :ref:`gc-oldest-generation` for
```

Oops, missed this grammar error (comma splice)

CAM-Gerlach (approved):
One grammar issue I missed, otherwise LGTM, thanks @pablogsal !

CAM-Gerlach (approved):

Issue #998, opened 2022-12-06
Add the other dev phase dates to Status of Python versions


CAM-Gerlach:
As [discussed] on PR #988 and at the Python Docs Community meeting, the [Status of Python versions](https://devguide.python.org/versions/) table is missing an indication of the expected/actual cut-over point between bugfix and security-only releases, which is fairly important information in line with the other dates displayed there. This is standard for most pages like this, and the [Wikipedia page](https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions) has it.

Additionally, it would seem to be quite helpful, especially since this is the CPython _developer_ guide, to also add the dates for the dev cycle start, alpha, beta and rc phases, to make it easier for devs to quickly check the current phase and when the next one starts.

We should, at least, add additional JSON fields with this information. Since unfortunately with Mermaid it doesn't seem possible to add multiple segments to the same line, and it would clutter the main chart and table to add all these dates, I suggest we output these dates to a separate chart and table in a "Development Status" section, leaving the main one at the top. Eventually, it would seem ideal to migrate the main chart and table (or at least a simplified version) to the main Python,org site, since it is focused on user-facing information, while the proposed new section would contain developer-facing information

#999 is a followup, adding the individual point release version dates as JSON and subsidiary tables. These could be added to the chart in the development status section (and perhaps the latest release added to one of the main tables), or possibly the more detailed charts could be added/moved to the per-version subsections.

m-aciek:
Translation teams (and docs building team) are interested especially in first beta date as it used to be point when they (we) start translating the documentation for a release.

nedbat:
One tweak to show the transition from bugfix to security: https://github.com/python/devguide/pull/1529

Issue #999, opened 2022-12-06
Add micro (pre-)release dates to JSON and as tables


CAM-Gerlach:
Followup to #998 (which would be implemented first, since it is more important and lower-hanging fruit).

As [discussed] on PR #988 and at the Python Docs Community meeting, it would be helpful to the release date of each micro release/pre-elease to the [Status of Python versions](https://devguide.python.org/versions/) JSON and page. This would be useful as both exposing this information, currently manually listed in each release PEP, in a machine-readable form, as well as generating standard, easily skimmable human-readable tables linked from the main version table and all on one page.

If this works out, we could then have the release PEPs just link the appropriate section on this page, or automatically inject the table via the same `csv-table` directive as used here, avoiding the need for RMs/others to update multiple places

 Someone even suggesting adding a small dot to the chart for each version release, though we'd have to see if it makes it too cluttered.

PR #1034, opened 2023-01-19
Generate the release cycle chart directly as SVG

encukou:
How's this for procrastination:
image

Please amend this PR if you want (and can); I might not get to procrastinate more soon.

Instead of Mermaid, use jinja2. Sphinx uses jinja for templating as well, so it might be possible to integrate this more tightly, but an include file works nicely for now.
(It's actually easy to generate this chart just with f-strings, but I don't want to set a bad example for people who shouldn't fully trust their input. Jinja has autoescaping to prevent SVG injections.)

Styling is done from the main stylesheet, and is bit more straightforward. I've adjusted the colours to be a bit friendlier to colour-blind people.

The labels on the boxes aren't the greatest, but I expect them to go away after #998 -- ideally replaced by a legend at the bottom.

There are vertical lines for all years now -- Mermaid's skipping of every other year was pretty confusing. Year labels have been shortened. This should work for another 10 years.

The caption is removed; it was redundant in our case.

Precise sizing is hard to do with SVG, but the font family, size and line height should nearly match the main text on big screens. At least in the current theme.

In the code, `sorted_versions` is now a list, and the dicts in it have some extra generated info.


encukou:
The current date line's position is now encoded in the SVG, rather than drawn by Javascript. With commits coming in every few weeks, it shouldn't matter. And if #998 gets in it'll ensure a commit every month or so.
It's possible to add cron to GitHub Actions, but  I don't think it's worth it.

CAM-Gerlach (commented):
IMO, while this is a lot of work and I'm sure you've done an excellent technical job, I would appreciate seeing a clear and concrete rationale for getting rid of the existing working mermaid approach and going with a handwritten SVG, given the amount of bespoke complexity it requires reviewing and maintaining ourselves indefinitely.

As to some possible reasons:

* It seems hard to justify this with the minuscule fraction of users who browse with JS fully disabled (those numbers often include those using NoScript, that can easily selectively enable it for things they want, which the `<noscript>` message makes clear), and given the same information is available in multiple other human and machine-readable forms.
* Making the SVG sharable and embeddable seems a somewhat more compelling reason, but right now it's just being injected directly into the page at build time (otherwise it couldn't be styled with the CSS), so in practice it doesn't appear any more easily sharable or embeddable than it was before.
* There are some minor improvements to the appearance, but also a few rough edges that need polishing, and perhaps some of them could be made to the Mermaid version too.

Besides the complexity/maintenance issues mentioned above, unless and until #998 and #999 are implemented, it will only be updated ≈twice a year when the branches change to a different major status (and even with #998, it only adds a couple more updates; #999 would be required to update it monthly). I'm also not totally sure about the complexity of incorporating the information in #998, let alone #999 into it, as hand-writing SVG is not really in my wheelhouse.

That said, if we do go ahead with it, there are some things I noticed that should be fixed:

* [x] The SVG template should be given an extension other than `SVG` (or similar), or configured with the appropriate `.gitattributes`, as this currently makes it impossible to easily view it or the changes to it on GitHub due to it attempting and failing to render as an SVG image.
* [x] The "current date" line is shown behind all of release bars, making it much harder to see than the original. Could it be moved in front?
* [x] To me, the inset text (particularly "security") actually looks harder to read than before with the drop shadow, even with the larger font size.

CAM-Gerlach:
For the record, followup to #997

hugovk:
# Benefits

## Better rendering

This new one renders much better on mobile:

<table>
<tr>
    <th>Before
    <th>After
<tr>
	<td>
	<td>
</table>

Plus when rotating from portrait to landscape, no need to reload the page for it to re-render:

<table>
<tr>
	<td>
	<td>
</table>

## Better performance

> * It seems hard to justify this with the minuscule fraction of users who browse with JS fully disabled (those numbers often include those using NoScript, that can easily selectively enable it for things they want, which the `<noscript>` message makes clear), and given the same information is available in multiple other human and machine-readable forms.

This also benefits people with JavaScript enabled.

It removes a 241 KB (compressed, 878K uncompressed) JS library resource.

This makes the page and the site load much quicker.

Currently the Mermaid JS is added to every page on the site, when we only need it for one page (I have an open PR to fix this: https://github.com/mgaitan/sphinxcontrib-mermaid/pull/101).

It's also quicker because instead of: 
* loading JS
* reading page data
* generating SVG via JS
* rendering SVG

we're now:
* loading static SVG
* rendering SVG

To get some numbers, compare Google's PageSpeed Insights (aka Lighthouse):

* https://pagespeed.web.dev/report?url=https%3A%2F%2Fdevguide.python.org%2Fversions%2F
* https://pagespeed.web.dev/report?url=https%3A%2F%2Fcpython-devguide--1034.org.readthedocs.build%2Fversions%2F

<table>
<tr>
	<td>image
	<td>image
</table>

This reduces: Time to Interactive, Blocking Time, Largest Contentful Paint

The time spent evaluating JS is down from 1,167 ms -> 228 ms

This "treemap" shows there's a lot less JS involved:

<table>
<tr>
	<td>image
	<td>
image
</table>

# Other things

> * Making the SVG sharable and embeddable seems a somewhat more compelling reason, but right now it's just being injected directly into the page at build time (otherwise it couldn't be styled with the CSS), so in practice it doesn't appear any more easily sharable or embeddable than it was before.

Sure, but no regression. Same also applies to the JSON. We can export them to a shareable location if we want (although we'd need to move some colours from CSS to the SVG for it to be standalone). And injecting into the page also removes an HTTP request.

> * There are some minor improvements to the appearance, but also a few rough edges that need polishing, and perhaps some of them could be made to the Mermaid version too.

👍 


# Complexity / maintenance

I'm going to be optimistic, it doesn't look too tricky? It's "just" some lines, rectangles and text boxes.

# Regular updates

> Besides the complexity/maintenance issues mentioned above, unless and until https://github.com/python/devguide/issues/998 and https://github.com/python/devguide/issues/999 are implemented, it will only be updated ≈twice a year when the branches change to a different major status (and even with https://github.com/python/devguide/issues/998, it only adds a couple more updates; https://github.com/python/devguide/issues/999 would be required to update it monthly). I'm also not totally sure about the complexity of incorporating the information in https://github.com/python/devguide/issues/998, let alone https://github.com/python/devguide/issues/999 into it, as hand-writing SVG is not really in my wheelhouse.

It does seem a problem that the "now" line is static and only rarely updated. But can we re-generate the SVG during each deploy? Possibly de-version from Git, similar to the RSS in the PEPs repo. This repo is deployed several times a week.

Re: #998 / #999, the [Mermaid Gantt chart](https://mermaid.js.org/syntax/gantt.html) doesn't support additional sections in the same line. As noted in #998, that may also make the chart too cluttered, but we could try it out. If we decide to add extra charts instead, they shouldn't be too difficult, and templating can help if there's more than one.

.gitattributes
```diff
@@ -0,0 +1,11 @@
+# Generated files
```
hugovk, line 1: I'd prefer to be able to see changes in these, as a sanity check rather, than them being opaque black boxes.

# Before

```diff
diff --git a/include/branches.csv b/include/branches.csv
index bfec9da..fbf0153 100644
--- a/include/branches.csv
+++ b/include/branches.csv
@@ -1,5 +1,5 @@
 Branch,Schedule,Status,First release,End of life,Release manager
-main,:pep:`693`,feature,*2023-10-02*,*2028-10*,Thomas Wouters
+main,:pep:`693`,feature,*2023-10-02*,*2028-11*,Thomas Wouters
 3.11,:pep:`664`,bugfix,2022-10-24,*2027-10*,Pablo Galindo Salgado
 3.10,:pep:`619`,bugfix,2021-10-04,*2026-10*,Pablo Galindo Salgado
 3.9,:pep:`596`,security,2020-10-05,*2025-10*,Łukasz Langa
diff --git a/include/end-of-life.csv b/include/end-of-life.csv
index be9fe1b..f2aa404 100644
--- a/include/end-of-life.csv
+++ b/include/end-of-life.csv
@@ -1,5 +1,5 @@
 Branch,Schedule,Status,First release,End of life,Release manager
-3.6,:pep:`494`,end-of-life,2016-12-23,2021-12-23,Ned Deily
+3.6,:pep:`494`,end-of-life,2016-12-23,2021-12-31,Ned Deily
 3.5,:pep:`478`,end-of-life,2015-09-13,2020-09-30,Larry Hastings
 3.4,:pep:`429`,end-of-life,2014-03-16,2019-03-18,Larry Hastings
 3.3,:pep:`398`,end-of-life,2012-09-29,2017-09-29,"Georg Brandl, Ned Deily (3.3.7+)"
diff --git a/include/release-cycle.svg b/include/release-cycle.svg
index 53714a1..65eed8b 100644
--- a/include/release-cycle.svg
+++ b/include/release-cycle.svg
@@ -1011,7 +1011,7 @@
             y="405.0"
             font-size="18"
         >
-            Python 3.12
+            Python 3.20
         </text>
 
         <!-- Colourful blob with a label -->
```

# After

```diff
diff --git a/include/branches.csv b/include/branches.csv
index bfec9da..fbf0153 100644
Binary files a/include/branches.csv and b/include/branches.csv differ
diff --git a/include/end-of-life.csv b/include/end-of-life.csv
index be9fe1b..f2aa404 100644
Binary files a/include/end-of-life.csv and b/include/end-of-life.csv differ
diff --git a/include/release-cycle.svg b/include/release-cycle.svg
index 53714a1..65eed8b 100644
Binary files a/include/release-cycle.svg and b/include/release-cycle.svg differ
```

hugovk (commented):
A few suggestions.

(And okay to retain Black formatting on this file?)

_tools/generate_release_cycle.py
```diff
@@ -70,38 +56,84 @@ def write_csv(self) -> None:
...
+        # those.)
+        SCALE = 18
+
+        # Width of the drawing and main parts
+        DIAGRAM_WIDTH = 46
+        LEGEND_WIDTH = 7
+        RIGHT_MARGIN = 0.5
+
+        # Height of one line. If you change this you'll need to tweak
+        # some positioning nombers in the template as well.
```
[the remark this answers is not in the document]
CAM-Gerlach: [Nom Nom Nom!](https://www.youtube.com/watch?v=SMWi7CLoZ2Q)

.github/workflows/release-cycle.yml
```diff
@@ -19,10 +19,17 @@ jobs:
...
+
+      - name: Install Jinja
+        run: python -m pip install jinja2
 
       - name: Generate release cycle output
         run: python -I -bb -X dev -X warn_default_encoding -W error _tools/generate_release_cycle.py
 
+      - name: Ignore release-cycle.svg, which encodes the current date
+        run: git restore include/release-cycle.svg
+
```
CAM-Gerlach, line 32: Maybe just add the ability to pass a fixed date to the script to be able to get reproducible output and still be able to check this?

_static/devguide_overrides.css
```diff
@@ -7,66 +7,70 @@
...
-    --mermaid-text-color: black;
+.release-cycle-chart .release-cycle-blob-label {
+    /* white looks good on both light & dark */
+    fill: white;
+    filter:
+        drop-shadow(1px 1px 0.5px rgba(0, 0, 0, 0.5))
+        drop-shadow(-1px 1px 0.5px rgba(0, 0, 0, 0.5))
+        drop-shadow(1px -1px 0.5px rgba(0, 0, 0, 0.5))
+        drop-shadow(-1px -1px 0.5px rgba(0, 0, 0, 0.5))
+        ;
```
CAM-Gerlach, line 55: Maybe drop the drop shadow and add a contrasting stroke instead? That might resolve the text readability issue I'm having

requirements.txt
```diff
@@ -2,6 +2,6 @@ Sphinx==6.0.0
 furo>=2022.6.4
 sphinx-lint==0.6.7
 sphinx_copybutton>=0.3.3
-sphinxcontrib-mermaid
 sphinxext-opengraph>=0.7.1
 sphinxext-rediraffe
+jinja2
```
CAM-Gerlach, line 7: Can we try to keep this sorted lexicographically?

_tools/generate_release_cycle.py
```diff
@@ -70,38 +56,84 @@ def write_csv(self) -> None:
...
-                mermaid_status=MERMAID_STATUS_MAPPING[details["status"]],
-            )
-            out.append(v)
+        def format_year(year):
+            """Format year number for display"""
+            return f"'{year % 100:02}"
 
         with open(
-            "include/release-cycle.mmd", "w", encoding="UTF-8", newline="\n"
+            "include/release-cycle.svg", "w", encoding="UTF-8",
```
CAM-Gerlach, line 117:
```suggestion
            "include/release-cycle.svg", "w", encoding="UTF-8", newline="\n"
```

Why not keep this? Without it, as discussed on the previous PR, this checked in file will be written with inconsistent line endings on different platforms, which can lead to serious annoyances with Git.

CAM-Gerlach (commented):
@hugovk 's rationale seems compelling enough to me :)

Remaining issues not addressed by individual comments:

* [ ] The SVG template should be given an extension other than `SVG` (or similar), or configured with the appropriate `.gitattributes`, as this currently makes it impossible to easily view it or the changes to it on GitHub due to it attempting and failing to render as an SVG image.
* [x] The "current date" line is shown behind all of release bars, making it much harder to see than the original. Could it be moved in front?

hugovk:
> * [x] The SVG template should be given an extension other than `SVG` (or similar), or configured with the appropriate `.gitattributes`, as this currently makes it impossible to easily view it or the changes to it on GitHub due to it attempting and failing to render as an SVG image.

Good idea, I renamed it to `.svg.jinja` per https://jinja.palletsprojects.com/en/3.0.x/templates/#template-file-extension.

(Funnily, https://stackoverflow.com/q/29590931/724176 suggests both `.jinja` and `.jinja2`: PyCharm highlights syntax for `.jinja2` but not `.jinja` and VS Code highlights for `.jinja` but not `.jinja2`. I use PyCharm so added `.jinja` [here](https://user-images.githubusercontent.com/1324225/213912055-f9f66a03-d928-42b2-9e16-bf259b4e1916.png))


> * [x] The "current date" line is shown behind all of release bars, making it much harder to see than the original. Could it be moved in front?

Done!

image

https://cpython-devguide--1034.org.readthedocs.build/versions/

.github/workflows/release-cycle.yml
hugovk: I've added an override to be able to pass a fixed date.

But of course this still fails because we're generating for a fixed date, but the one in Git was last generated on some other, arbitrary date...

https://github.com/python/devguide/actions/runs/3979549956/jobs/6822048311

_static/devguide_overrides.css
hugovk: Updated, how's this?

# Before

image

# After

image

hugovk:
Added the final year label (e.g. `'28`) to the axis, but not an extra vertical line after it:

<details>
<summary>Details</summary>


# Before

image


# After

image

</details>

_static/devguide_overrides.css
CAM-Gerlach: LGTM!

.github/workflows/release-cycle.yml
CAM-Gerlach: Oh, right—duh, silly me! Maybe store the date as a string in a comment in the first or second line of the SVG and read it out of there?

CAM-Gerlach:
To avoid these CI warnings on Windows:

```
warning: in the working copy of 'include/branches.csv', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'include/end-of-life.csv', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'include/release-cycle.svg', LF will be replaced by CRLF the next time Git touches it
```

You might want to set might want to set `git config --global core.autocrlf input` as the first step in the workflow, before checkout.

> It does seem a problem that the "now" line is static and only rarely updated. But can we re-generate the SVG during each deploy? Possibly de-version from Git, similar to the RSS in the PEPs repo. This repo is deployed several times a week.

Meant to respond to this before, but yeah, I think that's the best approach, unless there's something I'm missing. AFAIK there's no actual need to actually check it in to the source other than it is built asynchronously from the rest of the docs and relied upon by them. So, now that we're only relying on Jinja (which is a Sphinx dep anyway, AFAIK) and it _should_ be a fairly cheap operation, we could either just add the `versions` make target as a prerequisite for the main build targets, or add running the script to the appropriate Sphinx pre-build hook as a Sphinx extension (which would work for all types of builds, not just through `make`).

Other than these two issues and the fixed date comment above, LGTM!

hugovk:
> You might want to set might want to set `git config --global core.autocrlf input` as the first step in the workflow, before checkout.

Done.

> ... we could either just add the `versions` make target as a prerequisite for the main build targets...

Also done.

.github/workflows/release-cycle.yml
hugovk: Well, and now the SVG is generated for every deploy and removed from Git (https://github.com/python/devguide/pull/1034/commits/59c2bbcb716ab22389951f8bbf875bfa89c97f82), I think we're probably fine with just running the script and making sure it doesn't fail? The other workflow makes sure the file can be generated and included in the docs with no error.

If so, do we need the `--today` override?

.github/workflows/release-cycle.yml
```diff
@@ -14,14 +14,22 @@ jobs:
...
+      - name: Install Jinja
+        run: python -m pip install jinja2
 
       - name: Generate release cycle output
-        run: python -I -bb -X dev -X warn_default_encoding -W error _tools/generate_release_cycle.py
+        run: python -I -bb -X dev -X warn_default_encoding -W error _tools/generate_release_cycle.py --today 2023-01-21
 
       - name: Check for differences
         run: |
```
CAM-Gerlach, line 35: Since we're now running this script on every CI and local (`make`) build anyway to generate these files, wouldn't it make sense to removed the checked-in generated CSVs and add them to gitignore too?

And since that already checks if the script ran successfully, and the files exist, are up to date, and are valid (at least, to the extent they can be injected into and loaded by Sphinx) the generate files are always up to date (in the production site, at least), this workflow is entirely redundant (since `git diff` won't show anything on `gitignored` files) and can be removed?

hugovk: Yeah, I guess we could do that too.

@ezio-melotti Any preference on whether the CSVs are kept in the repo? 

If we later have some external need for them, we could export them to something like `_static`

(We can also deal with this in a follow-up)

---

@CAM-Gerlach About the SVG, do you think we still need the `--today` override?

CAM-Gerlach: > (We can also deal with this in a follow-up)

Not sure if you're referring to this whole thing or just specifically the part about exposing the CSVs externally, but IMO the latter makes more sense to me, since the former will simplify things substantially and avoid the caveats/regressions this is introducing.

> @CAM-Gerlach About the SVG, do you think we still need the --today override?

With this workflow being removed, it won't be actively required/used in CI currently, but either we or external parties could have a potential usecase for for reproducible/user-controlled output, so I don't think it hurts to leave it in now that you've added it.

hugovk: Updated to also remove the CSVs from Git and generate them as part of the deploy.

CAM-Gerlach: Gotcha. Can we also get rid of this workflow, too? Since as mentioned, it doesn't really do anything now that the script is being run as part of the normal CIs, and the files are `gitignored` so the `git diff` check will never see them anyway?

ezio-melotti: > Updated to also remove the CSVs from Git and generate them as part of the deploy.

So now we use the json as primary source to generate both the SVG and the CSVs, right?

CAM-Gerlach: image

.github/workflows/release-cycle.yml
CAM-Gerlach: Just to follow up, we concluded below that since the script is being run (ensuring it doesn't fail) now during the regular CI checks anyway, we can just remove the workflow as its redundant, which you went ahead and did.

And we also discussed just leaving it in other other potential use cases, so I think this can be marked resolved now, thanks!

CAM-Gerlach (approved):
LGTM now, thanks @encukou and @hugovk for all your work here (and putting up with my curmudgeon-ing)

pradyunsg (approved):
This is quite neat!

hugovk:
Thanks for the reviews and to Petr for starting this off!

hugovk:
Deployed: https://devguide.python.org/versions/

If you see something like this: 

image

Then do a hard reload to get load the new CSS:

image

hugovk:
> Then do a hard reload (e.g. cmd-shift-r) to get load the new CSS:

Easier said than done on mobile... The file that needs refreshing is https://devguide.python.org/_static/devguide_overrides.css

https://docs.readthedocs.io/en/stable/hosting.html says Read the Docs busts the cache on the CDN when a new version is built, but looks like the max-age on that file is 86400 seconds (one day).

I don't see an easy way to force it to clear, short of renaming the file in Git, so guess we should just wait. (Or have users go into browser settings and clear the cache...)

CAM-Gerlach:
> Easier said than done on mobile... The file that needs refreshing is [devguide.python.org/_static/devguide_overrides.css](https://devguide.python.org/_static/devguide_overrides.css)

When testing the PEPs banner PR, press and holding on the reload button would trigger a hard reload in the browsers I tried, or at least appeared to.

> I don't see an easy way to force it to clear, short of renaming the file in Git, so guess we should just wait. (Or have users go into browser settings and clear the cache...)

What static site generators typically do is embed a hash of the source asset as a dummy query string in the asset url that changes when the file is updated, avoiding caching. Lektor, for example, as an [asseturl Jinja filter](https://www.getlektor.com/docs/templates/urls/) that does this. Not sure if Sphinx has something similar; this would have to be done in the theme, of course.

@pradyunsg , ideas?

hugovk:
Short term: good news, the cache has now expired, I see the chart properly on my phone.

---

Longer term:

It's good Furo includes a digest hash for its resources:

```html
    <link rel="stylesheet" type="text/css" href="../_static/pygments.css" />
    <link rel="stylesheet" type="text/css" href="../_static/styles/furo.css?digest=91d0f0d1c444bdcb17a68e833c7a53903343c195" />
    <link rel="stylesheet" type="text/css" href="../_static/copybutton.css" />
    <link rel="stylesheet" type="text/css" href="../_static/styles/furo-extensions.css?digest=30d1aed668e5c3a91c3e3bf6a60b675221979f0e" />
    <link rel="stylesheet" type="text/css" href="../_static/devguide_overrides.css" /> 
```

But Sphinx doesn't for our CSS file:

```py
html_css_files = [
    'devguide_overrides.css',
]
```

Or the other CSS files.

Maybe this could be improved in Sphinx? Maybe there's a bug in RTD not busting the cache properly?

pradyunsg:
https://github.com/sphinx-doc/sphinx/issues/11133

PR #1035, opened 2023-01-23
Mention MicroPython and CircuitPython in other implementations.

gpshead:
They're both popular and important, so they belong in the list of other Python implementations.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -168,6 +168,11 @@ developers to work on them.  Some major examples that may be of interest are:
 * Stackless_: A Python interpreter focused on providing lightweight
   microthreads while remaining largely compatible with CPython specific
   extension modules
+* MicroPython_: A tiny Python interpreter with small subset of the Python
+  standard library that is optimised to run on microcontrollers and in
+  constrained environments.
+* CircuitPython_: A fork of MicroPython designed to simplify experimenting
+  and learning to code on low-cost microcontroller boards.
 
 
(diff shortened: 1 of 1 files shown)


erlend-aasland (approved):
Makes sense; lgtm

CAM-Gerlach:
Maybe [RustPython](https://github.com/RustPython/RustPython) is also worth a mention? It has 13.4k stars (for whatever that's worth), 5-20x that of the Jython and IronPython repos, though I haven't heard of it actually being used seriously in production yet. I also hear a lot about [GraalPython](https://github.com/oracle/graalpython), as it's considered a modern successor to Jython, is under continual development in collaboration with PyPy, has a decent number of stars similar to Jython and IronPython (1k) and sees at least some limited production use, though it is still somewhat rough around the edges.

gpshead:
possibly.  I try not to bless a project until it stands on its own merit, but this doc isn't high profile so linking to something that seems to both have users and be actively maintained isn't a bad thing.  we can change them later if they seem stale.

IronPython appears "maintained" but they're proud of releasing version 3.4 last month so I don't know what to make of that. ;)

FFY00:
I just found out that RustPython is actually used in https://github.com/charliermarsh/ruff/ for its parser, so IMO I'd say it is standing on its own merit :rofl:

gpshead:
just make another PR :)

PR #1038, opened 2023-01-25
Overhaul Building the Documentation section for clarity

### LINKED ISSUE #1037: Building the documentation confusion leads to error: no theme named 'python-docs-theme' found (missing theme.conf?)
# Documentation

After reading the devguide https://devguide.python.org/documentation/start-documenting/#building-the-documentation to build the documentation, I faced the very common error: no theme named 'python-docs-theme' found (missing theme.conf?).
```
(cpython) peter@xxx:~/repo/cpython/Doc$ sphinx-build -b html . build/html
Running Sphinx v4.3.2

Theme error:
no theme named 'python_docs_theme' found (missing theme.conf?)
(cpython) peter@xxx:~/repo/cpython/Doc$ which sphinx-build
/usr/bin/sphinx-build
```
The root cause is that **sphinx is not installed in the virtual env where python-docs-theme is installed**, as mentioned in python-docs-theme issue [#45](https://github.com/python/python-docs-theme/issues/45#issuecomment-1233881707).

Full Steps to Reproduce:
```
Use ubuntu 22.04 or 16.04
apt-get install python3-sphinx # as per https://www.sphinx-doc.org/en/master/usage/installation.html#linux
<create a virtual env>
pip install blurb
pip install python-docs-theme
cd Doc
make html
or 
sphinx-build -b html . build/html
```

Of course, it works if you follow exactly this:
```
cd Doc
make venv
make html
```

Where does the confusion come from:
"You are expected to have installed the latest stable version of [Sphinx](https://www.sphinx-doc.org/) and [blurb](https://pypi.org/project/blurb/) on your system or in a [virtualenv](https://virtualenv.pypa.io/) (which can be created using make venv), so that the Makefile can find the sphinx-build command. You can also specify the location of sphinx-build with the SPHINXBUILD make variable. "

1/  [Sphinx](https://www.sphinx-doc.org/) This links to https://www.sphinx-doc.org and it provides installation steps `apt-get install python3-sphinx` which installs sphinx in system path /usr/bin/sphinx-build. It won't work with python-docs-theme installed in a virtualenv.

[Suggestion] Change the sphinx  link to https://github.com/sphinx-doc/sphinx which provide install step `pip install -U sphinx` instead.

2/ "You can also specify the location of sphinx-build with the SPHINXBUILD make variable" => this also makes me think sphinx-build is okay to be anywhere, but it is not.

CAM-Gerlach:
As discussed in #1037 , the [Building the documentation section](https://devguide.python.org/documentation/start-documenting/#building-the-documentation) is potentially confusing and difficult to follow to a new contributor (i.e. the primary target audience), as it lacks a clear, concrete, step by step structure, muddles minor or advanced details and elaboration with key basic information, and has some out of date information (sphinx-build, dependency names/versions, third-party virtualenv, etc), which also isn't in keeping with the [Diataxis principles for guides](https://diataxis.fr/how-to-guides/) like this.

Furthermore, as @peterjpxie mentioned, it manually lists certain dependencies and states that they are required to be installed no less than three times (including twice in the intro paragraph), which can naturally confuse the contributor into thinking they need to install such themselves first, which is instead handled for the user automatically with `make venv`/running `make.bat`, and should be installed via the `requirements.txt` rather than manually otherwise. Furthermore, it doesn't at least link to how to create and activate the necessary environment on Windows and non-make usage, nor state how to install the required dependencies or link to the canonical listing of them (in the `requirements.txt`).

Therefore, this overhauls the section to address the above issues, with the following major changes (not exhaustive):

* Eliminate repeated and inconsistent listings of required dependencies, and instead refer to the canonical `requirements.txt` when appropriate
* Put step by step commands into codeblocks, so they are easier to find, follow and copy accurately
* Factor out implementation details and highly specific information that merely distract the reader from the core guide content, and refer to the readme for that instead
* Add an admonition at the top stating that all commands assume a CWD of `Doc`, and the user should `cd` to it if not, instead of an inconstantly either repeating `cd Docs`, assuming it's been run, or assuming the opposite.
* Instead of just an offhand link to the third-party `virtualenv` package, include a short section explicitly stating how to create a venv with the Makefile, or otherwise, linking to the [up-to-date section in the PyPUG](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) describing venv creation and activation, and more explicitly remind Windows users they need to manually activate it.
* Factor out the distinction between the Makefile and `make.bat` into the above section, making the actual Build with make/make.bat section work the same on *nix and Windows
* Revise the "sphinx-build" section to state how to install the requirements from the `requirements.txt` (instead of manually listing them yet again), and follow modern best practices with the invocation.

Fixes #1037

### DIFF
--- a/documentation/start-documenting.rst
+++ b/documentation/start-documenting.rst
@@ -103,7 +103,7 @@ before building the documentation.
 .. _using-make-make-bat:
 .. _doc-build-make:
 
-Build using Make / make.bat
+Build using make / make.bat
 ---------------------------
 
 A Unix ``Makefile`` is provided, :cpy-file:`Doc/Makefile`,
@@ -112,10 +112,10 @@ that attempts to emulate it as closely as practical.
 
 .. important::
 
-   The Windows ``make.bat`` batch file lacks a ``make venv`` target,
-   automatically installing any missing dependencies
-   into your current environment.
-   Make sure the environment you `created above <doc-create-venv-windows>`__
+   The Windows ``make.bat`` batch file lacks a ``make venv`` target.
+   Instead, it automatically installs any missing dependencies
+   into the currently activated environment (or the base Python, if none).
+   Make sure the environment you :ref:`created above <doc-create-venv-windows>`
    is `activated <venv-activate_>`__ before running ``make.bat``.
 
 To build the docs as HTML, run::
@@ -125,12 +125,12 @@ To build the docs as HTML, run::
 .. tip:: Substitute ``htmlview`` for ``html`` to open the docs in a web browser
          once the build completes.
 
-To check the docs for common errors with `sphinx-lint`_
+To check the docs for common errors with `Sphinx Lint`_
 (which is run on all :ref:`pull requests <pullrequest>`), use::
 
    make check
 
-To list other supported :program:`Make` targets for your platform, run::
+To list other supported :program:`make` targets, run::
 
    make help
 
@@ -146,7 +146,7 @@ Build using Sphinx directly
 Advanced users may want to invoke Sphinx directly,
 to pass specialized options or to handle specific use cases.
 
-First, ensure the environment you `created above <doc-create-venv-windows>`__
+Make sure the environment you :ref:`created above <doc-create-venv-windows>`
 is `activated <venv-activate_>`__.
 Then, install the documentation requirements, :cpy-file:`Doc/requirements.txt`.
 Using pip::
@@ -164,7 +164,7 @@ replace ``html`` above with the desired builder ``name``.
 .. _docutils: https://docutils.sourceforge.io/
 .. _Sphinx: https://www.sphinx-doc.org/
 .. _Sphinx builder: https://www.sphinx-doc.org/en/master/usage/builders/index.html
-.. _sphinx-lint: https://github.com/sphinx-contrib/sphinx-lint
+.. _Sphinx Lint: https://github.com/sphinx-contrib/sphinx-lint
 .. _venv-activate: https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#activating-a-virtual-environment
 .. _venv-create: https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment
 
--- a/documentation/start-documenting.rst
+++ b/documentation/start-documenting.rst
@@ -64,79 +64,109 @@ Building the documentation
 
 .. highlight:: bash
 
-The toolset used to build the docs is written in Python and is called Sphinx_.
-Sphinx is maintained separately and is not included in this tree.  Also needed
-are blurb_, a tool to create :file:`Misc/NEWS` on demand; and
-python-docs-theme_, the Sphinx theme for the Python documentation.
+To build the documentation, follow the steps in one of the sections below.
+You can view the documentation after building the HTML
+by opening the file :file:`Doc/build/html/index.html` in a web browser.
 
-To build the documentation, follow the instructions from one of the sections
-below.  You can view the documentation after building the HTML by pointing
-a browser at the file :file:`Doc/build/html/index.html`.
+.. note::
 
-You are expected to have installed the latest stable version of
-Sphinx_ and blurb_ on your system or in a virtualenv_ (which can be
-created using ``make venv``), so that the Makefile can find the
-``sphinx-build`` command.  You can also specify the location of
-``sphinx-build`` with the ``SPHINXBUILD`` :command:`make` variable.
+   The following instructions all assume your current working dir is
+   the ``Doc`` subdirectory in your :ref:`CPython repository clone <checkout>`.
+   Make sure to switch to it with ``cd Doc`` if necessary.
+
+
+.. _doc-create-venv:
+
+Create a virtual environment
+----------------------------
+
+.. _doc-create-venv-unix:
+
+**On Unix platforms** that support :program:`make`
+(including Linux, macOS and BSD),
+you can create a new :mod:`venv` with the required dependencies using::
+
+   make venv
+
+Building the docs with :program:`make` will automatically use this environment
+without you having to activate it.
+
+.. _doc-create-venv-windows:
+
+**On Windows**, or if not using :program:`make`,
+`create a new virtual environment <venv-create_>`__ manually.
+Always be sure to `activate this environment <venv-activate_>`__
+before building the documentation.
 
 
 .. _building-using-make:
+.. _using-make-make-bat:
+.. _doc-build-make:
 
-Using make / make.bat
----------------------
+Build using Make / make.bat
+---------------------------
 
-**On Unix**, run the following from the root of your :ref:`repository clone
-<checkout>` to build the output as HTML::
+A Unix ``Makefile`` is provided, :cpy-file:`Doc/Makefile`,
+along with a :cpy-file:`Doc/make.bat` batch file for Windows
+that attempts to emulate it as closely as practical.
 
-   cd Doc
-   make venv
+.. important::
+
+   The Windows ``make.bat`` batch file lacks a ``make venv`` target,
+   automatically installing any missing dependencies
+   into your current environment.
+   Make sure the environment you `created above <doc-create-venv-windows>`__
+   is `activated <venv-activate_>`__ before running ``make.bat``.
+
+To build the docs as HTML, run::
+
    make html
 
-or alternatively ``make -C Doc/ venv html``.  ``htmlview`` can be used
-instead of ``html`` to conveniently open the docs in a browser once the
-build completes.
+.. tip:: Substitute ``htmlview`` for ``html`` to open the docs in a web browser
+         once the build completes.
 
-You can also use ``make help`` to see a list of targets supported by
-:command:`make`.  Note that ``make check`` is automatically run when
-you submit a :ref:`pull request <pullrequest>`, so you should make
-sure that it runs without errors.
+To check the docs for common errors with `sphinx-lint`_
+(which is run on all :ref:`pull requests <pullrequest>`), use::
 
-**On Windows**, the :cpy-file:`Doc/make.bat` batchfile tries to emulate
-:command:`make` as closely as possible, but the venv target is not implemented,
-so you will probably want to make sure you are working in a virtual environment
-before proceeding, otherwise all dependencies will be automatically installed
-on your system.
+   make check
 
-When ready, run the following from the root of your :ref:`repository clone
-<checkout>` to build the output as HTML::
+To list other supported :program:`Make` targets for your platform, run::
 
-   cd Doc
-   make html
+   make help
 
-You can also use ``make help`` to see a list of targets supported by
-:cpy-file:`Doc/make.bat`.
+See :cpy-file:`Doc/README.rst` for more information.
 
-See also :cpy-file:`Doc/README.rst` for more information.
 
-Using sphinx-build
-------------------
+.. _using-sphinx-build:
+.. _doc-build-sphinx:
 
-Sometimes we directly want to execute the sphinx-build tool instead of through
-``make`` (although the latter is still the preferred way). In this case, you can
-use the following command line from the ``Doc`` directory (make sure to install
-Sphinx_, blurb_ and python-docs-theme_ packages from PyPI)::
+Build using Sphinx directly
+---------------------------
 
-   sphinx-build -b<builder> . build/<builder>
+Advanced users may want to invoke Sphinx directly,
+to pass specialized options or to handle specific use cases.
 
-where ``<builder>`` is one of html, text, latex, or htmlhelp (for explanations
-see the make targets above).
+First, ensure the environment you `created above <doc-create-venv-windows>`__
+is `activated <venv-activate_>`__.
+Then, install the documentation requirements, :cpy-file:`Doc/requirements.txt`.
+Using pip::
+
+   python -m pip install --upgrade -r requirements.txt
+
+Finally, directly invoke Sphinx with::
+
+   python -m sphinx -b html . build/html
+
+To use a different `Sphinx builder`_,
+replace ``html`` above with the desired builder ``name``.
 
 
 .. _docutils: https://docutils.sourceforge.io/
-.. _python-docs-theme: https://pypi.org/project/python-docs-theme/
 .. _Sphinx: https://www.sphinx-doc.org/
-.. _virtualenv: https://virtualenv.pypa.io/
-.. _blurb: https://pypi.org/project/blurb/
+.. _Sphinx builder: https://www.sphinx-doc.org/en/master/usage/builders/index.html
+.. _sphinx-lint: https://github.com/sphinx-contrib/sphinx-lint
+.. _venv-activate: https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#activating-a-virtual-environment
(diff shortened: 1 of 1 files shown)


documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+`create a new virtual environment <venv-create_>`__ manually.
+Always be sure to `activate this environment <venv-activate_>`__
+before building the documentation.
+
+
+.. _building-using-make:
+.. _using-make-make-bat:
+.. _doc-build-make:
+
+Build using Make / make.bat
```
hugovk, line 106:
```suggestion
Build using make / make.bat
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
-:cpy-file:`Doc/make.bat`.
+.. tip:: Substitute ``htmlview`` for ``html`` to open the docs in a web browser
+         once the build completes.
+
+To check the docs for common errors with `sphinx-lint`_
+(which is run on all :ref:`pull requests <pullrequest>`), use::
+
+   make check
+
+To list other supported :program:`Make` targets for your platform, run::
```
hugovk, line 133: Lowercase make, and we can remove "for your platform":
```suggestion
To list other supported :program:`make` targets, run::
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
-before proceeding, otherwise all dependencies will be automatically installed
-on your system.
+.. important::
 
-When ready, run the following from the root of your :ref:`repository clone
-<checkout>` to build the output as HTML::
+   The Windows ``make.bat`` batch file lacks a ``make venv`` target,
+   automatically installing any missing dependencies
+   into your current environment.
+   Make sure the environment you `created above <doc-create-venv-windows>`__
```
hugovk, line 118:
```suggestion
   Make sure the environment you `created above <doc-create-venv-windows_>`__
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+.. _doc-build-sphinx:
+
+Build using Sphinx directly
+---------------------------
+
+Advanced users may want to invoke Sphinx directly,
+to pass specialized options or to handle specific use cases.
 
-See also :cpy-file:`Doc/README.rst` for more information.
+First, ensure the environment you `created above <doc-create-venv-windows>`__
```
hugovk, line 149:
```suggestion
First, ensure the environment you `created above <doc-create-venv-windows_>`__
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+   python -m pip install --upgrade -r requirements.txt
 
-Sometimes we directly want to execute the sphinx-build tool instead of through
-``make`` (although the latter is still the preferred way). In this case, you can
-use the following command line from the ``Doc`` directory (make sure to install
-Sphinx_, blurb_ and python-docs-theme_ packages from PyPI)::
+Finally, directly invoke Sphinx with::
 
-   sphinx-build -b<builder> . build/<builder>
+   python -m sphinx -b html . build/html
```
hugovk, line 158: Also `-j auto` to 'build in parallel with N processes where possible (special value "auto" will set N to cpu-count)'?

```suggestion
   python -m sphinx -b html . build/html -j auto
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
 
-   cd Doc
    make html
 
-You can also use ``make help`` to see a list of targets supported by
-:cpy-file:`Doc/make.bat`.
+.. tip:: Substitute ``htmlview`` for ``html`` to open the docs in a web browser
+         once the build completes.
+
+To check the docs for common errors with `sphinx-lint`_
```
hugovk, line 128: The project repo calls it Sphinx Lint:

```suggestion
To check the docs for common errors with `Sphinx Lint`_
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+replace ``html`` above with the desired builder ``name``.
 
 
 .. _docutils: https://docutils.sourceforge.io/
-.. _python-docs-theme: https://pypi.org/project/python-docs-theme/
 .. _Sphinx: https://www.sphinx-doc.org/
-.. _virtualenv: https://virtualenv.pypa.io/
-.. _blurb: https://pypi.org/project/blurb/
+.. _Sphinx builder: https://www.sphinx-doc.org/en/master/usage/builders/index.html
+.. _sphinx-lint: https://github.com/sphinx-contrib/sphinx-lint
```
hugovk, line 167:
```suggestion
.. _Sphinx Lint: https://github.com/sphinx-contrib/sphinx-lint
```

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
-so you will probably want to make sure you are working in a virtual environment
-before proceeding, otherwise all dependencies will be automatically installed
-on your system.
+.. important::
 
-When ready, run the following from the root of your :ref:`repository clone
-<checkout>` to build the output as HTML::
+   The Windows ``make.bat`` batch file lacks a ``make venv`` target,
+   automatically installing any missing dependencies
+   into your current environment.
```
hugovk, line 117:
```suggestion
   The Windows ``make.bat`` batch file lacks a ``make venv`` target
   to automatically install missing dependencies
   into your current environment.
```

hugovk (commented):
Looks good! Tested out in a new clone as well on Mac and worked as expected.

documentation/start-documenting.rst
CAM-Gerlach: I wasn't initially sure about this, but the [official website for GNU Make](https://www.gnu.org/software/make/) (which is used on the two primary platforms in question) always refers to the name of the software as capitalized Make (as opposed to the ``make`` command), so I followed that convention here.

That said, arguably it is referring to the (unformatted) lowercase ``make`` command rather than the program name, which would be consistent with the general usage of `make.bat`, so we may as well go ahead with this change. Not that it matters *that* much...something something the hardest problems in programming :)

documentation/start-documenting.rst
CAM-Gerlach: Actually, this intends to say the opposite, that instead of having a `venv` target for it, the `make.bat` always automatically installs the required dependencies into the current environment whenever it is run. However, I can definitely see how that is unclear, so here's some edits to clarify that:

```suggestion
   The Windows ``make.bat`` batch file lacks a ``make venv`` target.
   Instead, it automatically installs any missing dependencies
   into the currently activated environment (or the base Python, if none).
```

documentation/start-documenting.rst
CAM-Gerlach: Oops, this should actually be a `:ref:` role instead—good catch.

```suggestion
   Make sure the environment you :ref:`created above <doc-create-venv-windows>`
```

documentation/start-documenting.rst
CAM-Gerlach: Yeah, I misinterpreted the docs on the intended semantics for `:program:` to mean it should refer to the program's proper name, as opposed to just omitting any non-required platform-specific file extension from the invocation.

documentation/start-documenting.rst
CAM-Gerlach: As above, I actually meant this to be a `:ref:` role instead:

```suggestion
   Make sure the environment you :ref:`created above <doc-create-venv-windows>`
```

documentation/start-documenting.rst
CAM-Gerlach: Hmm, perhaps, but since this is primarily intended to provide a minimal working invocation specifically for advanced users who want to pass their own custom options anyway, I'm not sure whether it makes sense to add here, considering it is entirely optional, can easily be added if desired, doesn't appear to do anything on Windows (the platform most likely to use it), and may not make sense for many use cases for which the direct invocation would be used, e.g. rebuilding only a single file or running on CI.

documentation/start-documenting.rst
hugovk: When I checked I ended up on https://www.gnu.org/software/make/manual/make.html, which *always?) uses lowercase except in Title Case, but it is also referring to the monospace font `make` command.

Well, either is fine, take your pick :)

documentation/start-documenting.rst
hugovk: Shame it doesn't work on Windows.

So it will either do nothing or make things much quicker: 62s -> 41s on my machine, which I think is a good win without much complexity. Especially if this bit is intended for advanced users.

And some Linux/Mac users who could use `make` will prefer to call things directly.

Anyway, I don't mind too much, take your pick here too!

hugovk (approved):

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+.. _doc-create-venv-unix:
 
-**On Unix**, run the following from the root of your :ref:`repository clone
-<checkout>` to build the output as HTML::
+**On Unix platforms** that support :program:`make`
+(including Linux, macOS and BSD),
+you can create a new :mod:`venv` with the required dependencies using::
 
-   cd Doc
    make venv
```
peterjpxie, line 89: I was having a hard time with `venv` on my ubuntu 16.04 dev machine where I installed and used python3.7 instead of the system default python 3.5.

It seems a quite common issue that people are facing.

https://askubuntu.com/questions/1268833/error-command-path-to-env-bin-python3-7-im-ensurepip-upgrade
```
$ make venv
Error: Command '['/root/repo/cpython/Doc/venv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
/root/repo/cpython/Doc/venv/bin/python3: No module named pip
/root/repo/cpython/Doc/venv/bin/python3: No module named pip
The venv has been created in the ./venv directory

$ python3.7 -m venv venv1
Error: Command '['/root/repo/cpython/Doc/venv1/bin/python3.7', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
```

So I was wondering we could further simplify the instruction by removing make venv and let the users create a virtual env with tools they are comfortable with, e.g. virtualenv, pipenv etc. 

And we could have streamlined steps for all platforms with something like this:
```
<Create a virtual environment>
cd Doc
python -m pip install --upgrade -r requirements.txt
# build using make
make html
# build using sphinx directly
python -m sphinx -b html . build/html
```

It is just my personal experience for your reference. Thanks

CAM-Gerlach: Hmm, thanks for the feedback. Personally, I mostly use Conda environments for development, including to build the CPython docs, devguide, PEPs, etc, and thus don't really use the Makefile at all. myself. And it would certainly be nice to simplify this down to one cross-platform workflow...but unfortunately, there are a lot of complexities here that make that not really possible.

Switching to telling the user to create an arbitrary environment using their preferred environment manager requires either:

* Major modifications to the implementation and usage of the Makefile that would break existing workflows, since it assumes that a venv-created virtual environment is present at a specific name and location with the required dependencies, or
* Discarding the whole Makefile-based approach completely, and just recommending the `python -m sphinx` approach.

*Perhaps* one of these two options could be considered, but would first require a whole separate discussion from just improving the devguide documentation on the existing approaches.

Furthermore, I'm really not sure either of those approaches would be a good idea, as they would substantially complicate environment setup for the bulk of Make users, especially for doc contributors who may or may not be experienced in Python package management, as it requires them to know/learn how to create an environment and activate it on their particular platform, and furthermore ensure it is always activated before building the docs, as opposed to running a single command once.

In fact, the `venv` target could be modified to work like the one we have in the PEPs repo, which is automatically run as a dependency of the default build targets, automatically creating the environment if it doesn't exist and existing cleanly otherwise, such that users don't need to do any setup step at all—just run `make html` and It Just Works™. But, that's a subject for another issue...

This likely benefits many more users than those affected by Debian's persistent and self-inflicted Python packaging problems, or those wanting to build the docs with their own custom Python. For those users who've installed their own custom Python and want to run it with that, building directly with `python -m sphinx` inside an environment created with that Python is really the way to go.

> And we could have streamlined steps for all platforms with something like this:

Unfortunately, beyond the above, this is missing a key step—_activating_ the environment, which (with user-created environments) must be performed every time rather than just once, and like creating it, will vary depending on the type of environment. It's also missing a lot of key details from the original, like clarifying that `make` is available on both Windows and *nix and the difference between the two, what other commands can be used, how to get help, when `make` should be used vs. direct invocation, making unambiguous which steps are run once vs. every session, and which are specific to make or not, etc. Also, putting all the commands in one block without explaining the purpose of each step doesn't help users understand what they are doing and why, particularly important when they run into trouble or want to do something beyond the specific task stated. Additionally, it's missing the links and cross-references from the original that the user can consult for more information and guidance.

documentation/start-documenting.rst
```diff
@@ -64,79 +64,109 @@ Building the documentation
...
+   make check
+
+To list other supported :program:`make` targets, run::
+
+   make help
+
+See :cpy-file:`Doc/README.rst` for more information.
+
+
+.. _using-sphinx-build:
```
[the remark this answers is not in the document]
CAM-Gerlach: This ref target label was added explicitly for backward compatibility, so that existing external URLs (and any implicit internal hyperlinks) that point to this section won't suddenly break. Renaming it would, of course, defeat that whole purpose :)

The up to date ref target label, `doc-build-sphinx`, is immediately below. FWIW, I've adopted the convention of putting the current canonical (i.e. newest) ref target label last, which makes sense both viz the syntax (as the other labels are de facto "aliases" to it), and by the general convention of adding newer things below older ones.

documentation/start-documenting.rst
CAM-Gerlach: I'm kinda torn here...it's nice that it speeds things up for some people, but adding extra optional, non-self-documenting arguments to the baseline invocation causes me a bit of concern over complicating things for users, and them retaining it in cases where it is not helpful (e.g. rebuilding only one file). Though, on the other hand, under the same assumption that users care about the arguments they are passing, they could just as easily be aware of what `-j auto` does, or be looking up the arguments anyway.

> Shame it doesn't work on Windows.

My understanding from what Adam told me is its due to lack of `fork` support, which makes spinning up new processes very expensive and rarely worth the trouble for the types of workloads Sphinx is typically run on.

> And some Linux/Mac users who could use `make` will prefer to call things directly.

FWIW, I have a working `make` installed through Rtools on Windows (and also exclusively use Git Bash), and use it myself to build the Spyder Docs I maintain (since we have a fairly complex workflow for multi-version support), but don't use it for the CPython-related projects that use it to build docs, since it assumes an in-tree venv (whereas I'm using conda environments), I need precise control over both the Python invocation and the arguments passed to Sphinx, and I prefer a consistent invocation and behavior, with no extra magic, across the many docs projects I work on.

documentation/start-documenting.rst
peterjpxie: @CAM-Gerlach Thanks! Let's leave it as is.

pradyunsg (approved):

documentation/start-documenting.rst
ferdnyc: Does `python -m sphinx` not accept `--parallel`, as a synonym for `-j`? If not (and AFAICT it truly, in fact, does not), _that's_ probably something that should be fixed (in sphinx), and would at least help to address the "non-self-documenting" aspect.

ferdnyc: (I've submitted https://github.com/sphinx-doc/sphinx/pull/11169 to do just that.)

PR #1041, opened 2023-02-01
Don't force table widths in versions.rst

ferdnyc:
Setting `:widths:` leads to weird wrapping, as some columns are too narrow. Just letting Sphinx figure it out, OTOH, looks fine. Let the software be smart.

Signed-off-by: FeRD (Frank Dana) <ferdnyc@gmail.com>

### DIFF
--- a/versions.rst
+++ b/versions.rst
@@ -23,7 +23,7 @@ Dates shown in *italic* are scheduled and can be adjusted.
 
 .. csv-table::
    :header-rows: 1
-   :widths: 5, 5, 15, 15, 12, 40
+   :width: 100%
    :file: include/branches.csv
 
 .. Remember to update main branch in the paragraph above too
@@ -34,7 +34,7 @@ Unsupported Versions
 
 .. csv-table::
    :header-rows: 1
-   :widths: 5, 5, 15, 15, 12, 40
+   :width: 100%
    :file: include/end-of-life.csv
 
 


hugovk:
This does look better to me.

In Chrome on macOS:

<details>
<summary>Before</summary>

image

</details>


<details>
<summary>After</summary>

image

</details>

CAM-Gerlach:
Using Firefox on Windows (desktop), at normal widths they look about the same, except with this PR, the widths of the two tables don't match, which looks a little odd and makes them harder to compare.

|Before|After|
|---|---|
|image|image|

At substantially narrower widths where they start to get squished, meanwhile, they look essentially exactly the same, as far as I can tell.

This seems to be an issue particular to Chrome, since I've seen similar differing table column behavior on other (non-Sphinx) sites I maintain, where Firefox handles the column widths much more sensibly whereas Chrome splits things into multiple columns.

While it's perhaps a bit icky to introduce a regression for a community-developed FOSS browser to benefit the monopoly one, the first issue is perhaps better handled in CSS with a custom CSS class that we then have make the table full width.

ezio-melotti (commented):
I added these widths for two reasons:
1. Make the columns wide enough to fit their content, and prevent it to be split on two lines
2. Make the two tables have the same widths and therefore be aligned

Admittedly I think I only tested it on Firefox, which might explain different behaviors on Chrome it other browsers.  IIRC there was a discussion about them in the original PR, and eventually I decided to keep them.

ferdnyc:
> I added these widths for two reasons:
> 
> 1. Make the columns wide enough to fit their content, and prevent it to be split on two lines

Unfortunately, that's the exact thing it's failing to do, but removing the `:widths:` easily achieves.

> 2. Make the two tables have the same widths and therefore be aligned

If you just want the _tables_ to have the same width, setting the `:width: 100%` option on both might be better than setting column widths explicitly. Granted, the _columns_ won't align between the two tables, but... I'm not sold on that being important, TBH. It looks fine to me without it, at least to me:

image

(That's with `:width: 100%` set on both tables, but `:widths:` still removed.)

> Admittedly I think I only tested it on Firefox, which might explain different behaviors on Chrome it other browsers. 

Or, it could be that it **did** originally fit, but some values have gotten wider/narrower since then. That "Georg Brandl, Ned Deily (3.3.7+)" string in the unsupported table is an outlier, and it's going to force that one out of whack no matter what, compared to the supported-versions table. I'm not personally convinced that it's really worth trying to struggle against that, but YMMV.

For any table where the _data_ isn't static, fixed table widths will **inevitably** fall out of alignment with reality unless they're constantly being adjusted to account for every single change in the data. IMHO, that kind of thing gets filed under, "that's the code's problem to solve, not mine!"

ferdnyc:
> > Admittedly I think I only tested it on Firefox, which might explain different behaviors on Chrome it other browsers.
> 
> Or, it could be that it **did** originally fit, but some values have gotten wider/narrower since then.

No, actually, FWIW the current live page does indeed still lay out un-wrapped in Firefox. But _only_ in Firefox, I guess, which... `¯\_(ツ)_/¯`

ezio-melotti:
>  IIRC there was a discussion about them in the original PR, and eventually I decided to keep them.

In the [original discussion](https://github.com/python/devguide/pull/884#issuecomment-1173186169) `:widths: auto` was suggested, and at the time it didn't seem to work too well, however I tried again now and it seems to work (even the preview in the original discussion looks ok now).

I'm still using Firefox, so I don't know if the issue got fixed by Firefox, Sphinx, or if it's simply because I'm using a different monitor (which affects the width of the viewport).

> If you just want the tables to have the same width, setting the `:width: 100%` option on both might be better than setting column widths explicitly.

This sounds like a good compromise.  I tested locally with `:width: 100%` and no `:widths:` and it looks ok.  With `:widths: auto` it also looks ok, but I think that's already the default and can be omitted.

>  Granted, the columns won't align between the two tables, but... I'm not sold on that being important, TBH. It looks fine without it, at least to me

Agreed -- aligned columns are nice to have, but hardly essential.

IOW, I'm fine with replacing the current `:widths:` with `:width: 100%`.

hugovk:
> IOW, I'm fine with replacing the current `:widths:` with `:width: 100%`.

@ferdnyc Please could you make this change?

ferdnyc:
@ezio-melotti @hugovk 

Done, sorry for the delay.

hugovk (approved):

ferdnyc:
_Really_, Firefox!??! Augh! :disappointed:

CAM-Gerlach:
To be fair, it was "_Really_, Chrome!??! *[sic]* Augh!" before, heh...

ezio-melotti (approved):

PR #1043, opened 2023-02-06
Link to Discourse first in "Help with Documentation" section

Mariatta:
The Docs Discourse forum is the preferred way.
I reordered the sentence so that the discourse forum is mentioned first.

### DIFF
--- a/documentation/help-documenting.rst
+++ b/documentation/help-documenting.rst
@@ -36,11 +36,10 @@ and :ref:`maintenance <maintbranch>` branches at https://docs.python.org/dev/.
 The in-development and recent maintenance branches are rebuilt once per day.
 
 If you would like to be more involved with documentation, consider subscribing
-to the `docs@python.org <https://mail.python.org/mailman3/lists/docs.python.org/>`_
-mailing list and the `Documentation category on the Python Discourse
-<https://discuss.python.org/c/documentation/26>`_,
-where user issues are raised and documentation toolchain, projects, and standards
-are discussed.
+to the `Documentation category on the Python Discourse
+<https://discuss.python.org/c/documentation/26>`_ and the `docs@python.org <https://mail.python.org/mailman3/lists/docs.python.org/>`_
+mailing list where user issues are raised and documentation toolchain,
+projects, and standards are discussed.  
 
 
 Helping with documentation issues


documentation/help-documenting.rst
```diff
@@ -36,11 +36,10 @@ and :ref:`maintenance <maintbranch>` branches at https://docs.python.org/dev/.
...
 If you would like to be more involved with documentation, consider subscribing
-to the `docs@python.org <https://mail.python.org/mailman3/lists/docs.python.org/>`_
-mailing list and the `Documentation category on the Python Discourse
-<https://discuss.python.org/c/documentation/26>`_,
-where user issues are raised and documentation toolchain, projects, and standards
-are discussed.
+to the `Documentation category on the Python Discourse
+<https://discuss.python.org/c/documentation/26>`_ and the `docs@python.org <https://mail.python.org/mailman3/lists/docs.python.org/>`_
+mailing list where user issues are raised and documentation toolchain,
+projects, and standards are discussed.
```
hugovk, line 42: Whitespace nit for the linter:

```suggestion
projects, and standards are discussed.
```

CAM-Gerlach, line 42: Fix trailing whitespace (causing the lint failure), and rationalize the line breaks to minimize diffs and keep the line lengths at least roughly consistent.

```suggestion
to the `Documentation category on the Python Discourse
<https://discuss.python.org/c/documentation/26>`_ and the
`docs@python.org <https://mail.python.org/mailman3/lists/docs.python.org/>`_ mailing list
where user issues are raised and documentation toolchain, projects, and standards
are discussed.
```

safwansamsudeen:
@Mariatta if you're busy right now, I'd love to take this up!

I have it [here](https://github.com/python/devguide/compare/main...safwansamsudeen:devguide:patch-2), and I'll submit a PR if this is closed - I'm not sure how I can contribute directly here.

I mean, it's a small change and you can easily do it yourself, but I'm new here and scrambling for issues to help with ;).

CAM-Gerlach:
Thanks for the offer @safwansamsudeen , but if @Mariatta 's busy, we can simply commit my suggestion here to achieve the same thing, without all the overhead of a whole other PR and without stripping author/attribution information.

However, your help on the other PR you commented would be much appreciated, and a new clean branch is indeed the approach that would be required there, if you're up for it. Thanks!

documentation/help-documenting.rst
CAM-Gerlach: Hey @Mariatta , do we just want to commit this and merge, or do you have other ideas here? Thanks!

CAM-Gerlach: Per Mariatta's confirmation, going ahead with this.

Mariatta:
Thanks for the ping and reminding me of this PR. I will take a look once I'm back on the computer.

ezio-melotti (approved):
PR LGTM with @CAM-Gerlach suggestion applied.

safwansamsudeen:
> Thanks for the offer @safwansamsudeen , but if @Mariatta 's busy, we can simply commit my suggestion here to achieve the same thing, without all the overhead of a whole other PR and without stripping author/attribution information.
> 
> However, your help on the other PR you commented would be much appreciated, and a new clean branch is indeed the approach that would be required there, if you're up for it. Thanks!

Sounds good!

documentation/help-documenting.rst
CAM-Gerlach: Because it seems the branch was mistakenly pushed to the upstream repo rather than the user's fork (as the user was an admin, the current protections against this didn't work—I've tested and confirmed an improved version on python/core-workflow#460 that would prevent accidents like this going forward), @Mariatta or one of her fellow admins will have to commit this suggestion.

Mariatta:
I recreated the PR using my own fork in #1118.

CAM-Gerlach:
Okay, thanks. Did it not allow you to commit the above suggestion? I thought it would as an admin (since it allowed you to push initially), but maybe not.

Anyway, would you ,omd cleaning up these two stale branches, as well as the other two merged ones? The [branches page](https://github.com/python/devguide/branches) lists all four, and you should be able to delete them with one click from there (assuming `Allow deletion` is enabled for the `*` branch protection rule, as it should be). This will unblock implementing the last remaining items from python/core-workflow#460. Thanks!

PR #1046, opened 2023-02-08
Restore `#what-s-new-and-news-entries` anchor

hugovk:
Used at least on https://blurb-it.herokuapp.com/: 

> [add a blurb](https://devguide.python.org/committing/#what-s-new-and-news-entries) to CPython pull request.


CAM-Gerlach:
Currently working on an approach to automatically detect broken fragment ids and redirect them...putting the finishing touches on a detailed proposal and will post that soon on docs-community.

Issue #1049, opened 2023-02-09
Envvars not set in subshell in macOS Python 3.7-3.9 configure invocation


thecesrom:
**Describe the bug**

I've tried building Python 3.9.16 on macOS following the dev guide and I believe there is an issue with the command as `pkg-config` fails to get the value set for `PKG_CONFIG_PATH` as, in my limited knowledge of the shell, I believe `pkg-config --libs tcl tk` runs on a subshell.

The current `configure` command produces the following output:

```sh
$ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
  LDFLAGS="-L$(brew --prefix gdbm)/lib -L$(brew --prefix xz)/lib" \
  PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
  ./configure --with-pydebug \
          --with-openssl=$(brew --prefix openssl@1.1) \
          --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
          --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
Package tcl was not found in the pkg-config search path.
Perhaps you should add the directory containing `tcl.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tcl' found
Package tk was not found in the pkg-config search path.
Perhaps you should add the directory containing `tk.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tk' found
Package tcl was not found in the pkg-config search path.
Perhaps you should add the directory containing `tcl.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tcl' found
Package tk was not found in the pkg-config search path.
Perhaps you should add the directory containing `tk.pc'
to the PKG_CONFIG_PATH environment variable
No package 'tk' found
```

**Expected behavior**

The one-line command should properly set `PKG_CONFIG_PATH`.

I've had success running the following:

```sh
$ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"; \
CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
LDFLAGS="-L$(brew --prefix gdbm)/lib -L$(brew --prefix xz)/lib" \
./configure --with-pydebug \
          --with-openssl=$(brew --prefix openssl@1.1) \
          --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
[...]

PR #892, opened 2022-06-10
Building _hashlib and _ssl modules

izumiberat:
I am using Mac OS Catalina 10.15.7.

To avoid getting the following messages : 

> The necessary bits to build these optional modules were not found:
> _hashlib              _ssl   

as well as:

> Could not build the ssl module!
> Python requires a OpenSSL 1.1.1 or newer

Based on the comments below, the solution is to indicate the OpenSSL version in the brew command: `brew install openssl@1.1` and make a note to update this line with `brew install openssl@3` when OpenSSL 3 is supported.

Initial (and outdated descritption below):
<details>
Running `brew install openssl@3` allowed the  _hashlib _ssl modules to build, but as suggested in a comment by @ezio-melotti it seemed strange that the version needed to be specified. 

I am therefore updating the documentation by leaving the suggested command line as `brew install openssl` rather than `brew install openssl@3 ` but I am adding a note saying that if user gets the message 

> Could not build the ssl module!
> Python requires a OpenSSL 1.1.1 or newer 

they can try specifying the version of OpenSSL by using `brew install openssl@3`.
</details>

Issue number: https://github.com/python/cpython/issues/93695

PR #1052, opened 2023-02-10
Update instructions for building on macOS

thecesrom:
1. Update the command for building Python 3.7 through 3.9 by exporting `PKG_CONFIG_PATH` so `pkg-config` can find `tcl` and `tk`
2. Instruct users on installing `openssl` with `brew` for building Python 3.10

Fixes: #1049

Signed-off-by: César Román <thecesrom@gmail.com>

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -413,14 +413,13 @@ Apple no longer provides header files for the deprecated system version of
 OpenSSL which means that you will not be able to build the ``_ssl`` extension.
 One solution is to install these libraries from a third-party package
 manager, like Homebrew_ or MacPorts_, and then add the appropriate paths
-for the header and library files to your ``configure`` command.  For example,
-with **Homebrew**,
+for the header and library files to your ``configure`` command.
 
-For Python 3.10 and newer::
+For example, with **Homebrew**, install the dependencies::
 
     $ brew install pkg-config openssl xz gdbm tcl-tk
 
-and ``configure``::
+Then, for Python 3.10 and newer::
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
@@ -428,11 +427,7 @@ and ``configure``::
                   --with-openssl="$(brew --prefix openssl)"
 
 
-For Python versions 3.9 through 3.7::
-
-    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
-
-and ``configure``::
+For Python 3.7 through 3.9::
 
     $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"; \
       CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -414,16 +414,16 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
 One solution is to install these libraries from a third-party package
 manager, like Homebrew_ or MacPorts_, and then add the appropriate paths
 for the header and library files to your ``configure`` command.  For example,
-
-with **Homebrew**::
+with **Homebrew**,
 
 For Python 3.10 and newer::
 
     $ brew install pkg-config openssl xz gdbm tcl-tk
 
+and ``configure``::
+
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
-      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug \
                   --with-openssl="$(brew --prefix openssl)"
 
@@ -431,6 +431,8 @@ For Python 3.10 and newer::
 For Python versions 3.9 through 3.7::
 
     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
+
+and ``configure``::
 
     $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"; \
       CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -419,7 +419,7 @@ For example, with **Homebrew**, install the dependencies::
 
     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
 
-Then, for Python 3.10 and newer::
+Then, for Python 3.10 and newer, run ``configure``::
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
@@ -438,11 +438,11 @@ Or, for Python 3.7 through 3.9::
               --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
               --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
 
-and ``make``::
+And finally, run ``make``::
 
     $ make -s -j2
 
-or **MacPorts**::
+Alternatively, with **MacPorts**::
 
     $ sudo port install pkgconfig openssl xz gdbm
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -427,7 +427,7 @@ Then, for Python 3.10 and newer::
                   --with-openssl="$(brew --prefix openssl@1.1)"
 
 
-For Python 3.7 through 3.9::
+Or, for Python 3.7 through 3.9::
 
     $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"; \
       CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -417,14 +417,14 @@ for the header and library files to your ``configure`` command.
 
 For example, with **Homebrew**, install the dependencies::
 
-    $ brew install pkg-config openssl xz gdbm tcl-tk
+    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
 
 Then, for Python 3.10 and newer::
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
       ./configure --with-pydebug \
-                  --with-openssl="$(brew --prefix openssl)"
+                  --with-openssl="$(brew --prefix openssl@1.1)"
 
 
 For Python 3.7 through 3.9::
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -417,24 +417,26 @@ for the header and library files to your ``configure`` command.  For example,
 
 with **Homebrew**::
 
-    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
-
 For Python 3.10 and newer::
+
+    $ brew install pkg-config openssl xz gdbm tcl-tk
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
       PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug \
-                  --with-openssl=$(brew --prefix openssl)
+                  --with-openssl="$(brew --prefix openssl)"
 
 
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
 One solution is to install these libraries from a third-party package
 manager, like Homebrew_ or MacPorts_, and then add the appropriate paths
 for the header and library files to your ``configure`` command.  For example,
+with **Homebrew**,
 
-with **Homebrew**::
+For Python 3.10 and newer::
```
CAM-Gerlach, line 420:
```suggestion
```

Moved

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
...
 
-    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
+    $ brew install pkg-config openssl xz gdbm tcl-tk
 
-For Python 3.10 and newer::
+and ``configure``::
```
CAM-Gerlach, line 423:
```suggestion
Then, for Python 3.10 and newer::
```

To use a more coherent structure, clearly specifying what individual steps are specific to which Python versions, and match with the above.

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
 One solution is to install these libraries from a third-party package
 manager, like Homebrew_ or MacPorts_, and then add the appropriate paths
 for the header and library files to your ``configure`` command.  For example,
+with **Homebrew**,
 
-with **Homebrew**::
+For Python 3.10 and newer::
 
-    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
+    $ brew install pkg-config openssl xz gdbm tcl-tk
```
CAM-Gerlach, line 421: Are we sure this is right?

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
 One solution is to install these libraries from a third-party package
 manager, like Homebrew_ or MacPorts_, and then add the appropriate paths
 for the header and library files to your ``configure`` command.  For example,
+with **Homebrew**,
```
CAM-Gerlach, line 417:
```suggestion
for the header and library files to your ``configure`` command.

For example, with **Homebrew**, install the dependencies::
```

I might be wrong, but it's not clear separate whether `brew` invocations are required for each thing here, as it significantly complicates the structure and makes it somewhat unclear and harder to follow here.

Furthermore, even if so, It can still be simplified, and the introductions to each code block should be consistent with one another, rather than one calling out the Python version and the other the name of the step/tool, when in fact the former applies to both.

And its worth putting the introduction to the example on a separate line from this long and complex paragraph.

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
...
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
-      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug \
-                  --with-openssl=$(brew --prefix openssl)
+                  --with-openssl="$(brew --prefix openssl)"
 
 
 For Python versions 3.9 through 3.7::
```
CAM-Gerlach, line 431:
```suggestion
For Python 3.7 through 3.9::
```

Eliminate redundant use of version inconsistent with the above, and put the versions in the conventional, expected order

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
...
 
-    $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
+    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
+
+and ``configure``::
+
```
CAM-Gerlach, line 436:
```suggestion
```

Hopefully this bit can be consolidated above; otherwise, we'll have to do some more significant restructuring of this section to make it properly clear

getting-started/setup-building.rst
```diff
@@ -414,27 +414,31 @@ OpenSSL which means that you will not be able to build the ``_ssl`` extension.
...
+    $ export PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig"; \
+      CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -L$(brew --prefix xz)/lib" \
-      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug \
-              --with-openssl=$(brew --prefix openssl@1.1) \
+              --with-openssl="$(brew --prefix openssl@1.1)" \
               --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
               --with-tcltk-includes="$(pkg-config --cflags tcl tk)"
```
CAM-Gerlach, line 444: Can't reach it with a suggestion, but the current syntax isn't explicit that "and make" applies to both branches, rather than just the latter as the structure would imply (particularly prior to my suggestions). I would also make similar clarifications with the "MacPorts" section, though maybe getting a little out of scope here.

CAM-Gerlach (commented):
Some suggestions and questions to clarify the advice and the structure

CAM-Gerlach:
(**Standard reminder: You can directly apply all the suggestions you want in one go with [`Files changed`](https://github.com/python/devguide/pull/1052/files) -> <kbd>Add to batch</kbd> -> <kbd>Commit</kbd>**)

getting-started/setup-building.rst
thecesrom: I added this line because the guide currently recommends installing `openssl@1.1`, but for `configure` we see the following:

`--with-openssl=$(brew --prefix openssl)`

Furthermore, for Homebrew `openssl` is an alias for `openssl@3`, not `openssl@1.1`.

I could either revert back the `brew install` command to its original form, but if that's the case, then I believe the best option would be to also use `openssl@1.1` for `--with-openssl` for Python 3.10 and newer to avoid confusion, i.e. `--with-openssl=$(brew --prefix openssl@1.1)`.

thecesrom: Same thing for MacPorts, `openssl` is an alias for `openssl3`.

thecesrom: Python 3.10 and above require OpenSSL 1.1.1 or newer, so both `openssl@3` and `openssl@1.1` in Homebrew or `openssl3` and `openssl11` in MacPorts work for building Python, I don't know if OpenSSL 3 might work for Python 3.7 through 3.9.

ezio-melotti: FWIW this was changed recently -- have you double checked `git blame` and the discussion in the related PR?

erlend-aasland: What @ezio-melotti said. Please have a look at the recent discussion. Check the closed issues and merged PRs for this repo.

thecesrom: @ezio-melotti, I found PR https://github.com/python/devguide/pull/892, and they only modified the package to install via `brew` from `openssl` to `openssl@1.1`, but for `configure` they're still referencing `openssl` (a.k.a. `openssl3`) instead of `openssl@1.1`.

So this is what I'm suggesting

Revert the `brew` command to: 

```sh
$ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
```

And update to `configure` options to:

```sh
$ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
   LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
   ./configure --with-pydebug \
                      --with-openssl="$(brew --prefix openssl@1.1)"
```

getting-started/setup-building.rst
```diff
@@ -413,28 +413,27 @@ Apple no longer provides header files for the deprecated system version of
...
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
-      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
       ./configure --with-pydebug \
-                  --with-openssl=$(brew --prefix openssl)
+                  --with-openssl="$(brew --prefix openssl@1.1)"
 
 
-For Python versions 3.9 through 3.7::
+For Python 3.7 through 3.9::
```
CAM-Gerlach, line 430:
```suggestion
Or, for Python 3.7 through 3.9::
```

Slight revision to my previous—be a little more explicit that this is an alternative path and flow better with the previous.

getting-started/setup-building.rst
```diff
@@ -413,28 +413,27 @@ Apple no longer provides header files for the deprecated system version of
...
+For example, with **Homebrew**, install the dependencies::
 
     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
 
-For Python 3.10 and newer::
+Then, for Python 3.10 and newer::
 
     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
-      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
```
ned-deily, line 426: Alas, contrary to my comment on the issue, the PKG_CONFIG_PATH line here is still needed for 3.10+ because of Homebrew's policy of not conflicting with Apple-supplied third-party libs, even though the Apple Tcl/Tk is deprecated and essentially useless. So this line should not be deleted.

ned-deily (approved):

getting-started/setup-building.rst
CAM-Gerlach: Seems like this all works and is approved by Ned, so marking this as resolved

CAM-Gerlach (commented):
This looks good from my side, except that as [mentioned previously](https://github.com/python/devguide/pull/1052#discussion_r1103323042), the immediately following introductory text (the last make step and the first line for MacPorts after it) should be updated to make sense with the changes here and ensure it is clearer for readers which instructions are specific to which distributor and Python version.

I can't make them as suggestions, but I can either just push a commit directly to your branch, or you can `apply` this diff:

<details>

<summary>``git diff`` output</summary>

```diff
diff --git a/getting-started/setup-building.rst b/getting-started/setup-building.rst
index b285a74..15e094a 100644
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -419,7 +419,7 @@ For example, with **Homebrew**, install the dependencies::

     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk

-Then, for Python 3.10 and newer::
+Then, for Python 3.10 and newer, run ``configure``::

     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
@@ -438,11 +438,11 @@ Or, for Python 3.7 through 3.9::
               --with-tcltk-libs="$(pkg-config --libs tcl tk)" \
               --with-tcltk-includes="$(pkg-config --cflags tcl tk)"

-and ``make``::
+And finally, run ``make``::

     $ make -s -j2

-or **MacPorts**::
+Alternatively, with **MacPorts**::

     $ sudo port install pkgconfig openssl xz gdbm
```

</details>

Or `am` this patch:

<details>

<summary>``git format-patch`` output</summary>

```patch
From 8ae83afc0cbf616c1937408c8e5767ecd4625cf1 Mon Sep 17 00:00:00 2001
From: "C.A.M. Gerlach" <CAM.Gerlach@Gerlach.CAM>
Date: Wed, 15 Feb 2023 20:01:14 -0600
Subject: [PATCH] Update following text to match revised macOS building
 language

---
 getting-started/setup-building.rst | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/getting-started/setup-building.rst b/getting-started/setup-building.rst
index b285a74..15e094a 100644
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -419,7 +419,7 @@ For example, with **Homebrew**, install the dependencies::

     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk

-Then, for Python 3.10 and newer::
+Then, for Python 3.10 and newer, run ``configure``::

     $ CFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
       LDFLAGS="-L$(brew --prefix gdbm)/lib -I$(brew --prefix xz)/lib" \
@@ -438,11 +438,11 @@ Or, for Python 3.7 through 3.9::

-and ``make``::
+And finally, run ``make``::

     $ make -s -j2

-or **MacPorts**::
+Alternatively, with **MacPorts**::

     $ sudo port install pkgconfig openssl xz gdbm

--
2.37.1.windows.1

```

</details>

thecesrom:
I'll work on that

thecesrom:
@CAM-Gerlach patch has been applied

pradyunsg:
Thanks @thecesrom!

PS: Merged this since this has addressed all outstanding comments, has an approval from the relevant expert, and existing changes look good to me!

erlend-aasland:
Thanks for handling this CAM and Ned!

getting-started/setup-building.rst
CAM-Gerlach: Implemented in followup review

PR #1053, opened 2023-02-16
Add Android platform to the experts list

FFY00:
We support Android, as best-effort, I believe. I don't think we have anyone leading it though, right?

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -297,6 +297,7 @@ Platforms
 Platform              Maintainers
 ===================   ===========
 AIX                   David.Edelsohn^
+Android
 Cygwin                jlt63^, stutzbach^
 FreeBSD
 HP-UX


CAM-Gerlach:
IIRC, it was previously better-supported but the support has rotted over time without an active maintainer. Since we do have buildbots, if someone had signed up to maintain it, it could potentially be made a [Tier 3 platform per PEP 11](https://peps.python.org/pep-0011/#tier-3).

ezio-melotti (commented):
I'd say to wait until Android is added to PEP 11 and/or has a maintainer before adding it to the platform list.

CAM-Gerlach:
To note, the Experts index does include platforms that are supported by non-core devs and explicitly not supported per PEP 11; see [Petr's comment](https://github.com/python/devguide/issues/848#issuecomment-1200963008) on your #848 . However, IMO it should at least have an active maintainer of some sort (core dev or external) to be newly added there.

hugovk:
On the other hand, someone may see the empty row and step up to fill it.

(But no strong opinion either way.)

Mariatta (approved):

Mariatta:
I agree with adding it to the list as a category and letting people step up to fill it.

eamanu (approved):

PR #1058, opened 2023-02-22
Use targeted env vars for overriding compiler/linker args on macOS

erlend-aasland:


CAM-Gerlach (commented):
FWIW, this looks good from a docs perspective—ultimately up to the subject matter experts (Ned, etc) of course.

ned-deily:
I like this approach but, as I've just discovered, Homebrew has just changed the game again by making their tcl-tk formula no longer "keg-only". As noted [here](https://github.com/python/cpython/pull/102186#issuecomment-1445464817), we should re-think this a bit in that light. I'll have some more thoughts on it soon.

erlend-aasland:
> I like this approach but, as I've just discovered, Homebrew has just changed the game again by making their tcl-tk formula no longer "keg-only". As noted [here](https://github.com/python/cpython/pull/102186#issuecomment-1445464817), we should re-think this a bit in that light. I'll have some more thoughts on it soon.

Ouch, it's unfortunate that our build instructions are so dependent on Homebrew distribution specifics. Let's create a separate issue for this (unless you already did so).

getting-started/setup-building.rst
```diff
@@ -419,7 +419,17 @@ For example, with **Homebrew**, install the dependencies::
 
     $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
 
-Then, for Python 3.10 and newer, run ``configure``::
+Then, for Python 3.11 and newer, run ``configure``::
```
ezio-melotti, line 422: With all these versions, it might be better to use that Sphinx extension that has different tabs on the code block

CAM-Gerlach: Yeah, I was thinking that too having looked at it. It would require adding a dep, but it could be likely useful in other places where the instructions differ by version, platform, etc.—which is quite common in the devguide.

erlend-aasland: SGTM

erlend-aasland:
> As noted https://github.com/python/cpython/pull/102186#issuecomment-1445464817, we should re-think this a bit in that light. I'll have some more thoughts on it soon.

I've marked this as a draft until the above has landed.

erlend-aasland:
> I like this approach but, as I've just discovered, Homebrew has just changed the game again by making their tcl-tk formula no longer "keg-only". As noted [here](https://github.com/python/cpython/pull/102186#issuecomment-1445464817), we should re-think this a bit in that light. I'll have some more thoughts on it soon.

@ned-deily, have you had the chance to think about this issue? IMO, it is unfortunate that we rely on Homebrew specifics. Perhaps the best thing to do, is to update the devguide recommendations for how to install the dependencies:

```console
$ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
$ brew link tcl-tk  # <= may produce a warning if already linked
```

We can then carry on with this PR (and also update the devguide again).

ned-deily:
> @ned-deily, have you had the chance to think about this issue? IMO, it is unfortunate that we rely on Homebrew specifics. Perhaps the best thing to do, is to update the devguide recommendations for how to install the dependencies:

That could be an option. There are a number of other side issues with some of the other dependencies, like which ones should we use from the operating system, and there are the Homebrew differences between Intel and Apple Silicon installations. Alas, I am away right now but will get back to this in a couple of weeks. If you think there is a need to update the guide sooner, we can always tweak it again later.

erlend-aasland:
No, there's no hurry 🙂 If you're coming to SLC, we can discuss it there. If not, let's pick it up when you're ready.

getting-started/setup-building.rst
```diff
@@ -419,7 +419,17 @@ For example, with **Homebrew**, install the dependencies::
...
+
+    $ GDBM_CFLAGS="-I$(brew --prefix gdbm)/include" \
+      GDBM_LIBS="-L$(brew --prefix gdbm)/lib -lgdbm" \
+      LIBLZMA_CFLAGS="-I$(brew --prefix xz)/include" \
+      LIBLZMA_LIBS="-I$(brew --prefix xz)/lib -llzma" \
+      PKG_CONFIG_PATH="$(brew --prefix tcl-tk)/lib/pkgconfig" \
+      ./configure --with-pydebug \
+                  --with-openssl="$(brew --prefix openssl@1.1)"
+
+For Python 3.10::
```
ned-deily, line 432: I've finally made some time to test and experiment with Homebrew on both Intel and Apple Silicon Macs with 3.12 through 3.7. The good news is that I think we only need two, somewhat simpler recipes now:

for 3.11+:

```
GDBM_CFLAGS="-I$(brew --prefix gdbm)/include" \
GDBM_LIBS="-L$(brew --prefix gdbm)/lib -lgdbm" \
./configure --with-pydebug \
                  --with-openssl="$(brew --prefix openssl@1.1)"

```
For 3.10 through 3.7:

```
CPPFLAGS="-I$(brew --prefix gdbm)/include -I$(brew --prefix xz)/include" \
LDFLAGS="-L$(brew --prefix gdbm)/lib -L$(brew --prefix xz)/lib" \
./configure --with-pydebug \
                  --with-openssl="$(brew --prefix openssl@1.1)" \

```

(The `--with-tcltk*` options aren't needed for 3.10, just for 3.9 and earlier, but they don't hurt for 3.10.)

This was with the current version of Homebrew and macOS (Ventura 13.3.1).  Additional testing in other environments would be good :) In particular, I did not attempt to test using `brew link`. And, obviously, this is a basic example; there's a lot more we *could* say about overriding Apple-supplied packages, licenses, etc. But that's something for another day.

The MacPorts recipes should also be similarly updated.

`sudo port install pkgconfig openssl11 xz gdbm tcl tk +quartz
`

3.11+:

```
GDBM_CFLAGS="-I$(dirname $(dirname $(which port)))/include" \
GDBM_LIBS="-L$(dirname $(dirname $(which port)))/lib -lgdbm" \
./configure --with-pydebug \
                  --with-openssl="$(dirname $(dirname $(which port)))/libexec/openssl11"

```

3.10 through 3.7:

```
CPPFLAGS="-I$(dirname $(dirname $(which port)))/include" \
LDFLAGS="-L$(dirname $(dirname $(which port)))/lib" \
./configure --with-pydebug \
                  --with-openssl="$(dirname $(dirname $(which port)))/libexec/openssl11" \

```

I didn't want to screw up the discussion here by pushing an update to the PR but I could do that if it would help.

ned-deily (changes requested):

getting-started/setup-building.rst
erlend-aasland: Thanks, Ned! I'll have a look at this tonight. I can update the PR and the docs (I'll request your review).

CAM-Gerlach: Hey @erlend-aasland , did you ever get a chance to get back to this? Thanks!

erlend-aasland: No, I got bogged down with other stuff and it slipped away from my mind. Thanks for the reminder!

CAM-Gerlach: I know that very well :sweat_smile:

getting-started/setup-building.rst
erlend-aasland: Let's keep this formatting improvement out of this PR. If we want to add the inline tabs extension, we can install that in a follow-up PR and then start to make use of it.

erlend-aasland:
@ned-deily, let's land this. We can handle the OpenSSL version in a follow-up PR (https://github.com/python/cpython/issues/99079#issuecomment-1581927371)

getting-started/setup-building.rst
erlend-aasland: Tested on 3.7 through main on macOS 13.4; works swell!

ned-deily (approved):
LGTM, thanks!

PR #1060, opened 2023-02-28
PEP 689: Add docs for unstable C API

encukou:


developer-workflow/c-api.rst
```diff
@@ -4,18 +4,20 @@
...
 
 1. The internal, private API, available with ``Py_BUILD_CORE`` defined.
    Ideally declared in ``Include/internal/``. Any API named with a leading
    underscore is also considered private.
-2. The public C API, available when ``Python.h`` is included normally.
+2. The “general” public C API, available when ``Python.h`` is included normally.
    Ideally declared in ``Include/cpython/``.
-3. The Limited API, available with ``Py_LIMITED_API`` defined.
+3. The Unstable C API, identified by the ``PyUnstable_`` name prefix.
+   Ideally declared in ``Include/cpython/`` along with the general public API.
```
CAM-Gerlach, line 15: The rest of the text should probably be updated (presumably in a separate PR) to use it too, but thanks to Ezio's `:cpy-file:` role we added, you can now do

```suggestion
   Ideally declared in :cpy-file:`Include/cpython/` along with the general public API.
```

to get a helpful link directly to the referred to source code, which can be linkchecked and flagged automatically if it gets out of date and breaks (so the doc can be updated accordingly).

developer-workflow/c-api.rst
```diff
@@ -4,18 +4,20 @@
...
+4. The Limited C API, available with ``Py_LIMITED_API`` defined.
```
CAM-Gerlach, line 16:
```suggestion
4. The Limited C API, available with :c:macro:`Py_LIMITED_API` defined.
```

Thanks to the intersphinx support we added, you can auto-format and cross-reference this macro directly here.

developer-workflow/c-api.rst
```diff
@@ -29,13 +31,18 @@ for building CPython itself, as indicated by a macro like ``Py_BUILD_CORE``.
 
 While internal API can be changed at any time, it's still good to keep it
 stable: other API or other CPython developers may depend on it.
+For users, internal API is sometimes the best a workaround for a thorny problem
```
CAM-Gerlach, line 34: Like many instances in this doc, "API" is missing an article here, but it's not totally clear which it should be, given the existing semantics not clear (clarifying which being one reason the article is needed in the first place).

If you mean the internal API as a whole here, this should be

```suggestion
For users, the internal API is sometimes the best a workaround for a thorny problem
```

Or if you mean a particular internal API, this should be

```suggestion
For users, internal API is sometimes the best a workaround for a thorny problem
```

developer-workflow/c-api.rst
```diff
@@ -29,13 +31,18 @@ for building CPython itself, as indicated by a macro like ``Py_BUILD_CORE``.
...
+-- though those use cases should be discussed on `Discourse <https://discuss.python.org/c/30>`_
```
CAM-Gerlach, line 35:
```suggestion
--- though those use cases should be discussed on the
`C API Discourse category <https://discuss.python.org/c/30>`_
```

Both for a11y purposes and for all users, the link title should be appropriately accurate and descriptive—this is a link to the C API Discourse (itself currently a sub-category of the Core Dev category), not to the Discourse instance as a whole as the current text implies.

Also, since I changed the line anyway, fixed a en dash that should have been an em dash

developer-workflow/c-api.rst
```diff
@@ -53,14 +60,18 @@ Private names
 --------------
 
 Any API named with a leading underscore is also considered internal.
-There are two main use cases for using such names rather than putting the
-definition in ``Include/internal/`` (or directly in a ``.c`` file):
+There is currently only one main use case for using such names rather than
+putting the definition in ``Include/internal/`` (or directly in a ``.c`` file):
+
+* Internal helpers for other public API, which users should not call directly.
```
CAM-Gerlach, line 66:
```suggestion
* Internal helpers for other public APIs, which users should not call directly.
```

Missing `s` for plural (like most but not quite all words in English, the [well-established plural of API is APIs](https://english.stackexchange.com/questions/323688/what-is-the-plural-of-api)—this seems to be a pervasive mistake in this document generally)

developer-workflow/c-api.rst
```diff
@@ -53,14 +60,18 @@ Private names
 --------------
 
 Any API named with a leading underscore is also considered internal.
-There are two main use cases for using such names rather than putting the
-definition in ``Include/internal/`` (or directly in a ``.c`` file):
+There is currently only one main use case for using such names rather than
+putting the definition in ``Include/internal/`` (or directly in a ``.c`` file):
```
CAM-Gerlach, line 64:
```suggestion
putting the definition in :cpy-file:`Include/internal/` (or directly in a ``.c`` file):
```

To get a link and automatic checking if it breaks

developer-workflow/c-api.rst
```diff
@@ -53,14 +60,18 @@ Private names
...
+
+* Internal helpers for other public API, which users should not call directly.
+
+Note that historically, underscores were used for API that is better served by
```
CAM-Gerlach, line 68:
```suggestion
Note that historically, underscores were used for APIs that are better served by
```

Fix plural spelling and singular-plural agreement problem (which initially confused me as to whether you meant it as plural at all)

developer-workflow/c-api.rst
```diff
@@ -53,14 +60,18 @@ Private names
...
+the :ref:`unstable-capi`:
+
+* “provisional” API, included in a Python release to test real-world
+  usage of new API;
+* API for very specialized uses like JIT compilers.
```
CAM-Gerlach, line 73:
```suggestion
* “provisional” APIs, included in a Python release to test real-world
  usage of new APIs;
* APIs for very specialized uses like JIT compilers.
```

Likewise

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
 ``PyExc_AssertionError`` unless actually testing custom exceptions.
 
 
+.. _unstable-capi:
+
+Unstable C API
+==============
```
CAM-Gerlach, line 151: I'm rather confused why this was added between Public C API and Limited C API, instead of between internal Internal API and Public C API. Isn't it specifically stated to be a step between the latter two, not the former two, so shouldn't it be moved there instead?

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+The unstable C API tier is meant for extensions that need tight integration
+with the interpreter, like debuggers and JIT compilers.
+Users of this tier may need to change their code with every minor release.
+
+In many ways, this tier is like the general C API:
+
+- it's available when ``Python.h`` is included normally,
+- it should be defined  in ``Include/cpython/``,
```
CAM-Gerlach, line 160:
```suggestion
- it should be defined  in :cpy-file:`Include/cpython/`,
```

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+- it requires tests, so we don't break it unintentionally
+- it requires docs, so both we and the users
```
CAM-Gerlach, line 162:
```suggestion
- it requires docs, so both we and the users,
```

Missing comma like the others have

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  period.
+- A stability note is appears in the docs.
+  This happens automatically, based on the name
+  (via :cpy-file:`Doc/tools/extensions/c_annotations.py`).
+
+Despite being “unstable”, there are rules to make sure third-party code can
+use this API reliably:
+
+* Changes and removals can be done in minor releases (3.x.0, including
+  Alphas and Betas for 3.x.0).
```
CAM-Gerlach, line 180:
```suggestion
* Changes and removals can be done in minor releases
  (:samp:`3.{X}.0`, including Alphas and Betas for :samp:`3.{X}.0`).
```

* Use `:samp:` to explicitly mark up and format placeholder
* Follow typical convention (at least that I've seen) to use capitals for placeholder letters in versions (with lowercase being a literal `x`)

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  can agree on the expected behavior,
+- it is tested and documented in the same way.
+
+The differences are:
+
+- Names of functions structs, macros, etc. start with the ``PyUnstable_``
+  prefix. This defines what's in the unstable tier.
+- The unstable API can change in minor versions, without any deprecation
+  period.
+- A stability note is appears in the docs.
```
CAM-Gerlach, line 172:
```suggestion
- A stability note appears in the docs.
```

Typo

CAM-Gerlach:
Freaking GitHub, every long review I've done lately it's silently auto-submitted it (without the top level review comment) right the middle of the review for no obvious reason, with no special keyboard shortcut on my end (as far as I'm aware) and no changes on the PR's end. It's immeasurably frustrating and I'm sick and tired of it but I can't figure out if it's GitHub's fault, GitHub Refined's fault or something I'm something I'm somehow doing.

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  This happens automatically, based on the name
+  (via :cpy-file:`Doc/tools/extensions/c_annotations.py`).
+
+Despite being “unstable”, there are rules to make sure third-party code can
+use this API reliably:
+
+* Changes and removals can be done in minor releases (3.x.0, including
+  Alphas and Betas for 3.x.0).
+* Adding new unstable API *for an existing feature* is allowed even after
```
CAM-Gerlach, line 181:
```suggestion
* Adding a new unstable API *for an existing feature* is allowed even after
```

Add missing indefinite article

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  Beta feature freeze, up until the first Release Candidate.
+  Consensus on Core Development Discourse or is needed in the Beta period.
```
CAM-Gerlach, line 183:
```suggestion
  Consensus on the `Core Development Discourse <https://discuss.python.org/c/core-dev/23>`_
  is needed in the Beta period.
```

* Add link
* Fix stray "or", presumably an editing oversight

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+* Backwards-incompatible changes should make existing C callers fail to compile.
+  For example, arguments should be added/removed, or a function should be
+  renamed.
+* When moving API into or out of the Unstable tier, the old name
```
CAM-Gerlach, line 187:
```suggestion
* When moving an API into or out of the Unstable tier, the old name
```

Add missing article

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  should continue to be available (but deprecated) until an incompatible
+  change is made. In other words, while we're allowed to break calling code,
+  we shouldn't break it *unnecessarily*.
+
+
+Moving API from the public tier to unstable
+-------------------------------------------
```
CAM-Gerlach, line 194: This either needs an article, or a plural (`s`). Since the section is framed in terms of a single API, I've suggested the former:

```suggestion
Moving an API from the public tier to unstable
----------------------------------------------
```

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+* Expose the API under its new name, with the ``PyUnstable_`` prefix.
+* Make the old name an alias (e.g. a ``static inline`` function calling the
+  new function).
+* Deprecate the old name, typically using ``Py_DEPRECATED``.
```
CAM-Gerlach, line 199:
```suggestion
* Deprecate the old name, typically using :c:macro:`Py_DEPRECATED`.
```

Cross-reference and directly link to the docs on the macro and how to use it.

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+* Announce the change in What's New.
+
+The old name should continue to be available until an incompatible change is
+made. Per Python’s backwards compatibility policy (:pep:`387`),
+this deprecation needs to last at least two releases (modulo SC exceptions).
+
+The rules are relaxed for API that was introduced before the official
+Unstable tier.
```
CAM-Gerlach, line 207:
```suggestion
The rules are relaxed for APIs that were introduced in Python versions
before 3.12, when the official Unstable tier was added.
```

* Clarify more specifically which APIs qualify per the below
* Presumably you're referring to multiple APIs here, so this should be plural

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+For the following, you can make an incompatible change (and remove the old name)
+as if the function was already part of the Unstable tier:
```
CAM-Gerlach, line 209:
```suggestion
You can make an incompatible change (and remove the old name)
as if the function was already part of the Unstable tier
for APIs introduced before Python 3.12 that are either:
```

Re-order the phrasing to make more sense, improve clarity and avoid the redundant and less easy to read duplication below

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+* API introduced before Python 3.12 that is documented to be less stable than default.
+* API introduced before Python 3.12 that was named with a leading underscore.
```
CAM-Gerlach, line 212:
```suggestion
* Documented to be less stable than default.
* Named with a leading underscore.
```

Remove redundant duplication per above

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+Moving API from the private tier to unstable
+--------------------------------------------
```
CAM-Gerlach, line 215:
```suggestion
Moving an API from the private tier to unstable
-----------------------------------------------
```

Likewise, fix grammar here

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+* Expose the API under its new name, with the ``PyUnstable_`` prefix.
+* If the old name is documented, or widely used externally,
+  make it an alias and deprecate it (typically with ``Py_DEPRECATED``).
```
CAM-Gerlach, line 219:
```suggestion
  make it an alias and deprecate it (typically with :c:macro:`Py_DEPRECATED`).
```

Likewise, cross-reference and provide a helpful link to this macro (and if it changes, a warning will be issued)

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+  It should continue to be available until an incompatible change is made,
+  as if it was previously public.
+
+  This applies even to underscored names. Python wasn't always strict with
+  the leading underscore.
+
+* Announce the change in What's New.
+
+Moving API from unstable to public
+----------------------------------
```
CAM-Gerlach, line 229:
```suggestion
Moving an API from unstable to public
-------------------------------------
```

Likewise, fix grammar

CAM-Gerlach (commented):
Add remaining comments as review suggestions.

(***Standard reminder: You can directly apply all the suggestions you want in one go with [`Files changed`](https://github.com/python/REPO/pull/NUMBER/files) -> <kbd>Add to batch</kbd> -> <kbd>Commit</kbd>***)

developer-workflow/c-api.rst
```diff
@@ -4,18 +4,20 @@
...
 =======================
 
-The C API is divided into three sections:
+The C API is divided into these tiers:
 
 1. The internal, private API, available with ``Py_BUILD_CORE`` defined.
    Ideally declared in ``Include/internal/``. Any API named with a leading
    underscore is also considered private.
-2. The public C API, available when ``Python.h`` is included normally.
+2. The “general” public C API, available when ``Python.h`` is included normally.
```
ezio-melotti, line 12:
```suggestion
2. The "general" public C API, available when :cpy-file:`Include/Python.h` is included normally.
```
Should ASCII quotes be preferred? IIRC Sphinx already converts them to fancy quotes in the output.
Since this is the first time that `Python.h` is mentioned, it might be worth to include the full path and turn it into a link.

developer-workflow/c-api.rst
ezio-melotti: There's also a typo here "the best a workaround".

developer-workflow/c-api.rst
```diff
@@ -53,14 +60,18 @@ Private names
...
-definition in ``Include/internal/`` (or directly in a ``.c`` file):
+There is currently only one main use case for using such names rather than
+putting the definition in ``Include/internal/`` (or directly in a ``.c`` file):
+
+* Internal helpers for other public API, which users should not call directly.
+
+Note that historically, underscores were used for API that is better served by
+the :ref:`unstable-capi`:
+
+* “provisional” API, included in a Python release to test real-world
```
ezio-melotti, line 71:
```suggestion
* "provisional" API, included in a Python release to test real-world
```

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+- Names of functions structs, macros, etc. start with the ``PyUnstable_``
+  prefix. This defines what's in the unstable tier.
+- The unstable API can change in minor versions, without any deprecation
+  period.
+- A stability note is appears in the docs.
+  This happens automatically, based on the name
+  (via :cpy-file:`Doc/tools/extensions/c_annotations.py`).
+
+Despite being “unstable”, there are rules to make sure third-party code can
```
ezio-melotti, line 176:
```suggestion
Despite being "unstable", there are rules to make sure third-party code can
```

developer-workflow/c-api.rst
ezio-melotti: Other occurrences of "Unstable" are capitalized, so maybe this should be too.

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+
+Moving API from the public tier to unstable
+-------------------------------------------
+
+* Expose the API under its new name, with the ``PyUnstable_`` prefix.
+* Make the old name an alias (e.g. a ``static inline`` function calling the
+  new function).
+* Deprecate the old name, typically using ``Py_DEPRECATED``.
+* Announce the change in What's New.
```
ezio-melotti, line 200:
```suggestion
* Announce the change in the "What's New".
```

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+The old name should continue to be available until an incompatible change is
+made. Per Python’s backwards compatibility policy (:pep:`387`),
+this deprecation needs to last at least two releases (modulo SC exceptions).
```
ezio-melotti, line 204:
```suggestion
this deprecation needs to last at least two releases
(modulo :abbr:`SC (Steering Council)` exceptions).
```

developer-workflow/c-api.rst
```diff
@@ -134,6 +145,101 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+
+* Expose the API under its new name, with the ``PyUnstable_`` prefix.
+* If the old name is documented, or widely used externally,
+  make it an alias and deprecate it (typically with ``Py_DEPRECATED``).
+  It should continue to be available until an incompatible change is made,
+  as if it was previously public.
+
+  This applies even to underscored names. Python wasn't always strict with
+  the leading underscore.
+
```
ezio-melotti, line 225: Should this empty line be here before the last item?
```suggestion
```

developer-workflow/c-api.rst
CAM-Gerlach: > Should ASCII quotes be preferred? IIRC Sphinx already converts them to fancy quotes in the output.

Yeah I prefer them personally as they make find and replace easier, are slightly more efficient to store avoid accidentally copying the wrong quote and work equally well in code and prose text (with Sphinx formatting them appropriately), but as I've been trying to be less nitpicky about minor typographical details that don't make a different to the semantics or the output, I didn't mention it above.

>  Since this is the first time that Python.h is mentioned, it might be worth to include the full path and turn it into a link.

Yeah, good idea—I thought of that but didn't end up remembering to make the suggestion

developer-workflow/c-api.rst
CAM-Gerlach: Thanks, edited to fix above

developer-workflow/c-api.rst
CAM-Gerlach: Or simply

```suggestion
this deprecation needs to last at least two releases
(modulo Steering Council exceptions).
```

developer-workflow/c-api.rst
encukou: An “application programming interface” is (roughly) a *collection* of functions, types, macros, etc. Ideally there'd be a good general word for one of those.
*API* is the best I know, but it's overloading the term in a confusing way.

developer-workflow/c-api.rst
encukou: My reasoning is that the description builds on top of the public API section. If you're reading the whole document, the references would make you jump around so you'd end up with the order in this PR.
And if you're not reading from top to bottom, does the order matter?

developer-workflow/c-api.rst
encukou: I'm not aware of that convention. I've always used lowercase. (And I like it because it stands out among the numbers...)
Should https://docs.python.org/3.10/c-api/stable.html be updated?

developer-workflow/c-api.rst
encukou: I strongly prefer using English quotes when writing English.

developer-workflow/c-api.rst
encukou: `:samp:` looks useful! TIL. thanks!

developer-workflow/c-api.rst
encukou: As I see it, these arguments for `"` are exactly the same as arguments for one vs. two spaces after a period. For that there was a [recent decision](https://github.com/python/peps/pull/3027) to just use whatever the author prefers. So I'll do that. (I'm consciously trying to train my fingers to type the proper quotes for the language I'm writing: “English”, „Czech“, `'python'` or `"C"`.)
If you disagree, I don't think this PR is the right place for that discussion.

developer-workflow/c-api.rst
CAM-Gerlach: Yeah; the real linguistic issue is there's no real noun in English that spans for all the _elements_ (functions, types, macros, etc) that together fall under an API. It's not ideal, but its standard convention and the best we have.

developer-workflow/c-api.rst
CAM-Gerlach: Hmm, maybe it's just what I've seen then—particularly since I've had to use a version placeholder in a decent number of instances that also contain a literal `x`, such as branch names (`X.Y.x`). In any case, here's a version of the suggestion with a lowercase `x`:

```suggestion
* Changes and removals can be done in minor releases
  (:samp:`3.{x}.0`, including Alphas and Betas for :samp:`3.{x}.0`).
```

developer-workflow/c-api.rst
CAM-Gerlach: > I strongly prefer using English quotes when writing English.

> (I'm consciously trying to train my fingers to type the proper quotes for the language I'm writing: “English”, „Czech“, `'python'` or `"C"`.)

Docutils/Sphinx will automatically convert standard ASCII straight quotes in the source files to the correct quote character for the configured document language (or leave them verbatim, for literal code blocks), so spending the extra effort of manually hardcoding the desired output typographical quote just bakes that complexity into the source and increases the probability of making a mistake (wrong quote char, wrong direction, etc) for no real practical gain.

That said, as mentioned above (and also in the discussions you linked), I don't see it as important enough to request changing in a review, much less argue about. :)

developer-workflow/c-api.rst
CAM-Gerlach: Hmm, I see. I understand the rationale there, though IMO heading order still matters both for folks consulting specific sections/subsections (which are likely the majority of readers) and those reading the whole document sequentially, as it establishes a clear order and hierarchy of the API levels and the desired precedence for usage (limited, public, unstable, internal) rather confusingly implying a structure that doesn't reflect their actual conceptual organization, and requiring users to read the full text to ensure they know the correct order.

In light of that, since the Limited API section doesn't really directly refer much to the Public C API section and is the highest-level tier (rather than the lowest, as the current order implies), what about instead switching it with the Internal API, which would make the order of the page from highest-level tier to lowest-level tier, keeping the Unstable API listed after Public C API as you have it now, but making a lot more sense overall. I.e.:

* Limited API
    * Guidelines for changing the Limited API
    * ...
 * Public C API
    * C API Tests
* Unstable C API
    * Moving an API from the public tier to Unstable
    * ...
* Internal API
    * With PyAPI_FUNC or PyAPI_DATA
    * ...

developer-workflow/c-api.rst
```diff
@@ -4,18 +4,20 @@
 Changing Python's C API
 =======================
 
-The C API is divided into three sections:
+The C API is divided into these tiers:
```
CAM-Gerlach, line 7: Same comment with the order here, only more crucial—"tiers" explicitly denotes an ordered hierarchy, and yet internal API is listed and numbered first, then public API, then unstable API, then Limited API, which directly implies that this is the order of the tiers. Swapping "internal" and "limited" would fix this (as would swapping "general" and "unstable", as I originally suggested, but wouldn't match the order in the document, and going from highest-level/most recommended to lowest-level/least recommended makes more sense anyway).

developer-workflow/c-api.rst
```diff
@@ -4,18 +4,20 @@
...
    Ideally declared in ``Include/cpython/``.
-3. The Limited API, available with ``Py_LIMITED_API`` defined.
+3. The Unstable C API, identified by the ``PyUnstable_`` name prefix.
+   Ideally declared in :cpy-file:`Include/cpython/` along with the general public API.
+4. The Limited C API, available with :c:macro:`Py_LIMITED_API` defined.
    Ideally declared directly under ``Include/``.
 
-Each section has higher stability & maintenance requirements, and you will
-need to think about more issues when you add or change definitions in it.
+Each tier has different stability & maintenance requirements to think about
```
FFY00, line 19: nitpick
```suggestion
Each tier has different stability & maintenance requirements to consider
```

FFY00 (commented):
I think it makes sense to

developer-workflow/c-api.rst
CAM-Gerlach: Slightly more conventional/idiomatic, though small and subjective enough that I didn't mention it. But if we're going to nitpick anyway, typically "&" is only used as a replacement for "and" in English in [very specific contexts](https://en.wikipedia.org/wiki/Ampersand#Usage and carries a somewhat distinct meaning

developer-workflow/c-api.rst
ncoghlan: I think the ordering issue is real, but I think trying to resolve it directly in this PR will make it harder to review the content changes.
My suggestion would be to instead file a docs enhancement request to restructure the API tier docs to run from either most-to-least or least-to-most portable, rewording the individual sections as necessary.

developer-workflow/c-api.rst
ncoghlan: The suggested restructuring sounds plausible to me, but also feels like it would be better served by being done as its own PR rather than both adding new content and restructuring the existing content in this PR.

developer-workflow/c-api.rst
```diff
@@ -134,6 +146,103 @@ When moving existing tests, feel free to replace ``TestError`` with
...
+Moving an API from unstable to public
+-------------------------------------
+
+* Expose the API under its new name, without the ``PyUnstable_`` prefix.
+* Make the old ``PyUnstable_*`` name be an alias (e.g. a ``static inline``
+  function calling the new function).
+* Announce the change in What's New.
+
+The old name should continue to be available until an incompatible change is
+made.
```
ncoghlan, line 240: Such incompatible changes after a tier promotion should be rare though, since the unstable API entry is an alias for a *public* API at that point. Suggested wording tweak:

```
The previously unstable name should continue to be made available as long as the new public
name remains available. If the public name is ever subsequently deprecated, then its unstable
alias will be deprecated at the same time.

While the unstable name could theoretically be removed or deprecated immediately without
violating the expectations set for the unstable API tier, there's also no need to break working
code just because a previously unstable API is now considered ready for a wider audience.
```

ncoghlan (approved):

developer-workflow/c-api.rst
CAM-Gerlach:
```suggestion
2. The “general” public C API, available when :cpy-file:`Include/Python.h` is included normally.
```

The above substantive change without the quote change

developer-workflow/c-api.rst
CAM-Gerlach: I brought it up here since it was this PR that introduced the regression on the ordering (it previously was ordered least to most portable, which IMO isn't perfect but much better than all over the place), and my initial suggestion was to just insert the category at the appropriate point in that order, i.e. above Public API.

However, as @encukou brought up the added description of the unstable API substantially refers to that of the Public API, and as you mentioned this change would also be best made as a separate PR given, as I'd also noticed, it would conflate the two distinct changes (with squash-merge in use). Ideally, that change would have been considered and implemented first to avoid introducing a regression here and only fixing it later, but pragmatically speaking at this point I guess we just go ahead with this and fix the order in an immediate followup.

developer-workflow/c-api.rst
CAM-Gerlach: Yeah, while is this PR that is introducing a regression (as the order currently _is_ least-to-most portable), practically speaking the approach you suggest is the most straightforward to implement at this point.

That being said, we should at least try to avoid introducing too much of an unnecessary regression here, particularly in this list where the regression is more serious but much easier to fix. Therefore, to help mitigate this while minimizing the further changes here, the simplest fix is to just add the unstable tier to its correct place in this tier list per the current least-to-most portable order (while keeping the document section order as it is now in this PR). This is an ordered list of API tiers, not sections on the page (that's what the ToC is for, and the page section order not being as important is the justification for the regression in the first place).

developer-workflow/c-api.rst
CAM-Gerlach, line 240: I have no subject matter expert opinion on the changes, of course (though ISTM that removing the alias after a backward-incompatible change to a public API should still be at least mentioned, as it is safer than silently changing behavior) but if you want to accept it, I've made @ncoghlan 's comment from the above into an actual suggestion, on the correct lines, and with some copyediting cleanup to make it more concise and avoid repetition.

```suggestion
The previous unstable name should remain available as long as the new public name is.
If the public name is ever subsequently deprecated, its unstable alias should be too.

While the unstable name could theoretically be removed or deprecated immediately without
violating the expectations set for the unstable API tier, there's also no need to break working
code just because a previously unstable API is now considered ready for a wider audience.
```

Or, with an explicit mention of removal after an incompatible change, as before:

```suggestion
The previous unstable name should remain available as long as the new public name is,
unless an incompatible change is made to the public API that could break users anyway.
If the public name is ever subsequently deprecated, its unstable alias should be too.

While the unstable name could theoretically be removed or deprecated immediately without
violating the expectations set for the unstable API tier, there's also no need to break working
code just because a previously unstable API is now considered ready for a wider audience,
unless it may break or change its behavior anyway due to another incompatible change.
```

CAM-Gerlach: I of course have no subject matter expert opinion, but I've [made this into an actual suggestion, on the correct lines](https://github.com/python/devguide/pull/1060#discussion_r1133040798) along with some additional copyediting and an alternate suggestion to still include the existing mention if desired.

ncoghlan: It's a public API, so incompatible changes are not allowed. This new text should *not* suggest that they are. If we want to make an incompatible change to a public API, we have to add a new API and deprecate the old one instead.

ferdnyc: @CAM-Gerlach It's basically the entirety of the ["Making incompatible changes" section of PEP 387](https://peps.python.org/pep-0387/#making-incompatible-changes). (Particularly steps 2 and 3.)

> ### Making incompatible changes
> Making an incompatible change is a gradual process performed over several releases:
> 
> 1. Discuss the change. Depending on the degree of incompatibility, this could be on the bug tracker, python-dev, python-list, or the appropriate SIG. A PEP or similar document may be written. Hopefully users of the affected API will pipe up to comment.
> 1. Add a warning. If behavior is changing, the API may gain a new function or method to perform the new behavior; old usage should raise the warning. If an API is being removed, simply warn whenever it is entered. `DeprecationWarning` is the usual warning category to use, but `PendingDeprecationWarning` may be used in special cases where the old and new versions of the API will coexist for many releases [[2]](https://peps.python.org/pep-0387/#warnings). Compiler warnings are also acceptable. The warning message should include the release the incompatibility is expected to become the default and a link to an issue that users can post feedback to.
> 1. Wait for the warning to appear in at least two minor Python versions of the same major version, or one minor version in an older major version (e.g. for a warning in Python 3.10, you either wait until at least Python 3.12 or Python 4.0 to make the change). It’s fine to wait more than two releases.
> [...]

CAM-Gerlach: Well, its certainly not "the entirety of the section"; it's the very specific point above I'm after. On careful inspection I do see what specifically implies this, though it is a bit buried and took multiple reads to find:

> If behavior is changing, the API may gain a new function or method to perform the new behavior; old usage should raise the warning. 

The language here really should be more clear and explicit, as the "may"s and "should"s used here certainly don't send nearly the same message as "incompatible changes are not allowed" and "If we want to make an incompatible change to a public API, we have to add a new API and deprecate the old one instead." Also, "old usage" is somehwt unclear terminology. For that, it should instead state something like

> If an incompatible behavior change to an existing API is needed, a new function or method MUST be added and the old construct deprecated, so existing uses don't silently break.

To note, that isn't always true at least for Python APIs; I can recall several instances (e.g. regex group identifiers) where incompatible changes were made to existing functions for various reasons.

encukou:
I will get back to this, but I'm catching up after COVID and the deadline for 3.12 *features* is coming closer, so it might be another few weeks :(
If anyone wants to take over this PR in the mean time, go ahead.

developer-workflow/c-api.rst
encukou: That sounds reasonable, thanks.

FWIW, the original order matches an ideal of how APIs are “promoted” as they mature: first implement the functionality, then make it usable for others, then make sure it's rock-solid.
In this list, the unstable API is on the same level as the general public API. And it's a niche enough topic to be documented with “same as *that* except for the following”.

developer-workflow/c-api.rst
encukou: That sounds like a good order for the user-facing docs. Developers who *implement* the API generally go in the opposite direction.

developer-workflow/c-api.rst
encukou: Actually, I think the alias should be removed as soon as the public API is deprecated. If you signed up for updating your code to react to implementation changes, you should update it when the API is no longer favoured.

```suggestion
The old name should remain available until the
new public name is deprecated or removed.
```

This is technically a change in the PEP, I'll need to run it by the SC.

encukou:
I'm back! Does this look good?

developer-workflow/c-api.rst
encukou: >  I'll need to run it by the SC.

(I plan to do this after merging, and revert if necessary)

lhmzhou (approved):

encukou:
Thank you for the reviews!
If there are no objections, I intend to merge **tomorrow** so the content is up and linkable, even if it's not presented optimally.
This will probably reset GitHub reviews, so please post what you have pending and leave the rest for follow-up discussion/PRs/issues (which are *very* welcome of course).

PR #1063, opened 2023-03-08
add carljm to experts index

carljm:
I talked to @ericvsmith and confirmed that he is OK with me joining him in the dataclasses experts list.

----
:books: Documentation preview :books:: https://cpython-devguide--1063.org.readthedocs.build/

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -85,7 +85,7 @@ codeop
 collections           rhettinger*
 collections.abc       rhettinger*, stutzbach^
 colorsys
-compileall
+compileall            carljm
 concurrent.futures    pitrou, brianquinlan, gpshead*
 configparser          ambv*
 contextlib            ncoghlan, 1st1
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
Just to make sure you saw—you'll need to add `*` after your name if you actually want to be tagged on issues/PRs that concern any of these topics.

carljm:
> Just to make sure you saw—you'll need to add * after your name if you actually want to be tagged on issues/PRs that concern any of these topics.

According to the wording above the table in that page, `*` specifically means "OK to assign to." Not sure what you mean by "tagged," but I take that to mean a simple cc by `@name` highlight.

I'd rather not have issues/PRs assigned to me by others, since I take assignment to mean "actively working on this, nobody else should work on it without asking me first." But I'm happy to be tagged/cced on issues in my areas of interest. According to my reading, leaving off the `*` should lead to what I want :)

PR #1076, opened 2023-04-15
Rename expert labels to topic labels

hugovk:
For https://github.com/python/core-workflow/issues/450.

To merge once labels are renamed for https://github.com/python/core-workflow/issues/450#issuecomment-1509743345.

New name and anchor:

* https://cpython-devguide--1076.org.readthedocs.build/triage/labels/#topic-labels

Keep the old anchor pointing to the same bit:

* https://cpython-devguide--1076.org.readthedocs.build/triage/labels/#expert-labels

----
:books: Documentation preview :books:: https://cpython-devguide--1076.org.readthedocs.build

### DIFF
--- a/triage/labels.rst
+++ b/triage/labels.rst
@@ -64,6 +64,7 @@ and :gh-label:`OS-freebsd`.
 
 
 .. _Expert labels:
+.. _Topic labels:
 
 Topic labels
 ============
--- a/triage/labels.rst
+++ b/triage/labels.rst
@@ -68,7 +68,7 @@ and :gh-label:`OS-freebsd`.
 Topic labels
 ============
 
-These labels are used to specify the area of expertise required to address
+These labels are used to denote the specific topic area, if any, of
 the issue/PR.  This includes both specific modules/packages and generic
 interest areas.
 
--- a/triage/labels.rst
+++ b/triage/labels.rst
@@ -54,9 +54,10 @@ These labels are mostly used to specify which :ref:`part of the codebase
   (written in Python) and other changes related to tests, :mod:`unittest`,
   or :mod:`doctest`.
 
+.. _Expert labels:
 
-Expert labels
-=============
+Topic labels
+============
 
 These labels are used to specify the area of expertise required to address
 the issue/PR.  This includes both specific modules/packages and generic
@@ -66,8 +67,8 @@ Adding these labels is also a way to notify the relevant experts, since
 they are encouraged to subscribe to them.  Depending on the label,
 this might also automatically add the issue to a GitHub project.
 
-You can see the `full list of expert labels on GitHub
-<https://github.com/python/cpython/labels?q=expert>`_.
+You can see the `full list of topic labels on GitHub
+<https://github.com/python/cpython/labels?q=topic>`_.
 
 
 OS labels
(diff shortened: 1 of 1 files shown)


triage/labels.rst
```diff
@@ -54,9 +54,19 @@ These labels are mostly used to specify which :ref:`part of the codebase
...
+the only available labels are :gh-label:`OS-windows`, :gh-label:`OS-mac`,
+and :gh-label:`OS-freebsd`.
+
+
+.. _Expert labels:
+
+Topic labels
+============
 
 These labels are used to specify the area of expertise required to address
```
CAM-Gerlach, line 71: Maybe worth updating this a bit too, to reflect the title change and avoid implying that being an expert is required to interact with the PR?

```suggestion
These labels are used to denote the specific topic area, if any, of
```

CAM-Gerlach (commented):
One comment of a corresponding change to consider to the label description, otherwise LGTM from me, thanks

CAM-Gerlach (approved):
LGTM from me, FWIW, thanks

triage/labels.rst
```diff
@@ -54,29 +54,30 @@ These labels are mostly used to specify which :ref:`part of the codebase
...
+These labels are used to specify which operating systems are affected.
+Since most issues either affect all systems or are specific to Unix,
+the only available labels are :gh-label:`OS-windows`, :gh-label:`OS-mac`,
+and :gh-label:`OS-freebsd`.
+
+
+.. _Expert labels:
+
+Topic labels
+============
```
merwok, line 69: Is this heading going to get the automatic `Topic labels` reST target and `topic-labels` HTML ID and an extra target `Expert labels` for compatibility, or is the explicit markup going to override the automatic things?

hugovk: We get these HTML anchor IDs:

* https://cpython-devguide--1076.org.readthedocs.build/triage/labels/#expert-labels
* https://cpython-devguide--1076.org.readthedocs.build/triage/labels/#topic-labels

And we get just the reST target to to `Expert labels`.

The main reason for adding `.. _Expert labels:` is to make sure the old HTML anchor still works.

But shall we also add a `Topic labels` reST target, like this?

```diff
diff --git a/triage/labels.rst b/triage/labels.rst

 .. _Expert labels:
+.. _Topic labels:
 
 Topic labels
 ============
```

ezio-melotti (approved):

triage/labels.rst
merwok: Ideally the reST target would be `Topic labels`.  It that requires explicit markup without losing the other ID, then let’s do that.  (If it doesn’t work try switching the two lines.)

hugovk: Updated, so now we have HTML anchors and reST targets for both!

pradyunsg (approved):

triage/labels.rst
CAM-Gerlach: The ref target label, if needed, should be the standard `.. _topic-labels:` , as used for all other modern ref target labels in the devguide. While there's no immediate benefit to adding such if the target isn't used, it makes it easier for any new inbounds to add such, so I generally favor doing so.

PR #1086, opened 2023-04-26
Add willingc as asyncio expert and update motivation entry

willingc:
----
:books: Documentation preview :books:: https://cpython-devguide--1086.org.readthedocs.build/


JelleZijlstra (approved):
great!

PR #1087, opened 2023-04-29
Don't suggest to start build with infinite number of threads

Delgan:
Hi.

Just a small enhancement proposal, as I don't think calling `make` with unbounded number of threads is desirable.

Using `make -j` will start "infinite" number of jobs according to `make` docs, which slows down the build and consumes too many resources.

Suggesting `make -j4` seems to be a sane default, advanced users will know how to adjust it.

----
:books: Documentation preview :books:: https://cpython-devguide--1087.org.readthedocs.build/

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -29,7 +29,7 @@ instructions please see the :ref:`setup guide <setup>`.
 
 3. Build Python, on UNIX and macOS use::
 
-      ./configure --with-pydebug && make -j
+      ./configure --with-pydebug && make -j4
 
    and on Windows use:
 


hugovk:
Thanks for the suggestion.

What machine are you using? Does the build take longer with `make -j` than `make -j4`, or otherwise slow the machine down? "consumes too many resources" - in what way?

On macOS with 8 CPUs, `make -j` uses them all and I don't notice any other slowdown. Having said that, it takes about the same time to run as  `make -j4`, around 32s.

Delgan:
Hi @hugovk.

It's not so much related to my machine in particular. It's just that once all the CPU cores are used for compiling, adding more jobs generally do not improve the performance and rather tend to make it worst because of the overhead of context switches [^1]. It's not very much noticeable in practice, though [^2][^3].

However, what is growing continuously is the RAM consumption. Each started job will consume memory, which causes a noticeable increase in overall RAM usage (finally "wasted" since an infinite number of jobs cannot run in parallel). This is not really desirable, and this is frequently related to out-of-memory problems for a beginner who is very familiar with `make` building process [^4].

Also I suggested `-j4` but maybe `-j2` is preferable as it matches the example given when clicking "[more detailed instructions](https://devguide.python.org/getting-started/setup-building/#compiling)" on the page.

[^1]: [Disadvantages to high make job values](https://stackoverflow.com/a/17349734/2291710)
[^2]: [How to speed up compilation time in linux](https://stackoverflow.com/a/17749621/2291710) 
[^3]: [GNU make: should the number of jobs equal the number of CPU cores in a system?](https://stackoverflow.com/a/18853310/2291710)
[^4]: [Is `make -j` (with no argument) dangerous?](https://unix.stackexchange.com/q/316644)

terryjreedy:
For the test suite, -j starts a number of parallel tests appropriate to the CPU.  Since, you say, make  -j does not do that, why not recommend -j# where # is appropriate for the machine?

Delgan:
@terryjreedy On the same page, tests are suggested to be run with `./python -m test -j3`, details about the `-j` argument are only provided on [the linked page](https://devguide.python.org/testing/run-write-tests/#runtests). I suppose it's best to show a simple and functional example by default, and only go into detail on the appropriate page. It's easier to copy and paste. But in the end, it's up to what you prefer.

AA-Turner:
Superseeded by either #1541 or #1453 or both.

PR #1097, opened 2023-05-16
Replace deprecated utcnow() with now(UTC)

hugovk:
`datetime.datetime.utcnow` raises a deprecation warning in Python 3.12: https://github.com/python/cpython/issues/103857.

`datetime.UTC` is new in 3.11, and is an alias of `datetime.timezone.utc`, but I think we're fine requiring 3.11+ here.

https://docs.python.org/3/library/datetime.html#datetime.UTC

----
:books: Documentation preview :books:: https://cpython-devguide--1097.org.readthedocs.build/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,8 +24,6 @@ The official home of this guide is https://devguide.python.org.
 Compilation
 -----------
 
-For the compilation of the devguide, you need to use a version of Python which
-supports the ``venv`` module, because the ``make html`` command will create a
-virtual environment and will install the ``Sphinx`` package::
+For the compilation of the devguide, Python 3.11+ is needed::
 
     make html
--- a/_tools/generate_release_cycle.py
+++ b/_tools/generate_release_cycle.py
@@ -44,7 +44,7 @@ class Versions:
 
     def write_csv(self) -> None:
         """Output CSV files."""
-        now_str = str(dt.datetime.utcnow())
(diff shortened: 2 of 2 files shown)


Mariatta (approved):
Thanks

Mariatta:
Should we mention on the readme that things should be run using Python 3.11?

ezio-melotti (approved):

CAM-Gerlach:
I was taking a break at the time due to PyCon, sirrtm but I'm afraid I don't really understand the rationale for requiring the latest upstream Python version released only a few months ago to be able to render the devguide locally. Sure, many folks will have it one way or another, but it's far from a given as it targets contributors of all experience levels (especially newer ones) and those that make non-code contributions, and if they're using the `make` workflow they will likely need to figure out how to set the right version manually.

By contrast to how much this is likely to make life harder for many contributors, especially newer ones, it seems easy to avoid this minor issue (a mere warning) all in one line by, e.g., using a ternary conditional:

```python
now_str = str(dt.datetime.now(dt.UTC) if sys.version_info > (3, 10) else dt.datetime.utcnow())
```

Is there something I'm missing here?

hugovk:
Sure, please see PR https://github.com/python/devguide/pull/1107.

CAM-Gerlach:
Thanks @hugovk , as always!

PR #1098, opened 2023-05-20
Rename README.rst to README.md

CodeMaster7000:
----
:books: Documentation preview :books:: https://cpython-devguide--1098.org.readthedocs.build/


AlexWaygood:
This change wouldn't be correct, as the README is written in ReStructuredText, not MarkDown.

CAM-Gerlach:
It appears to be a spam user, with a similar strategy and nearly identical profile to @smashedfrenzy16 on #1105 . Their names follow a similar pattern, and even interact on many of the same repos. They have a similarly long history of spamming hundreds of similarly nonsense PRs, some even making the same or similar changes (e.g. chesslablab/.github#2 for this user, which was a day after chesslablab/.github#1 by @smashedfrenzy16 ).

However, in this case this user appears much more clearly to be a spammer, and may in fact just be a copycat of the previous, or controlled by them, and this PR is much more clearly invalid especially to a user of several years experience. As such it appears to be a much clearer candidate for an org ban by @python/organization-owners , and a report to GitHub.

CodeMaster7000:
@CAM-Gerlach I sincerely apologise for the useless PRs, but I am not a spam user and neither is SmashedFrenzy16. We are two separate people; in this case, I was just trying to tidy up small parts of the repository. I appreciate it is not a good thing, and I have since stopped.

SmashedFrenzy16:
@CAM-Gerlach, we are not spam bots. As you can see from my pofile, I have been gradually building up my projects. CodeMaster7000 joined later on, and had more projects since he had been coding from before he joined. So he had more projects at the launch of his GitHub. Furthermore, you might want to check our other social media accounts. For example, if you check my Twitter and both of our YouTube channels, you can evidently see no marks of spam. We are just two aspiring developers who want to change the world. As both of us are friends, we might tend to get a bit annoyed at each other and perhaps copy each other for various reasons (although this could also be a mistake that we copy each other). But we intend no harm, and are promptly fixing any repetitive behavior we have. For example, we don't copy each other in PR companies as much anymore. But sometimes we both need inspiration from each other, so we might have the same project ideas but with different code, or same company, but different PR changes. Small PR changes are for our learning experiences and to get better at making PRs. The frequency of these have decreased as we make more code related PRs. Furthermore, I don't think spam accounts go throught the hassle of setting up sponsors so tediously to help the community. Now I see you may regard this as a message that a spam bot made to support another, but I don't think this is the case as I have pure evidence to surround the fact that both of us are not bots. I understand your concern to stop GitHub being flooded with spam or other bad account types, but we simply are not bots. It would be more evident if we were. Also, more people would have picked up that we were spam bots if we were. But that does not present to be the case. I would like to end off with the fact that humans can make errors and that the changing to this README file's extension type might have been him not knowing the changes it would make. We both help each other to help learn different aspects of GitHub properly, and not everything can happen at worse.

ewdurbin:
hi. one does not need to a bot to produce spam, which distracts maintainers and wastes their time.

you do not need to make a case or defend yourself, consider taking time to ensure changes submitted are of substance, complete, explained well, and provided in good faith moving forward.

PR #1101, opened 2023-05-29
Update run-write-tests.rst

Jorgemucisalumfilho:
----
:books: Documentation preview :books:: https://cpython-devguide--1101.org.readthedocs.build/

### DIFF
--- a/testing/run-write-tests.rst
+++ b/testing/run-write-tests.rst
@@ -4,13 +4,13 @@
 =========================
 Running and Writing Tests
 =========================
-
-.. note::
-
-    This document assumes you are working from an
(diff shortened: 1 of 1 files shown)


Mariatta:
Thanks but at this time we will be keeping the devGuide in English.

CAM-Gerlach:
Checking the user history, it's a spambot that mass-spams PRs (sometimes dozens or hundreds per day, and multiple in the same repo) auto-translating bits of random text in popular repos, not just UI text but also syntax as well (which is obviously invalid). I'm reporting it, and suggest others do the same.

PR #1112, opened 2023-06-08
Bump macOS build recommendations to OpenSSL v3.0

### LINKED ISSUE #1111: Update macOS build recommendations to use OpenSSL 3.0
_Quoting @ned-deily in https://github.com/python/cpython/issues/99079#issuecomment-1581927371_:

> If we assume that that we should focus on 3.0.x support on macOS for the time being at least, I think the first step would be to update our GitHub Workflow macOS CI jobs for 3.11+ to use 3.0.x from Homebrew: currently they are using 1.1.x. We should also update the devguide recommendations for macOS to use 3.0.x; there is a [WIP issue for the devguide](https://github.com/python/devguide/pull/1058) that should be changed to use 3.0.x. Also, any macOS buildbots should be updated, as feasible. With that additional experience we can then move the macOS installers to using 3.0.x sometime in the not-too-distant future.

erlend-aasland:
Fixes #1111 

----
:books: Documentation preview :books:: https://cpython-devguide--1112.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -417,14 +417,14 @@ for the header and library files to your ``configure`` command.
 
 For example, with **Homebrew**, install the dependencies::
 
-    $ brew install pkg-config openssl@1.1 xz gdbm tcl-tk
+    $ brew install pkg-config openssl@3.0 xz gdbm tcl-tk
 
 Then, for Python 3.11 and newer, run ``configure``::
 
     $ GDBM_CFLAGS="-I$(brew --prefix gdbm)/include" \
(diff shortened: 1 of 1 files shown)


ned-deily (commented):
Thanks! I've pushed some changes to the PR. Rather than try to differentiate between OpenSSL 1.1.1 and 3.x, which would complicate things (like the brew installs) and since we want to get to 3.0 anyway, let's just bite the bullet here. I've also simplified the MacPorts recipes by removing the suggestion for 3.10 and earlier as that doesn't quite build correctly.

erlend-aasland:
Thanks!

erlend-aasland:
LGTM

PR #1113, opened 2023-06-10
Correct punctuation (add full stops)

safwansamsudeen:
As described in the title, merely a aesthetic change.

First contribution here, so lemme know if I've messed something up!

----
:books: Documentation preview :books:: https://cpython-devguide--1113.org.readthedocs.build/


AlexWaygood (approved):
Looks good, thanks!

PR #1129, opened 2023-06-26
Improve C API guidelines for return values

erlend-aasland:
Clarify that returning an error indicator should be done iff an
exception is raised.

For functions returning an int, a _negative value_ should be returned.
Suggest -1 unless there is a need to distinguish between error types.

----
:books: Documentation preview :books:: https://cpython-devguide--1129.org.readthedocs.build/

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -113,8 +113,8 @@ Guidelines for expanding/changing the public API
   and ``NULL`` if and only if an exception is raised.
   Other API must return ``0`` on success,
   and a negative value if and only if an exception is raised.
-  Unless there is need to distinguish between error types,
-  ``-1`` should be used as the error value.
+  ``-1`` should be used as the error value;
+  the caller should use the C idiom ``if (func() < 0)`` to check for errors.
 
 - APIs with lesser and greater results must return ``0`` for the lesser result,
   and ``1`` for the greater result.
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -110,9 +110,9 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` if and only if an exception is raised.
+  and ``NULL`` if, and only if, an exception is raised.
   Most other API must return ``0`` on success,
-  and a negative value if and only if an exception is raised.
+  and a negative value if, and only if, an exception is raised.
   Unless there is need to distinguish between error types,
   ``-1`` should be used as the error value.
 
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -110,9 +110,9 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` if, and only if, an exception is raised.
+  and ``NULL`` if and only if an exception is raised.
   Most other API must return ``0`` on success,
-  and a negative value if, and only if, an exception is raised.
+  and a negative value if and only if an exception is raised.
   Unless there is need to distinguish between error types,
   ``-1`` should be used as the error value.
 
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -111,7 +111,7 @@ Guidelines for expanding/changing the public API
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
   and ``NULL`` if and only if an exception is raised.
-  Most other API must return ``0`` on success,
+  Other API must return ``0`` on success,
   and a negative value if and only if an exception is raised.
   Unless there is need to distinguish between error types,
   ``-1`` should be used as the error value.
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` with an exception raised on error.
-  Most other API must return ``-1`` with an exception raised on error,
-  and ``0`` on success.
+  and ``NULL`` if and only if an exception is raised.
+  Most other API must return ``0`` on success,
+  and a negative value if and only if an exception is raised.
+  Unless there is need to distinguish between error types,
+  ``-1`` should be used as the error value.
 
 - APIs with lesser and greater results must return ``0`` for the lesser result,
   and ``1`` for the greater result.
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -111,7 +111,7 @@ Guidelines for expanding/changing the public API
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
   and ``NULL`` if and only if an exception is raised.
-  Other API must return ``0`` on success,
+  Other API must return an ``int``: return ``0`` on success,
   and a negative value if and only if an exception is raised.
   ``-1`` should be used as the error value;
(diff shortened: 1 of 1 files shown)


erlend-aasland:
@encukou, are you ok with this change?

vstinner (approved):
LGTM

developer-workflow/c-api.rst
```diff
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` with an exception raised on error.
-  Most other API must return ``-1`` with an exception raised on error,
-  and ``0`` on success.
+  and ``NULL`` if, and only if, an exception is raised.
```
AA-Turner, line 113: Minor syntax suggestion on commas
```suggestion
  and ``NULL`` if and only if an exception is raised.
```

developer-workflow/c-api.rst
```diff
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
...
+  Most other API must return ``0`` on success,
+  and a negative value if, and only if, an exception is raised.
```
AA-Turner, line 115:
```suggestion
  and a negative value if and only if an exception is raised.
```

erlend-aasland:
Thanks, Victor and Adam.

vstinner:
Sadly, the newly added PyLong_AsInt() doesn't respect new guidelines, but it was a deliberate choice. I like the new PyDict_GetItemRef() API, it's pleasant to use it.

developer-workflow/c-api.rst
```diff
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` with an exception raised on error.
-  Most other API must return ``-1`` with an exception raised on error,
-  and ``0`` on success.
+  and ``NULL`` if and only if an exception is raised.
+  Most other API must return ``0`` on success,
```
encukou, line 114: Which API is not in this “most” category?
```suggestion
  Other API must return ``0`` on success,
```

vstinner:
> Well, I've thought about this, and realized I haven't seen an actual use case for distinguishing between error types. Is there one?

It's not common to report different error cases, but it happens, especially if the API does not raise an exception.

On error, Py_DecodeLocale() sets size to ``(size_t)-1`` on memory allocation failure and ``(size_t)-2`` on decoding error.

_Py_DecodeLocaleEx() has 3 error cases:

* return -1: memory allocation failure
* return -2: decoding error
* return -3: unsupported error handler

For the caller, it's important to be able to distinguish the different errors. Extract of ``unicode_decode_locale()``:

```c
    int res = _Py_DecodeLocaleEx(str, &wstr, &wlen, &reason,
                                 current_locale, errors);
    if (res != 0) {
        if (res == -2) {
            PyObject *exc;
            exc = PyObject_CallFunction(PyExc_UnicodeDecodeError, "sy#nns",
                                        "locale", str, len,
                                        (Py_ssize_t)wlen,
                                        (Py_ssize_t)(wlen + 1),
                                        reason);
            if (exc != NULL) {
                PyCodec_StrictErrors(exc);
                Py_DECREF(exc);
            }
        }
        else if (res == -3) {
            PyErr_SetString(PyExc_ValueError, "unsupported error handler");
        }
        else {
            PyErr_NoMemory();
        }
        return NULL;
    }
```

developer-workflow/c-api.rst
ezio-melotti: I think it refers to the ones described below that return e.g. `0` for the lesser return or for "no item found", even though the wording might be a bit better.

erlend-aasland:
> Well, I've thought about this, and realized I haven't seen an actual use case for distinguishing between error types. Is there one?

I'd prefer if the guideline simply said "return -1 on error"; after all a guideline is meant for the _general_ case, not the _special_ case. For the extraordinary cases, it is ok to not follow the general guideline (after a discussion, of course).

developer-workflow/c-api.rst
```diff
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
...
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` with an exception raised on error.
-  Most other API must return ``-1`` with an exception raised on error,
-  and ``0`` on success.
+  and ``NULL`` if and only if an exception is raised.
+  Other API must return ``0`` on success,
+  and a negative value if and only if an exception is raised.
+  Unless there is need to distinguish between error types,
+  ``-1`` should be used as the error value.
```
erlend-aasland, line 117: @encukou, I suggest either:

```suggestion
  and a ``-1`` if and only if an exception is raised.
```

... or:

```suggestion
  and a negative value if and only if an exception is raised.
```

encukou:
> It's not common to report different error cases, but it happens, especially if the API does not raise an exception.

Well, if an exception isn't raised, then the function should not return a negative value (or NULL) at all.
That's the main consistency win we're going for: you should be able to tell whether an exception was raised from the return value alone.

---

There are two parts of the question of “Does a negative value other than -1 signal that an exception was raised?”

- what should functions return? (we can say `-1` only for now, it's easy to allow more values later)
- what should callers check for, `== -1` or `< 0`? (on this we can't really change our minds later)

encukou:
Yeah, `return -1` and `if (... < 0)` is probably the best solution. It means all the negative return values except -1 probably can't be repurposed for anything useful in the future, but that's probably OK.

But perhaps this shouldn't be decided in an issue, especially since there'll be lots of API discussions at the sprint in about a month.

erlend-aasland:
> But perhaps this shouldn't be decided in an issue, especially since there'll be lots of API discussions at the sprint in about a month.

Well, this PR has been up for two months already; there's no harm in waiting another month :)

vstinner:
Even when I'm 100% sure that a function can only return -1 on error, I prefer to write ``if (func() < 0)``. For me, it's a way to say: "if the function fails".

In the C library, it's unclear if a function returns 0 or 1 on success. It depends. But in the Python C API, it's likely that a function returns "a negative number" on error.

erlend-aasland:
> Even when I'm 100% sure that a function can only return -1 on error, I prefer to write `if (func() < 0)`. For me, it's a way to say: "if the function fails".

Yeah, that's the C idiom, so that would be my preferred style.

vstinner:
If a function can raise an exeeption, I don't see the point of having more than one negative value. The exception should contain the details: exception type and exception message.

In the worst case, if a very important additonal information cannot be passed in the exception, i would prefer to add a separated argument, than break the rule: return 1, 0 or -1.

vstinner:
My Py_DecodeLocale() is unrelated: it doesn't raise an exception, and so uses a different API. I'm talking about public C API which raise exceptions.

developer-workflow/c-api.rst
erlend-aasland:
```suggestion
  and a negative value if and only if an exception is raised.
  ``-1`` should be used as the error value;
  the caller should use the C idiom ``if (func() < 0)`` to check for errors.
```

vstinner: I prefer to be specific about -1 and suggest testing func() < 0.

developer-workflow/c-api.rst
```diff
@@ -110,9 +110,11 @@ Guidelines for expanding/changing the public API
   fields, arguments and return values are well defined.
 
 - Functions returning ``PyObject *`` must return a valid pointer on success,
-  and ``NULL`` with an exception raised on error.
-  Most other API must return ``-1`` with an exception raised on error,
-  and ``0`` on success.
+  and ``NULL`` if and only if an exception is raised.
+  Other API must return ``0`` on success,
```
vstinner, line 114:
```suggestion
  Other API must return an ``int``: return ``0`` on success,
```

erlend-aasland:
I'll leave this for the C API workgroup.

PR #1143, opened 2023-07-23
Reviewing PRs: On GitHub the nosy list is called subscribers

zormit:
When reading through this section, I stumbled over the term "nosy list", not knowing what it means.

It seems to be a [legacy concept](https://devguide.python.org/triage/github-bpo-faq/#where-is-the-nosy-list),
so this PR rephrases it to fit in with existing GitHub [terminology](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-issues#stay-up-to-date).


zormit:
I'm not sure if it's trivial enough to live without an issue. Happy to create one if necessary.

hugovk:
We can skip an issue for this :)

erlend-aasland (approved):
Thanks; makes sense

PR #1150, opened 2023-08-04
Add a new coordinator for zh_CN

Dumeng:
Add myself as a new translating coordinator for zh_CN

----
:books: Documentation preview :books:: https://cpython-devguide--1150.org.readthedocs.build/


ezio-melotti (approved):

ezio-melotti:
Is @zhsj still active on the translations?

Dumeng:
Actually, we have attempted different ways to contact @zhsj in the past months. Unfortunately, we cannot hear any back. So we are not sure about @zhsj 's status.

PR #1151, opened 2023-08-09
Migrate Argument Clinic docs from CPython to the devguide

### LINKED ISSUE #1148: Migrate Argument Clinic docs from the CPython repo to the devguide
See https://github.com/python/cpython/issues/107518

TODO:

- [ ] Where should `clinic.rst` live?
- [ ] Which section should the Argument Clinic docs live in? (Advanced Tools? CPython Internals?)

erlend-aasland:
Resolves #1148 

----
:books: Documentation preview :books:: https://cpython-devguide--1151.org.readthedocs.build/


ezio-melotti (commented):
Should this documented be added to the table in https://devguide.python.org/#contributing too?

erlend-aasland:
> Should this documented be added to the table in https://devguide.python.org/#contributing too?

That sounds like a good idea.

erlend-aasland:
> > Should this documented be added to the table in https://devguide.python.org/#contributing too?
> 
> That sounds like a good idea.

But, IMO, we should do that after the migration. I think it's better if this PR only focus on the act of migrating clinic.rst.

conf.py
```diff
@@ -110,6 +110,7 @@ def _asset_hash(path: os.PathLike[str]) -> str:
 rediraffe_redirects = {
     # Development Tools
     "clang.rst": "development-tools/clang.rst",
+    "clinic.rst": "development-tools/clinic.rst",
```
AA-Turner, line 113: These redirects were for when pages were at the top level -- it isn't really needeed for an entirely new page.

AA-Turner (changes requested):
I think the git history is not be quite right here (due to https://github.com/python/devguide/pull/1151/commits/d3b2f416e26b5e53012e636d2a6467551e49588d and https://github.com/python/devguide/pull/1151/commits/62ffd34a07b6c3b54b3c4878a8d8df6977c53f07) -- if you look in this branch's [blame](https://github.com/erlend-aasland/devguide/blame/migrate-clinic-docs/development-tools/clinic.rst) for ``clinic.rst``, it only goes back to the "Add clinic.rst with history", rather than any earlier commits.

I *believe* that if you squash the two commits I linked above, git will properly detect it as a file moved operation & history will work. (This worked locally for me, at least!)

A

conf.py
erlend-aasland: Ok, thanks!

erlend-aasland:
> I think the git history is not be quite right here (due to [d3b2f41](https://github.com/python/devguide/commit/d3b2f416e26b5e53012e636d2a6467551e49588d) and [62ffd34](https://github.com/python/devguide/commit/62ffd34a07b6c3b54b3c4878a8d8df6977c53f07)) -- if you look in this branch's [blame](https://github.com/erlend-aasland/devguide/blame/migrate-clinic-docs/development-tools/clinic.rst) for `clinic.rst`, it only goes back to the "Add clinic.rst with history", rather than any earlier commits.
> 
> I _believe_ that if you squash the two commits I linked above, git will properly detect it as a file moved operation & history will work. (This worked locally for me, at least!)

Thanks for the heads-up. I'll look into it.

erlend-aasland:
Closing in favour of #1160

PR #1158, opened 2023-09-01
Clarify that PyUnstable macros use "PyUnstable" prefix

vstinner:
Discussion: https://github.com/python/cpython/pull/108440#issuecomment-1699105964

----
:books: Documentation preview :books:: https://cpython-devguide--1158.org.readthedocs.build/

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -217,6 +217,8 @@ Moving an API from the public tier to Unstable
 ----------------------------------------------
 
 * Expose the API under its new name, with the ``PyUnstable_`` prefix.
+  The ``PyUnstable_`` prefix must be used for all symbols (functions, macros,
+  variables, etc.).
 * Make the old name an alias (e.g. a ``static inline`` function calling the
   new function).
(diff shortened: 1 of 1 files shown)


vstinner:
My previous abandoned PR for the Python documentation (``Doc/c-api/stable.rst``): https://github.com/python/cpython/pull/108710

cc @encukou

willingc (approved):
This looks like a reasonable clarification and guidance to me.

willingc:
@vstinner I'm going to merge. If @encukou has a different viewpoint, we can always go back and modify. Thanks Victor.

vstinner:
You're welcome Carol.

encukou:
Sorry, I dropped the ball here. Thank you Carol for picking it up!

A technicality: I wouldn't call macros *symbols* -- it's better to reserve that term to the stuff that's exported from a `.so`/DLL. 

But with the list of examples, it's fine.

vstinner:
> A technicality: I wouldn't call macros symbols -- it's better to reserve that term to the stuff that's exported from a .so/DLL.

PR #1162, opened 2023-09-13
Use sentence case for titles

hugovk:
The [devguide's style guide](https://devguide.python.org/documentation/style-guide/#capitalization) says:

> In the Python documentation, the use of sentence case in section titles is preferable, but consistency within a unit is more important than following this rule. If you add a section to a chapter where most sections are in title case, you can either convert all titles to sentence case or use the dominant style in the new section title.

The devguide has a wide range within units, many pages freely mixing both styles. You can see at https://devguide.python.org/#full-table-of-contents

 I think the devguide, and at the very least its style guide, should set an example with sentence case.

----
:books: Documentation preview :books:: https://cpython-devguide--1162.org.readthedocs.build/

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -213,7 +213,7 @@ use this API reliably:
   we shouldn't break it *unnecessarily*.
 
 
-Moving an API from the public tier to unstable
+Moving an API from the public tier to Unstable
 ----------------------------------------------
 
 * Expose the API under its new name, with the ``PyUnstable_`` prefix.
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -58,7 +58,7 @@ When in doubt, new internal C functions should be defined in
 ``Include/internal`` using the ``extern`` keyword.
 
 Private names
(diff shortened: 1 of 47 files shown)


triage/labels.rst
```diff
@@ -2,7 +2,7 @@
 .. _gh-labels:
 
 =============
-GitHub Labels
+Github labels
```
AA-Turner, line 5:
```suggestion
GitHub labels
```

AA-Turner (approved):
Thanks!

A

developer-workflow/c-api.rst
```diff
@@ -213,7 +213,7 @@ use this API reliably:
   we shouldn't break it *unnecessarily*.
 
 
-Moving an API from the public tier to Unstable
+Moving an API from the public tier to unstable
```
ezio-melotti, line 216: I think this should be reverted, since "Unstable" is capitalized in the text below too.
```suggestion
Moving an API from the public tier to Unstable
```

ezio-melotti (approved):

PR #1167, opened 2023-09-26
Update ``clinic.rst`` from CPython

AA-Turner:
cc: @erlend-aasland; there were two commits that I've only just noticed weren't included in the earlier branch. Please may you rebase-merge this PR to include them?

A

----
:books: Documentation preview :books:: https://cpython-devguide--1167.org.readthedocs.build/

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -152,7 +152,7 @@ process a single source file, like this:
 The CLI supports the following options:
 
 .. program:: ./Tools/clinic/clinic.py [-h] [-f] [-o OUTPUT] [-v] \
-             [--converters] [--make] [--srcdir SRCDIR] [FILE ...]
+             [--converters] [--make] [--srcdir SRCDIR] [--limited] [FILE ...]
 
 .. option:: -h, --help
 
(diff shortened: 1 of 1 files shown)


AA-Turner:
We'll need to wait a few hours until the Python documentation is rebuilt (hence why the workflow failed when re-run).

A

hugovk:
I'd like to move the _English HTML_ docs build from the docs server and use Read the Docs, so we get ~instant deploys instead of waiting ~24 hours. The non-HTML and translations can remain on the docs server (at least to begin with).

But that will need more planning (adding the language/version switchers, perhaps co-ordination with RtD) so something for future discussion :)

AA-Turner:
> waiting ~24 hours

Well, of course ``E[wait time]`` is only 12 hours..!

I have hesitations about moving to RtD, but again best part of a dedicated discussion. I think we could (and should) look into rebuilding ``/3/`` (perhaps also the pre-release versions, but certainly only the English HTML) more frequently than 24 hours, potentially every 4 or 6. I'm led to believe the expensive bits of the build are not the HTML bits. I'll open a docsbuild-scripts issue.

A

AA-Turner:
Opened https://github.com/python/docsbuild-scripts/issues/163

hugovk:
Thanks for opening that! 

---

Today's daily build has built today.

I've restarted the GHA job and it's passed. I don't see an option to rebuild RTD, but feel free to close/re-open if you like.

erlend-aasland (approved):

AA-Turner:
@erlend-aasland would you mind rebase-merging this please? I wouldn't want to progress the CPython PR without having the developer's guide being properly up-to-date.

A

erlend-aasland:
> @erlend-aasland would you mind rebase-merging this please? I wouldn't want to progress the CPython PR without having the developer's guide being properly up-to-date.

Thanks for taking care of this!

Issue #1168, opened 2023-09-26
Require to backport all documentation and tests changes to stable Python branches


vstinner:
All documentation and tests changes done in the Python main branch should be backported to stable branches (currently 3.11 and 3.12). It should be documented in the devguide.

Context: https://github.com/python/cpython/issues/108303#issuecomment-1729024106

hugovk:
This should probably go in https://devguide.python.org/developer-workflow/development-cycle/

(https://devguide.python.org/getting-started/git-boot-camp/#backporting-merged-changes says how to backport, not why/when.)

JacobCoffee:
Hi, I'd like to grab this if its untaken

AA-Turner:
Go ahead, feel free to ping in the PR for feedback.

A

JacobCoffee:
Also does this go along with #1069?

AA-Turner:
I'd probably add a note below that list saying that documentation and tests may be backported to any bugfix branch. See also [Jelle's notes](https://jellezijlstra.github.io/cpython#backporting).

A

JacobCoffee:
Hi @AA-Turner please see my changes - are they to little or just bad in general? 😅

hugovk:
Fixed in https://github.com/python/devguide/pull/1169, thanks @JacobCoffee!

vstinner:
Thanks.

Issue #1069, opened 2023-03-21
Clarify where docs fixes can be backported


encukou:
The [status key](https://devguide.python.org/versions/#status-key) for branches has lists of changes that can go in, and the lists look exclusive:

> feature:
>  -  new features, bugfixes, and security fixes are accepted.
>
> prerelease:
>  -  feature fixes, bugfixes, and security fixes are accepted for the upcoming feature release.
>
> bugfix:
>  -  bugfixes and security fixes are accepted, new binaries are still released. (Also called maintenance mode or stable release)
>
>security:
>  -  only security fixes are accepted and no more binaries are released, but new source-only versions can be released
>
>end-of-life:
>  -  release cycle is frozen; no further changes can be pushed to it.

This makes it a bit awkward to backport docs changes, especially now that the lists are included in the tag descriptions.

AFAIK, docs fixes don't need RM approval in feature, prerelease and bugfix branches. (Security branches are locked so any change needs the RM in the loop.) 
Is that right, @pablogsal @Yhg1s ?

encukou:
Looks like the detailed general policy should be described in [Development cycle](https://devguide.python.org/developer-workflow/development-cycle/#branches), and the status page should be a quick overview of the current state of the branches.

CAM-Gerlach:
Ah, okay. In that case, we could just reword the first three here to say `... bugfixes, docs changes and security fixes ...` and the security one to say `only security fixes (and security-related docs fixes) are...` and a few sentences or a short paragraph to the Development cycle page under the [Maintenance Branches hdaing](https://devguide.python.org/developer-workflow/development-cycle/#maintenance-branches) describing the current docs policy as detailed above. does that work?

encukou:
That sounds good!

hugovk:
@JelleZijlstra has written some notes at https://jellezijlstra.github.io/cpython#backporting

Related: https://github.com/python/devguide/issues/503 asking about backporting in general.

hugovk:
The related https://github.com/python/devguide/pull/1169 has been merged.

For this issue, do we still need some changes?

PR #1169, opened 2023-09-26
Note when to backport changes

JacobCoffee:
- Adds guidelines on what release cycles (where) should get backports of documentation and testing improvements. 

## Closes

- #1168
- Maybe #1069(?)

<details>
<summary>Changes visualized:</summary>
<a href="https://cpython-devguide--1169.org.readthedocs.build/developer-workflow/development-cycle/#maintenance-branches">Dev. Cycle</a>

<a href="https://cpython-devguide--1169.org.readthedocs.build/versions/#status-key">Versions</a>

</details>

----
:books: Documentation preview :books:: https://cpython-devguide--1169.org.readthedocs.build/

### DIFF
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -78,7 +78,7 @@ releases; the terms are used interchangeably. These releases have a
 
 The only changes allowed to occur in a maintenance branch without debate are
 bug fixes, test improvements, and edits to the documentation.
-Also, a general rule for maintenance branches is that compatibility 
+Also, a general rule for maintenance branches is that compatibility
 must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -83,7 +83,7 @@ must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
 
-Back-porting changes reduces the risk of future conflicts.
+Backporting changes reduces the risk of future conflicts.
 For documentation, it increases the visibility of improvements,
 since most readers access the `stable documentation <https://docs.python.org/3/>`__
 rather than the `development documentation <https://docs.python.org/dev/>`__.
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -83,15 +83,10 @@ must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
 
-Backporting serves dual purposes. First, it increases the visibility of
-documentation changes since most users refer to stable versions at
-`docs.python.org/3/ <https://docs.python.org/3/>`_ rather
-than the `docs.python.org/dev/ <https://docs.python.org/dev/>`_ documentation.
-Second, it minimizes conflicts for future bugfix backports.
-
-Backporting should target enhancements in documentation and tests for
-currently **stable** versions, specifically branches in a **bugfix** release cycle
-or higher.
+Back-porting changes reduces the risk of future conflicts.
+For documentation, it increases the visibility of improvements,
+since most readers access the `stable documentation <https://docs.python.org/3/>`__
+rather than the `development documentation <https://docs.python.org/dev/>`__.
 
 A new maintenance branch is normally created when the next feature release
 cycle reaches feature freeze, i.e. at its first beta pre-release.
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -77,10 +77,21 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes.  Also, a general rule for maintenance branches is that compatibility
+bug fixes. Also, a general rule for maintenance branches is that compatibility
 must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
-etc.).  For both rules, only rare exceptions are accepted and **must** be
-discussed first.
+etc.). For both rules, only rare exceptions are accepted and **must** be
+discussed first. Among the rare exceptions to the rules of bug fixes and compatibility,
+backporting of documentation and tests is encouraged.
+
+Backporting serves dual purposes. First, it increases the visibility of
+documentation changes since most users refer to stable versions at
+`docs.python.org/3/ <https://docs.python.org/3/>`_ rather
+than the `docs.python.org/dev/ <https://docs.python.org/dev/>`_ documentation.
+Second, it minimizes conflicts for future bugfix backports.
+
+Backporting should target enhancements in documentation and tests for
+currently **stable** versions, specifically branches in a **bugfix** release cycle
+or higher.
 
 A new maintenance branch is normally created when the next feature release
 cycle reaches feature freeze, i.e. at its first beta pre-release.
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -77,11 +77,10 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes. Also, a general rule for maintenance branches is that compatibility
-must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
-etc.). For both rules, only rare exceptions are accepted and **must** be
-discussed first. Among the rare exceptions to the rules of bug fixes and compatibility,
-backporting of documentation and tests is encouraged.
+bug fixes and the backporting of documentation and tests.  Also, a general rule
+for maintenance branches is that compatibility must not be broken at any point
+between sibling micro releases (3.5.1, 3.5.2, etc.). For both rules,
+only rare exceptions are accepted and **must** be discussed first.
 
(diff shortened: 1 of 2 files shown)


developer-workflow/development-cycle.rst
```diff
@@ -67,6 +67,10 @@ for release 3.n.
 Maintenance branches
 --------------------
 
+.. note:: Any documentation and test improvements should be backported to enhance currently
+    **stable** versions. This includes any branch currently in a release cycle of
+    **maint** (**bugfix**) and above.
```
vstinner, line 72: Backporting doc is to make doc changes more visible, since the majority of users visit docs.python.org/3/ or a specific version, not /dev/ doc. According to stats.

Bakcporting tests is to reduce conflicts when bugfixes are backported later.

vstinner: Would it be possible to write a regular paragraph, rather than a note?

versions.rst
```diff
@@ -41,6 +41,9 @@ Unsupported versions
 Status key
 ==========
 
+.. note:: Documentation and test changes may be backported to any branch in a release
+    cycle above **bugfix**.
```
vstinner, line 45: Can you add a link somehow to "Maintenance branches" section?

versions.rst
```diff
@@ -50,7 +53,7 @@ Status key
     but new source-only versions can be released
 :end-of-life: release cycle is frozen; no further changes can be pushed to it.
 
-See also the :ref:`devcycle` page for more information about branches.
+.. seealso:: Review the :ref:`devcycle` page for more information about branches.
```
AA-Turner, line 56: Let's not change this.

developer-workflow/development-cycle.rst
AA-Turner: I'd write this as regular prose, but also note that line 83 already contains text about backporting, so we should really make the change in that para, or nearby.

versions.rst
AA-Turner: Jacob -- I'd delete this note, and add a few words to the 'see also' below on backporting.

AA-Turner (changes requested):
Thanks for this Jacob. I agree with Victor's comments, this needs a few edits in the first instance.

A tip for using admonitions: they stand out and are highlighted, which means that using them must be done sparingly -- when is is most critical for a reader to be aware of a point? I think in this instance, backporting is not of sufficient import; so regular prose is fine. This is very akin to *the boy who cried wolf*, just in this instance replace the villagers with readers and the wolf with admonitions!

A

developer-workflow/development-cycle.rst
hugovk, line 72: "Maint" isn't used anywhere else, let's not introduce new terminology:
```suggestion
    **bugfix** and above.
```

JacobCoffee:
Thanks Adam, and understood. That makes sense - the use of admonitions is quite a bit different than I am used to 😄

JacobCoffee:
Please re-review at your earliest convenience to see if it is in line with what we want to communicate

Makefile
```diff
@@ -206,3 +206,7 @@ include/release-cycle.svg: include/release-cycle.json
 .PHONY: versions
 versions: venv include/branches.csv include/end-of-life.csv include/release-cycle.svg
 	@echo Release cycle data generated.
+
+.PHONY: docs-serve
+docs-serve:
+	sphinx-autobuild ./ ./_build -j auto --watch ./
```
AA-Turner, line 212: Let's not change unrelated things
```suggestion
```

JacobCoffee: oh oops I didnt mean to commit this

developer-workflow/development-cycle.rst
```diff
@@ -77,10 +77,21 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes.  Also, a general rule for maintenance branches is that compatibility
```
AA-Turner, line 80: Two spaces is valid (primarily emacs). We tend to avoid unrelated changes.

developer-workflow/development-cycle.rst
```diff
@@ -77,10 +77,21 @@ releases; the terms are used interchangeably. These releases have a
...
+bug fixes. Also, a general rule for maintenance branches is that compatibility
 must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
-etc.).  For both rules, only rare exceptions are accepted and **must** be
-discussed first.
+etc.). For both rules, only rare exceptions are accepted and **must** be
+discussed first. Among the rare exceptions to the rules of bug fixes and compatibility,
+backporting of documentation and tests is encouraged.
```
AA-Turner, line 84: The same point on double spaces.

I'd start the new sentence on a new line.

Edit: actually, please merge this sentence with L80. Docs/tests are not exceptions, but part of the standard set of permitted backports.

developer-workflow/development-cycle.rst
```diff
@@ -77,10 +77,21 @@ releases; the terms are used interchangeably. These releases have a
...
+
+Backporting serves dual purposes. First, it increases the visibility of
+documentation changes since most users refer to stable versions at
+`docs.python.org/3/ <https://docs.python.org/3/>`_ rather
+than the `docs.python.org/dev/ <https://docs.python.org/dev/>`_ documentation.
+Second, it minimizes conflicts for future bugfix backports.
+
+Backporting should target enhancements in documentation and tests for
+currently **stable** versions, specifically branches in a **bugfix** release cycle
+or higher.
```
AA-Turner, line 94: Personally I think this is too verbose and I would delete or cut significantly. But I'll wait for Victor or Hugo's thoughts before committing to a request.

versions.rst
AA-Turner: This was marked as resolved but the text remains?

versions.rst
```diff
@@ -50,7 +50,8 @@ Status key
     but new source-only versions can be released
 :end-of-life: release cycle is frozen; no further changes can be pushed to it.
 
-See also the :ref:`devcycle` page for more information about branches.
+See also :ref:`devcycle` for branch information and :ref:`maintbranch` for details on
+backporting documentation and test changes above the **bugfix** level.
 
 By default, the end-of-life is scheduled 5 years after the first release,
 but can be adjusted by the release manager of each branch.  All Python 2
```
AA-Turner, line 57: To be honest I'd probably revert this. Or: the much more minimal 'about branches and backporting'

AA-Turner:
Please avoid force-pushing, it makes reviewing in the GH interface much harder. PRs are squash merged.

A

developer-workflow/development-cycle.rst
JacobCoffee: Maybe just removing that line 86-90?

developer-workflow/development-cycle.rst
JacobCoffee: Maybe it seeming to be still around was part of my force push? Sorry 😦

developer-workflow/development-cycle.rst
```diff
@@ -77,10 +77,20 @@ releases; the terms are used interchangeably. These releases have a
...
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes.  Also, a general rule for maintenance branches is that compatibility
-must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
-etc.).  For both rules, only rare exceptions are accepted and **must** be
-discussed first.
+bug fixes and the backporting of documentation and tests.  Also, a general rule
+for maintenance branches is that compatibility must not be broken at any point
+between sibling micro releases (3.5.1, 3.5.2, etc.). For both rules,
+only rare exceptions are accepted and **must** be discussed first.
```
AA-Turner, line 83:
```suggestion
bug fixes, improved tests, and edits to the documentation.
Also, a general rule for maintenance branches is that compatibility 
must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
etc.).  For both rules, only rare exceptions are accepted and **must** be
discussed first.
```

developer-workflow/development-cycle.rst
AA-Turner:
```suggestion
Back-porting changes reduces the risk of future conflicts.
For documentation, it increases the visibility of improvements,
since most readers access the `stable documentation <https://docs.python.org/3/>`__
rather than the `development documentation <https://docs.python.org/dev/>`__.
```

developer-workflow/development-cycle.rst
```diff
@@ -77,11 +77,17 @@ releases; the terms are used interchangeably. These releases have a
...
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes.  Also, a general rule for maintenance branches is that compatibility
+bug fixes, test improvements, and edits to the documentation.
+Also, a general rule for maintenance branches is that compatibility 
 must not be broken at any point between sibling micro releases (3.5.1, 3.5.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
 
+Back-porting changes reduces the risk of future conflicts.
```
hugovk, line 86: Little nit: we use the closed, hyphenless version elsewhere, let's do the same here:

```suggestion
Backporting changes reduces the risk of future conflicts.
```

developer-workflow/development-cycle.rst
```diff
@@ -77,11 +77,17 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes.  Also, a general rule for maintenance branches is that compatibility
+bug fixes, test improvements, and edits to the documentation.
+Also, a general rule for maintenance branches is that compatibility
```
hugovk, line 81: For the linter:
```suggestion
Also, a general rule for maintenance branches is that compatibility
```

hugovk (approved):

AA-Turner (approved):
Thanks!

A

PR #1184, opened 2023-10-10
Fix wrongly indented nested list in `index.rst`.

ezio-melotti:
Follow-up of #1183 -- fixes a similar error in `index.rst`.

----
:books: Documentation preview :books:: https://cpython-devguide--1184.org.readthedocs.build/

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -180,14 +180,20 @@ Key resources
 -------------
 
 * Coding style guides
-    * :PEP:`7` (Style Guide for C Code)
-    * :PEP:`8` (Style Guide for Python Code)
+
+  * :PEP:`7` (Style Guide for C Code)
+  * :PEP:`8` (Style Guide for Python Code)
+
 * `Issue tracker`_
-    * :ref:`experts`
+
+  * :ref:`experts`
+
 * `Buildbot status`_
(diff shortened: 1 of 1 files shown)


AA-Turner (approved):

index.rst
```diff
@@ -180,14 +180,20 @@ Key resources
 -------------
 
 * Coding style guides
-    * :PEP:`7` (Style Guide for C Code)
-    * :PEP:`8` (Style Guide for Python Code)
+
+  * :PEP:`7` (Style Guide for C Code)
```
hugovk, line 184: Should these be three-space indents?

ezio-melotti: Lists work a bit differently (and that's the reason why people often get them wrong).  See
https://devguide.python.org/documentation/markup/#lists-and-quotes

We could clarify these exceptions in the devguide (possibly in both places).

hugovk (approved):

index.rst
pradyunsg: For posterity: Using three-space indents in this specific spot would result in a `blockquote` (or `dl`, depending on the absence of a blank line) in this specific spot.

PR #1188, opened 2023-10-11
Updating CPython Internals

### LINKED ISSUE #1159: Update internals section of devguide
# Documentation

The internals section of the devguide https://devguide.python.org/internals/ could do with updating.

The [compiler page](https://devguide.python.org/internals/compiler/) could do with updating. Particularly the "Introducing New Bytecode" doesn't belong here, and the "Known Compiler-related Experiments" should just be deleted.

The internals also needs links to the up to date docs in https://github.com/python/cpython/tree/main/Objects

willingc:
Closes #1159 

- Updates the Compiler doc
- Moves content from Compiler doc to Interpreter doc

This is the third of three PRs about CPython Internals.

----
:books: Documentation preview :books:: https://cpython-devguide--1188.org.readthedocs.build/

### DIFF
--- a/internals/compiler.rst
+++ b/internals/compiler.rst
@@ -142,7 +142,7 @@ Memory management
 Before discussing the actual implementation of the compiler, a discussion of
 how memory is handled is in order.  To make memory management simple, an **arena**
 is used that pools memory in a single location for easy
-allocation and removal.  This benefits us in the removal of explicit memory
+allocation and removal.  This enables the removal of explicit memory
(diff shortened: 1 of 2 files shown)


pradyunsg (commented):
> Moves content from Compiler doc to Interpreter doc

Is there a specific reason that the "Introducing new bytecode" section was moved?

willingc:
@pradyunsg Thanks for the review. The request to move the bytecode stuff was Mark's from the original issue.

PR #1201, opened 2023-10-19
Fix PowerShell command

fipachu:
Quoting PowerShell: `The term 'python.bat' is not recognized as the name of a cmdlet, function, script file, or operable program.`

`.\python.bat` however works perfectly.


fipachu:
Ooops, wrong file.

Issue #1196, opened 2023-10-17
Use tabs to display  commands in macOS or Windows


Mariatta:
Sometimes we need to display command lines that are different in macos and windows.

That's a feature in furo that lets you display such commands depends on the different environments, called Tabs:
https://pradyunsg.me/furo/reference/tabs/

Using the tabs feature in furo will make the docs look cleaner.

* [x] https://devguide.python.org/#quick-reference
* [x] https://devguide.python.org/documentation/start-documenting/ https://github.com/python/devguide/pull/1224
* [x] https://devguide.python.org/documentation/devguide/ https://github.com/python/devguide/pull/1223
* [x] https://devguide.python.org/getting-started/git-boot-camp/#downloading-other-s-patches https://github.com/python/devguide/pull/1225
* [x] https://devguide.python.org/getting-started/setup-building/#install-dependencies https://github.com/python/devguide/pull/1226
* [x] https://devguide.python.org/testing/coverage/#install-coverage https://github.com/python/devguide/pull/1228
* [x] https://devguide.python.org/testing/run-write-tests/

Mariatta:
I guess this is a feature of a sphinx extension.
Regardless whether it's Furo or a Sphinx extension, I think we should make use of the feature, and I identified a few places where Tabs could be used (and there might be more)

- https://devguide.python.org/getting-started/setup-building/#install-dependencies 

There are different commands for the different environments ie Linux, mac, etc. 

- https://devguide.python.org/#quick-reference

The commands for macOS vs Windows can be contained in Tabs.

- https://devguide.python.org/testing/coverage/#install-coverage

It has instructions for windows, macOs, Unix.

- https://devguide.python.org/getting-started/git-boot-camp/#downloading-other-s-patches

There's a different command for windows vs macos/unix.

hugovk:
Good idea.

For some examples using tabs, see:

* https://pip.pypa.io/en/stable/installation/
* https://packaging.python.org/en/latest/tutorials/packaging-projects/
* https://pillow.readthedocs.io/en/stable/installation.html#basic-installation

Mariatta:
It was suggested on Discord that we use the sphonx-inline-tabs extenstion for this.

lancegoyke:
I was playing around with this today and have some notes to share.

## Todo

Another location where tabs could be used:

- https://devguide.python.org/testing/run-write-tests/#runtests

## Testing environments

[...]

PR #1205, opened 2023-10-25
Add inline tabs on Quick Reference page for commands on different systems

lancegoyke:
This is for work on #1196

This PR introduces a new dependency: [sphinx-inline-tabs](https://sphinx-inline-tabs.readthedocs.io/en/latest/)

There were two design decisions I am not set on:

1. Should we say "Unix" or "Linux"? I picked "Unix" because the [setup docs](https://devguide.python.org/getting-started/setup-building/#mac-python-exe) say "Unix".
2. Is the most useful order Unix, then macOS, then Windows. I would presume most developers are using macOS, but other examples I've seen start with "Linux" (e.g., [PyPI](https://pip.pypa.io/en/stable/installation/), [Pillow](https://pillow.readthedocs.io/en/stable/installation.html#basic-installation)), so I put "Unix" first.

----
:books: Documentation preview :books:: https://cpython-devguide--1205.org.readthedocs.build/

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -11,9 +11,9 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx_copybutton',
+    'sphinx_inline_tabs',
     'sphinxext.opengraph',
     'sphinxext.rediraffe',
-    'sphinx_inline_tabs',
 ]
 
 # The master toctree document.
--- a/conf.py
+++ b/conf.py
@@ -13,6 +13,7 @@ extensions = [
     'sphinx_copybutton',
     'sphinxext.opengraph',
     'sphinxext.rediraffe',
+    'sphinx_inline_tabs',
 ]
 
 # The master toctree document.
--- a/index.rst
+++ b/index.rst
@@ -27,7 +27,7 @@ instructions please see the :ref:`setup guide <setup>`.
       git clone https://github.com/<your_username>/cpython
       cd cpython
 
-3. Build Python, on Unix and macOS use:
+3. Build Python:
 
    .. tab:: Unix
 
--- a/index.rst
+++ b/index.rst
@@ -66,15 +66,15 @@ instructions please see the :ref:`setup guide <setup>`.
 
          ./python.exe -m test -j3
 
+      Note: :ref:`Most <mac-python.exe>` macOS systems use
+      :file:`./python.exe` in order to avoid filename conflicts with
+      the ``Python`` directory.
+
    .. tab:: Windows
 
       .. code-block:: dosbatch
 
          .\python.bat -m test -j3
-
-   Note: :ref:`Most <mac-python.exe>` macOS systems use
-   :file:`./python.exe` in order to avoid filename conflicts with
-   the ``Python`` directory.
 
 5. Create a new branch where your work for the issue will go, e.g.::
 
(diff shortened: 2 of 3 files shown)


conf.py
```diff
@@ -13,6 +13,7 @@
     'sphinx_copybutton',
     'sphinxext.opengraph',
     'sphinxext.rediraffe',
+    'sphinx_inline_tabs',
```
hugovk, line 16: Let's sort this list in alphabetical order.

index.rst
```diff
@@ -57,17 +84,25 @@ instructions please see the :ref:`setup guide <setup>`.
...
 
-6. Once you fixed the issue, run the tests, and the patchcheck.
+6. Once you fixed the issue, run the tests, and the patchcheck:
+
+   .. tab:: Unix
+
+      .. code-block:: shell
+
+         make patchcheck
+
```
hugovk, line 94: To make the linter happy:
```suggestion

```

hugovk (commented):
It's working well at https://cpython-devguide--1205.org.readthedocs.build/ :)

> 1. Should we say "Unix" or "Linux"? I picked "Unix" because the [setup docs](https://devguide.python.org/getting-started/setup-building/#mac-python-exe) say "Unix".

Let's keep the existing terminology for this PR, so "Unix".

> 2. Is the most useful order Unix, then macOS, then Windows. I would presume most developers are using macOS, but other examples I've seen start with "Linux" (e.g., [PyPI](https://pip.pypa.io/en/stable/installation/), [Pillow](https://pillow.readthedocs.io/en/stable/installation.html#basic-installation)), so I put "Unix" first.


Without knowing what operating systems our visitors use, I'd probably stick with the same order as given in the text: Unix, macOS, then Windows.

(For the main docs, it's something like 59% Windows, 22% macOS and 10% Linux, but the audience of this site could be different.)

Mariatta:
I looked at the preview and it looks great. Thanks for working on this.

index.rst
```diff
@@ -27,27 +27,54 @@ instructions please see the :ref:`setup guide <setup>`.
       git clone https://github.com/<your_username>/cpython
       cd cpython
 
-3. Build Python, on Unix and macOS use::
+3. Build Python, on Unix and macOS use:
```
Mariatta, line 30:
```suggestion
3. Build Python:
```

The tab now shows instructions for Unix, macOS, and Windows.

index.rst
```diff
@@ -27,27 +27,54 @@ instructions please see the :ref:`setup guide <setup>`.
...
+
+   .. tab:: Windows
+
+      .. code-block:: dosbatch
+
+         .\python.bat -m test -j3
+
+   Note: :ref:`Most <mac-python.exe>` macOS systems use
+   :file:`./python.exe` in order to avoid filename conflicts with
+   the ``Python`` directory.
```
hugovk, line 77: Shall we put this note inside the macOS tab?

lancegoyke: Yes!

index.rst
hugovk: This bit still needs doing.

lancegoyke:
@Mariatta and @hugovk -- thank you for your input! I have updated every item you both raised. 

I made one additional change: `requirements.txt` is now in alphabetical order as that seemed similar to the `extensions` order in `conf.py`.

index.rst
lancegoyke: Weird, I know I accepted that suggestion. Updating...

hugovk (approved):
Thank you!

pradyunsg:
Thanks @lancegoyke! ^.^

PR #1213, opened 2023-11-02
Update issue-tracker.rst - change Roundup link to canonical link

rouilj:
Use www.roundup-trcker.org rather than roundup.sourceforge.io.

----
:books: Documentation preview :books:: https://cpython-devguide--1213.org.readthedocs.build/


willingc (approved):
Thanks for the update @rouilj.

PR #1206, opened 2023-10-26
Add inline tabs on Running and Writing Tests page for commands on different systems (work on #1196)

lancegoyke:
Hey team!

I tried adding those tabs to another page. Let me know if anything needs to be changed.

PR #1216, opened 2023-11-09
Activate tabs based on browser's operating system

hugovk:
For https://github.com/python/devguide/issues/1196.

After https://github.com/python/devguide/pull/1205 and https://github.com/python/devguide/pull/1206, we now have Unix/macOS/Windows tabs for https://devguide.python.org/ and https://devguide.python.org/testing/run-write-tests/.

This PR adds some JavaScript to check the browser's operating system, and then activates the relevant tab.

I've tested:

* [x] macOS with Chrome, Safari, Firefox, Opera -> macOS tab
* [x] Android with Chrome and Samsung Internet -> Unix tab (default)

Would be good if some other people could test:

* [ ] Linux -> Unix tab
* [ ] Windows -> Windows tab
* [ ] iPhone/iPad -> Unix tab (default)
* [ ] Anything else? -> Unix tab (default)

### DIFF
--- a/testing/run-write-tests.rst
+++ b/testing/run-write-tests.rst
@@ -5,6 +5,14 @@
 Running and writing tests
 =========================
 
+.. raw:: html
+
+    <script>
+    document.addEventListener('DOMContentLoaded', function() {
+      activateTab(getOS());
+    });
+    </script>
+
 .. note::
 
     This document assumes you are working from an
--- a/_static/activate_tab.js
+++ b/_static/activate_tab.js
@@ -1,0 +1,42 @@
+// Based on https://stackoverflow.com/a/38241481/724176
+function getOS() {
+  const userAgent = window.navigator.userAgent,
+    platform =
+      window.navigator?.userAgentData?.platform || window.navigator.platform,
+    macosPlatforms = ["macOS", "Macintosh", "MacIntel", "MacPPC", "Mac68K"],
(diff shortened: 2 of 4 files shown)


testing/run-write-tests.rst
```diff
@@ -5,6 +5,14 @@
...
 =========================
 
+.. raw:: html
+
+    <script>
+    document.addEventListener('DOMContentLoaded', function() {
+      activateTab(getOS());
+    });
+    </script>
+
```
ezio-melotti, line 15: I wonder if it's better to include this in every page that uses tabs or just once at the top level:
* if we do it for each page, we need to duplicate the function call for each page, and remember to add this whenever OS-specific tabs are added to a page.
* if we do it at the top level, it will be executed for each page, even when not needed

Another option would be contributing this upstream, and add a flag to enable/disable it.  It should also possible to include this snippet of HTML in an rst `|substitution|` to avoid code duplication, even if it might be overkill.

FWIW I tested on Linux/Firefox and I get "Unix" tabs (my `window.navigator.platform` is `"Linux x86_64"`).

hugovk: Yeah, I wasn't sure about including for all or just the required pages. It probably doesn't matter much either way, it's only a small bit of JS. 

I thought about contributing upstream, but tabs can be about anything, not only operating systems. Although as you mention, a flag would help. Another complication is the OS->tab mappings. For example, we have "Unix" shown for Linux.

Thanks for testing Linux!

PR #1217, opened 2023-11-10
Add "how to" for the critical_section Argument Clinic directive.

colesbury:
See also https://github.com/python/cpython/issues/111903.

----
:books: Documentation preview :books:: https://cpython-devguide--1217.org.readthedocs.build/

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -1955,6 +1955,47 @@ The generated docstring ends up looking like this:
    Return a copy of the code object with new values for the specified fields.
 
 
+.. _clinic-howto-critical-sections:
+
+How to use critical sections with Argument Clinic
+-------------------------------------------------
+
+You can use the ``@critical_section`` directive to instruct Argument Clinic to
+wrap the call to the "impl" function in a Python critical section.  In
+``--disable-gil`` builds of CPython, the critical section will acquire the
+per-object lock of the first argument.  Python critical sections are no-ops in
+builds of CPython with the GIL.
+
+See :cpy-file:`Include/internal/pycore_critical_section.h` and :pep:`703` for
+more details about critical sections.
+
+Example from :cpy-file:`Modules/_io/bufferedio.c`::
+
+   /*[clinic input]
+   @critical_section
+   _io._Buffered.close
+   [clinic start generated code]*/
+
+The generated glue code looks like this:
+
+.. code-block:: c
+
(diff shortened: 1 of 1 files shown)


development-tools/clinic.rst
```diff
@@ -1955,6 +1955,47 @@ The generated docstring ends up looking like this:
...
+-------------------------------------------------
+
+You can use the ``@critical_section`` directive to instruct Argument Clinic to
+wrap the call to the "impl" function in a Python critical section.
+In ``--disable-gil`` builds of CPython,
+the critical section will acquire the per-object lock of the first argument.
+Python critical sections are no-ops in builds of CPython with the GIL.
+
+See :cpy-file:`Include/internal/pycore_critical_section.h` and :pep:`703` for
+more details about critical sections.
```
AlexWaygood, line 1970: I'd flesh this out a little bit more, for people who don't know PEP-703 by heart ;)

Also, it would be great if we could link directly to https://peps.python.org/pep-0703/#python-critical-sections rather than just https://peps.python.org/pep-0703 (can't remember off the top of my head if there's a fancy intersphinx way to do that, or if we'd have to hardcode the link)

```suggestion
In builds of CPython without the Global Interpreter Lock (GIL),
critical sections are required in order to achieve
thread safety without causing deadlocks between threads.
When a critical section is entered into, a per-object lock associated
with the first argument of the decorated function is acquired.
The lock is released when the critical section on exiting the critical section.

Python critical sections are no-ops in builds of CPython with the GIL.
See :cpy-file:`Include/internal/pycore_critical_section.h` and :pep:`703` for
more details about critical sections.
```

AlexWaygood: Putting my and @hugovk's suggestions together, we have:

```suggestion
In builds of CPython without the Global Interpreter Lock ("GIL"),
critical sections are required in order to achieve
thread safety without causing deadlocks between threads.
When a critical section is entered into, a per-object lock associated
with the first argument of the decorated function is acquired.
The lock is released on exiting the critical section.

Python critical sections are no-ops in builds of CPython with the GIL.
See :cpy-file:`Include/internal/pycore_critical_section.h`
and :pep:`PEP 703 <703#python-critical-sections>`
for more details about critical sections.
```

ezio-melotti (approved):

erlend-aasland:
Thanks, everyone!

PR #1219, opened 2023-11-14
Document 'make htmllive'

hugovk:
As suggested on [Discuss](https://discuss.python.org/t/make-htmllive-new-docs-makefile-target-for-easier-editing/38574/2?u=hugovk).

Use the same "Tip" infobox on the pages for CPython docs and for the devguide.

image

Note, I didn’t add `htmllive` to `make.bat` because I don’t have Windows to test it. Would a Windows user like to add the equivalent commands to `make.bat`, so we can remove the "(Unix only)"? I’d suggest here for the devguide first.

----
:books: Documentation preview :books:: https://cpython-devguide--1219.org.readthedocs.build/

### DIFF
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -44,11 +44,11 @@ in the checkout directory.  On Windows use:
 
 You will find the generated files in ``_build/html``.
 
-.. tip:: * Substitute ``htmlview`` for ``html`` to open the docs in a web browser
+.. tip:: * Replace ``html`` with ``htmlview`` to open the docs in a web browser
            once the build completes.
-         * Substitute ``htmllive`` (Unix only) for ``html`` to automatically
-           rebuild the docs, start a local server, and reload the page in your
-           browser after you save reST files.
+         * Replace ``html`` with ``htmllive`` to rebuild the docs,
+           start a local server, and automatically reload the page in your
+           browser when you make changes to reST files (Unix only).
 
 Note that ``make check`` runs automatically when you submit
 a :ref:`pull request <pullrequest>`.  You may wish to run ``make check``
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -42,9 +42,15 @@ in the checkout directory.  On Windows use:
 
    > .\make html
 
-You will find the generated files in ``_build/html`` or, if you use
-``make htmlview``, the docs will be opened in a browser once the build
-completes.  Note that ``make check`` runs automatically when you submit
+You will find the generated files in ``_build/html``.
+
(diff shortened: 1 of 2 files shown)


documentation/devguide.rst
```diff
@@ -42,9 +42,15 @@ in the checkout directory.  On Windows use:
...
-You will find the generated files in ``_build/html`` or, if you use
-``make htmlview``, the docs will be opened in a browser once the build
-completes.  Note that ``make check`` runs automatically when you submit
+You will find the generated files in ``_build/html``.
+
+.. tip:: * Substitute ``htmlview`` for ``html`` to open the docs in a web browser
+           once the build completes.
+         * Substitute ``htmllive`` (Unix only) for ``html`` to automatically
+           rebuild the docs, start a local server, and reload the page in your
+           browser after you save reST files.
```
ezio-melotti, line 51:
```suggestion
.. tip:: * Substitute ``html`` with ``htmlview`` to open the docs in a web browser
           once the build completes.
         * Substitute ``html`` (Unix only) with ``htmlview`` to automatically
           rebuild the docs, start a local server, and reload the page in your
           browser after you save reST files.
```
I find this less confusing.

Another option is to just say:
```suggestion
.. tip:: * ``make htmlview`` will open the docs in a web browser
           once the build completes.
         * ``make htmllive`` (Unix only) will rebuild the docs,
            start a local server, and automatically reload the page in your
           browser when you make changes to reST files.
```

hugovk: Yeah, I also found "Substitute X for Y" unclear. Maybe the original idea was that plain `make` doesn't work for Windows?.

How about your first option with "Replace X with Y", or your second one and we can add Unix/Windows tabs (https://github.com/python/devguide/issues/1196)?

ezio-melotti: How about this?
```suggestion
.. tip:: * Substitute ``html`` with ``htmlview`` to open the docs in a web browser
           once the build completes.
         * Substitute ``html`` with ``htmllive`` to rebuild the docs,
           start a local server, and automatically reload the page in your
           browser when you make changes to reST files (Unix only).
```
I rephrased the second tip to emphasize that the automatism is in the reloading of the page when changes are made.  The "when" could also be changed to "every time" or "whenever" to further emphasize that you can keep making changes and the page will keep reloading automatically.  The "Unix only" applies to the whole tip, so it could either go at the beginning (`On Unix, ``make htmllive``...`, but this breaks symmetry with the first tip), or at the end like I did.

hugovk: Thanks, I went with this, with "Replace" instead of "Substitute".

ezio-melotti (approved):

PR #1220, opened 2023-11-15
Add `-j auto` to the Makefile.

ezio-melotti:
This PR adds `-j auto` to the `Makefile`.

I found out that it was missing while working on https://github.com/sphinx-doc/sphinx/issues/10779, and after adding it the time required to build the devguide went down from 2.9s to 1.4s, so about twice as fast (this is on a 16-cores CPU).

----
:books: Documentation preview :books:: https://cpython-devguide--1220.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,19 +4,21 @@
 # You can set these variables from the command line.
 PYTHON        = python3
 VENVDIR       = ./venv
-BUILDDIR      = _build
-SPHINXOPTS    = -W --keep-going
 SPHINXBUILD   = $(VENVDIR)/bin/sphinx-build
+SPHINXOPTS    = -W --keep-going
+BUILDDIR      = _build
 BUILDER       = html
-SPHINXLINT    = $(VENVDIR)/bin/sphinx-lint
+JOBS          = auto
 PAPER         =
+SPHINXLINT    = $(VENVDIR)/bin/sphinx-lint
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -b $(BUILDER) \
                   -d $(BUILDDIR)/doctrees \
-                  $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) \
+                  $(PAPEROPT_$(PAPER)) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,19 +4,21 @@
...
 PYTHON        = python3
 VENVDIR       = ./venv
-BUILDDIR      = _build
-SPHINXOPTS    = -W --keep-going
 SPHINXBUILD   = $(VENVDIR)/bin/sphinx-build
+SPHINXOPTS    = -W --keep-going
+BUILDDIR      = _build
 BUILDER       = html
-SPHINXLINT    = $(VENVDIR)/bin/sphinx-lint
+JOBS          = auto
```
ezio-melotti, line 11: `JOBS          = auto` and the `-j $(JOBS)` below are the only actual additions.  While I was at it I tried to reorganize the vars a bit to make them closer to `cpython`'s `Doc/Makefile`, including the order of the vars below.

itamaro (approved):

Makefile
```diff
@@ -4,19 +4,21 @@
...
+SPHINXLINT    = $(VENVDIR)/bin/sphinx-lint
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -b $(BUILDER) \
                   -d $(BUILDDIR)/doctrees \
-                  $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) \
+                  $(PAPEROPT_$(PAPER)) \
+                  -j $(JOBS) $(SPHINXOPTS) \
```
hugovk, line 21: Shall we put each `-x SOME_OPTION` in its own line, and then the `PAPEROPT_xxx` and `SPHINXOPTS` lucky bags on their own lines too?

(`PAPEROPT_xxx` is a `-D` under the hood, but we can't see it here.)

```suggestion
                  -j $(JOBS) \
                  $(PAPEROPT_$(PAPER)) \
                  $(SPHINXOPTS) \
```

hugovk (approved):
Nice little speedup!

egeakman (approved):
Agree with Hugo's suggestion, LGTM

PR #1223, opened 2023-11-18
Add inline tabs on Helping with the Developer’s Guide page for commands on different systems

lancegoyke:
Hey again team! I was hoping to knock a few more of these out this weekend.

This PR should close a single TODO in #1196.

I did restructure the writing a bit to break up the directive "to build the devguide..." from the long paragraph before it.

----
:books: Documentation preview :books:: https://cpython-devguide--1223.org.readthedocs.build/documentation/devguide/

### DIFF
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -46,6 +46,8 @@ To build the devguide from the checkout directory:
 
 .. tab:: Unix/macOS
 
+   .. code-block:: shell
+
       make html
 
 .. tab:: Windows
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -44,15 +44,7 @@ you*.
 
 To build the devguide from the checkout directory:
 
-.. tab:: Unix
-
-   .. code-block:: shell
-
-      make html
-
-.. tab:: macOS
-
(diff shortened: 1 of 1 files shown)


hugovk:
Hello! Thanks again for these!

We've also since added some JavaScript to activate the relevant tab based on the browser's operating system in https://github.com/python/devguide/pull/1216.

Like in that PR, please could you add this snippet near the top of the new pages with tabs?

```rst
.. raw:: html

    <script>
    document.addEventListener('DOMContentLoaded', function() {
      activateTab(getOS());
    });
    </script>
```

lancegoyke:
Of course! Apologies, I had seen you mention that, but I missed the fact that it needed to be included on each page.

hugovk:
No apology needed! Yeah, we could have loaded the JS for all pages, but for now we went with explicitly adding it to the tab pages.

On reflection, it's such a small amount of JS it's probably fine to always load it, but we can think about that later.

Anyway, I think we could also combine the Unix and macOS tabs in this one into a single Unix/macOS tab?

lancegoyke:
Sounds good to me!

hugovk:
Thanks!

PR #1224, opened 2023-11-18
Add inline tabs on Documentation Getting Started page for commands on different systems

lancegoyke:

There is a little bit of discussion in the [Build using make / make.bat](https://devguide.python.org/documentation/start-documenting/#build-using-make-make-bat) section. My understanding is the commands are the same so I didn't try to make tabs for them all, but we can adjust if you believe that's better.

----
:books: Documentation preview :books:: https://cpython-devguide--1224.org.readthedocs.build/documentation/start-documenting/

### DIFF
--- a/documentation/start-documenting.rst
+++ b/documentation/start-documenting.rst
@@ -128,9 +128,19 @@ that attempts to emulate it as closely as practical.
    Make sure the environment you :ref:`created above <doc-create-venv-windows>`
    is `activated <venv-activate_>`__ before running ``make.bat``.
 
-To build the docs as HTML, run::
+To build the docs as HTML, run:
 
-   make html
+.. tab:: Unix/macOS
+
+   .. code-block:: shell
+
+      make html
+
+.. tab:: Windows
+
+   .. code-block:: dosbatch
+
+      .\make html
 
 .. tip:: * Replace ``html`` with ``htmlview`` to open the docs in a web browser
            once the build completes.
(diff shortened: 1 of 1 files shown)


hugovk:
On this page, we originally had:

> **On Unix platforms** that support **make** (including Linux, macOS and BSD)...

So we can lump all those under Unix, and just have two tabs: Unix and macOS.

The JavaScript tab selector (please add it like https://github.com/python/devguide/pull/1223#issuecomment-1817517424) will default to Unix, so it that's fine for macOS.

---



> There is a little bit of discussion in the [Build using make / make.bat](https://devguide.python.org/documentation/start-documenting/#build-using-make-make-bat) section. My understanding is the commands are the same so I didn't try to make tabs for them all, but we can adjust if you believe that's better.

Yeah, let's see how it goes with tabs here.

So for the Unix tab, just the Unix bit of this:

image

And for the Windows tab, the Windows bit of this, plus "Important" note:

image

---

We also have a few `make X` commands on this page. We can now tabify these too, so Windows users can copy/paste their `.\make X` versions.

Thanks!

documentation/start-documenting.rst
```diff
@@ -81,22 +89,33 @@ Create a virtual environment
...
 .. _doc-create-venv-unix:
+.. _doc-create-venv-windows:
 
-**On Unix platforms** that support :program:`make`
-(including Linux, macOS and BSD),
-you can create a new :mod:`venv` with the required dependencies using::
+You can create a new :mod:`venv` with the required dependencies using:
 
-   make venv
+.. tab:: Unix
```
ezio-melotti, line 96:
```suggestion
.. tab:: Unix/macOS
```

documentation/start-documenting.rst
```diff
@@ -81,22 +89,33 @@ Create a virtual environment
...
+   without you having to activate it.
+
+.. tab:: macOS
+
+   .. code-block:: shell
+
+      make venv
+
+   Building the docs with :program:`make` will automatically use this environment
+   without you having to activate it.
```
ezio-melotti, line 112:
```suggestion
```

ezio-melotti (commented):
ISTM that all the commands on this page that are valid for Unix are also valid for macOS, so we should be able to merge those two tabs.  Using `Unix/macOs` as tab name throughout the page should be fine as well, but I haven't tested it.

documentation/start-documenting.rst
lancegoyke: Sounds good to me!

lancegoyke:
Pushed some new commits with those changes.

Here's that intro on the makefile for Unix:

image

And for Windows:

image

hugovk (approved):
Thanks again!

PR #1228, opened 2023-11-18
Add inline tabs on Increase Test Coverage page for commands on different systems

lancegoyke:

----
:books: Documentation preview :books:: https://cpython-devguide--1228.org.readthedocs.build/testing/coverage/

### DIFF
--- a/testing/coverage.rst
+++ b/testing/coverage.rst
@@ -4,6 +4,14 @@
 Increase test coverage
 ======================
 
+.. raw:: html
+
+    <script>
+    document.addEventListener('DOMContentLoaded', function() {
+      activateTab(getOS());
+    });
+    </script>
+
 Python development follows a practice that all semantic changes and additions
 to the language and :abbr:`stdlib (standard library)` are accompanied by
 appropriate unit tests. Unfortunately Python was in existence for a long time
--- a/testing/coverage.rst
+++ b/testing/coverage.rst
@@ -87,23 +87,31 @@ just built, and this built version of Python will not see packages installed
 into your default version of Python. One option is to use a virtual environment
 to install coverage.
 
-On Unix run::
+.. tab:: Unix
 
-    ./python -m venv ../cpython-venv
-    source ../cpython-venv/bin/activate
-    pip install coverage
+    .. code-block:: shell
 
-On :ref:`most <mac-python.exe>` macOS systems run::
+        ./python -m venv ../cpython-venv
+        source ../cpython-venv/bin/activate
+        pip install coverage
 
-    ./python.exe -m venv ../cpython-venv
-    source ../cpython-venv/bin/activate
-    pip install coverage
+.. tab:: macOS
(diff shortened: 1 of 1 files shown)


testing/coverage.rst
```diff
@@ -87,23 +87,31 @@ just built, and this built version of Python will not see packages installed
 into your default version of Python. One option is to use a virtual environment
 to install coverage.
 
-On Unix run::
+.. tab:: Unix
 
-    ./python -m venv ../cpython-venv
-    source ../cpython-venv/bin/activate
-    pip install coverage
+    .. code-block:: shell
```
hugovk, line 92: Let's keep a simple prompt:
```suggestion
    Run:

    .. code-block:: shell
```

testing/coverage.rst
```diff
@@ -87,23 +87,31 @@ just built, and this built version of Python will not see packages installed
...
+
+        ./python.exe -m venv ../cpython-venv
+        source ../cpython-venv/bin/activate
+        pip install coverage
+
+.. tab:: Windows
+
+    .. code-block:: dosbatch
```
hugovk, line 110:
```suggestion
    Run:

    .. code-block:: dosbatch
```

hugovk (commented):
The rest of the page has `./python X` and `make X` commands that we tabify to help macOS/Windows users copy and paste the right one.

And let's also add the JS.

lancegoyke:
@hugovk 

> And let's also add the JS.

Thank you for the explicit repetition on this instruction.

> The rest of the page has `./python X` and `make X` commands that we tabify to help macOS/Windows users copy and paste the right one.

I just want to clarify something before adding all of the tabs. There's a line earlier in the doc after installing `coverage` that reads: "You can now use python without the ./ for the rest of these instructions, as long as your venv is activated."

Should we change all of these commands to `python` rather than their tabbed `./`, `.exe`, and `.bat` counterparts?

It feels like a cleaner change, but having not gone through this document myself, I'm unaware if it will add confusion.

Another possibility would be to prepend a `(cpython-venv)` to imply that the virtual env is activated, but that feels unnecessarily messy.

hugovk:
Yeah, I saw that "You can now..." line earlier and wasn't sure... I haven't actually used any of the instructions on this page, they might be outdated and need updating at some point, so let's not worry too much about them now and update them later, if needed.

So I suggest we leave those `./python X` as is, and only tabify the `make X` ones?

hugovk:
Hmm, now we have three "Unix" / "macOS" / "Windows" tabs at the top.

And further down two "Unix/macOS" / "Windows" tabs.

The auto-detection is fine on first load, and if for example you start with macOS, then select "Windows", they all flip over. Good so far.

But if you then select "macOS", it doesn't also flip to the "Unix/macOS" one. But I don't think it really matters, it's just another click.

Thanks!

hugovk (approved):

PR #1232, opened 2023-11-27
Add "how to" for the getter Argument Clinic directive.

corona10:

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate getter with Argument Clinic
+--------------------------------------------
+
+You can use ``@getter`` directive to generate "impl" function
+to defining getter.
+
+Example from :cpy-file:`Modules/_io/bufferedio.c`::
+
+    /*[clinic input]
+    @critical_section
+    @getter
+    _io._Buffered.closed
+    [clinic start generated code]*/
+
+The generate glue code looks like this.
+Note that this example is mixture usage with ``@critical_section`` directive
+to achieve thread safety without causing deadlocks between threads:
+
+.. code-block:: c
+
+    static PyObject *
+    _io__Buffered_closed_get(buffered *self, void *context)
+    {
+        PyObject *return_value = NULL;
+
+        Py_BEGIN_CRITICAL_SECTION(self);
+        return_value = _io__Buffered_closed_get_impl(self);
+        Py_END_CRITICAL_SECTION();
+
+        return return_value;
+    }
+
+.. versionadded:: 3.13
+
+
 .. _clinic-howto-deprecate-positional:
 .. _clinic-howto-deprecate-keyword:
 


hugovk:
@AlexWaygood GitHub UI fail on my part! I've fixed it locally and force pushed! Sorry about that!

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate getter with Argument Clinic
+--------------------------------------------
```
hugovk, line 2007: This whole page is about Argument Clinic, we can make the title shorter:
```suggestion
How to generate a getter
------------------------
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
...
+
+You can use ``@getter`` directive to generate "impl" function
+to defining getter.
```
hugovk, line 2010:
```suggestion
You can use the ``@getter`` directive to generate an "impl" function
to define a getter.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
...
+
+    /*[clinic input]
+    @critical_section
+    @getter
+    _io._Buffered.closed
+    [clinic start generated code]*/
+
+The generate glue code looks like this.
+Note that this example is mixture usage with ``@critical_section`` directive
+to achieve thread safety without causing deadlocks between threads:
```
hugovk, line 2022:
```suggestion
The generated glue code looks like this.
Note that this example is a mixture using the ``@critical_section`` directive
to achieve thread safety without causing deadlocks between threads:
```

hugovk (commented):
A few wording suggestions.

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate a getter
+------------------------
+
+You can use the ``@getter`` directive to generate an "impl" function
+to define a getter.
```
erlend-aasland, line 2010: It might be worth it to link to the C API docs here.

https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef.get

corona10: Done!

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate a getter
+------------------------
+
+You can use the ``@getter`` directive to generate an "impl" function
+to define a `getter <https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef.get>`_.
```
erlend-aasland, line 2010: Hm, when I follow the link, I get a feeling we shouldn't link to the `getter` specifically, but rather the whole `PyGetSetDef`. For someone unfamiliar with getters and setters, it might be confusing to arrive at the minimalistic docs of the `get` member :)

```suggestion
to define a `getter <https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef>`_.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
...
+
+    /*[clinic input]
+    @critical_section
+    @getter
+    _io._Buffered.closed
+    [clinic start generated code]*/
+
+The generated glue code looks like this.
+Note that this example is a mixture using the ``@critical_section`` directive
+to achieve thread safety without causing deadlocks between threads:
```
erlend-aasland, line 2022: To me this reads slightly weird; I expected the generated code snippet to follow straight after the first sentence! :) Also, to me the "note" sentence belongs _before_ the clinic input, not after it.

Moreover, is the `@critical_section` needed for the example? If not, I think we should remove it, so the example is stripped down to the bare minimum. OTOH, if all getters actually have a critical section (probably most of them do, @colesbury?), perhaps the `@getter` directive should imply `@critical_section`; it would reduce clinic input boiler plate code.

corona10: > Moreover, is the @critical_section needed for the example?

If we attach the real example, "yes"

> perhaps the @getter directive should imply @critical_section; it would reduce clinic input boiler plate code.

I think that it should be optional if possible.

erlend-aasland: Ok, so let's leave the `@critical_section` directive in the example for now. I still would like to mention that _before_ the clinic input, though. Also, we should link to the `@critical_section` how-to.

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,43 @@ The generated glue code looks like this:
...
+
+How to generate a getter
+------------------------
+
+You can use the ``@getter`` directive to generate an "impl" function
+to define a `getter <https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef>`_.
+
+Example from :cpy-file:`Modules/_io/bufferedio.c`::
+Note that this example is a mixture using the :ref:`@critical_section <clinic-howto-critical-sections>` directive
+to achieve thread safety without causing deadlocks between threads
```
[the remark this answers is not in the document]
corona10: > a simpler example should be chosen instead

Yeah, I think that we should replace it at the near future :)
See: https://github.com/python/devguide/pull/1232#discussion_r1408459396

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,44 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate a getter
+------------------------
+
+You can use the ``@getter`` directive to generate an "impl" function
+that defines a `getter <https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef>`_.
```
AlexWaygood, line 2010: I think this _should_ link correctly due to intersphinx:

```suggestion
that defines a :c:func:`getter <PyGetSetDef>`.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,44 @@ The generated glue code looks like this:
...
+
+This example --- taken from :cpy-file:`Modules/_io/bufferedio.c` ---
+shows the use of ``@getter`` in combination with
+the :ref:`@critical_section <clinic-howto-critical-sections>` directive
+(which achieves thread safety without causing deadlocks between threads)::
```
AlexWaygood, line 2015: You've shown me how to _use_ a `getter` here, but you haven't told me anywhere what a `getter` function actually _is_ or why I'd use it. I assume it's like a getter of a property, right? Could you go into that a little bit more? :)

You could maybe give a REPL example of how the generated C extension behaves at the Python level after it's been compiled?

corona10: Hmm, There might be some issue with markup

```
/home/runner/work/devguide/devguide/development-tools/clinic.rst:2009: WARNING: c:func reference target not found: PyGetSetDef
/home/runner/work/devguide/devguide/development-tools/clinic.rst:2039: WARNING: py:meth reference target not found: property.getter
```

development-tools/clinic.rst
AlexWaygood: Oh, my bad, I think it should be this instead:

```suggestion
that defines a :c:type:`getter <PyGetSetDef>`.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,54 @@ The generated glue code looks like this:
...
+
+.. code-block:: c
+
+    static PyObject *
+    _io__Buffered_closed_get(buffered *self, void *context)
+    {
+        PyObject *return_value = NULL;
+
+        Py_BEGIN_CRITICAL_SECTION(self);
+        return_value = _io__Buffered_closed_get_impl(self);
```
AlexWaygood, line 2033: There doesn't seem to be an `_io._Buffered` class at runtime AFAICS, and in the Python REPL example below you use `_io._BufferedIOBase`. Did you maybe mean to write this?

```suggestion
    _io._BufferedIOBase.closed
    [clinic start generated code]*/

The generated glue code looks like this:

.. code-block:: c

    static PyObject *
    _io__BufferedIOBase_closed_get(buffered *self, void *context)
    {
        PyObject *return_value = NULL;

        Py_BEGIN_CRITICAL_SECTION(self);
        return_value = _io__BufferedIOBase_closed_get_impl(self);
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,54 @@ The generated glue code looks like this:
...
+        Py_END_CRITICAL_SECTION();
+
+        return return_value;
+    }
+
+And then the implementation will work the same as the Python method which is
+decorated by `property <https://docs.python.org/3/library/functions.html#property>`_.
```
AlexWaygood, line 2040:
```suggestion
decorated by :py:func:`property`.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,54 @@ The generated glue code looks like this:
...
+
+.. code-block:: python
```
AlexWaygood, line 2042: REPL examples should use `pycon`, which stands for "Python console", and instructs Sphinx to apply syntax highlighting tailored for the Python REPL

```suggestion
.. code-block:: pycon
```

corona10: Oh not PyCon.. lol

AlexWaygood: I absolutely thought it was a typo the first time I saw `.. code-block:: pycon` 😆

development-tools/clinic.rst
corona10: Well, that's because..

https://github.com/python/cpython/blob/07ebd46f9e55ed2f18c5ea2a79ec5054bc26b915/Modules/_io/bufferedio.c#L2529

https://github.com/python/cpython/blob/07ebd46f9e55ed2f18c5ea2a79ec5054bc26b915/Modules/_io/bufferedio.c#L2589

https://github.com/python/cpython/blob/07ebd46f9e55ed2f18c5ea2a79ec5054bc26b915/Modules/_io/bufferedio.c#L2707

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,54 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate a getter
+------------------------
+
+You can use the ``@getter`` directive to generate an "impl" function
+that defines a :c:type:`getter <PyGetSetDef>`.
```
erlend-aasland, line 2010: IMO, we should explain getters first, and then introduce the directive.

```suggestion
"Getters" are C functions that facilitate property-like access for a type.
See :c:type:`getter <PyGetSetDef>` for details.
You can use the ``@getter`` directive to generate an "impl" function for a getter.
```

development-tools/clinic.rst
AlexWaygood: I see, thanks! Feel free to ignore me then (though if we can think of an example that _does_ have a one-to-one correlation with a Python object at runtime, that would be perfect! :)

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,55 @@ The generated glue code looks like this:
 .. versionadded:: 3.13
 
 
+.. _clinic-howto-getter:
+
+How to generate a getter
+------------------------
+
+"Getters" are C functions that facilitate property-like access for a type.
```
AlexWaygood, line 2009:
```suggestion
"Getters" are C functions that facilitate property-like access for a class.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,55 @@ The generated glue code looks like this:
...
+See :c:type:`getter <PyGetSetDef>` for details.
+You can use the ``@getter`` directive to generate an "impl" function for a getter.
```
AlexWaygood, line 2011:
```suggestion
You can use the ``@getter`` directive to generate an "impl" function for a
getter using Argument Clinic.
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,55 @@ The generated glue code looks like this:
...
+
+This example --- taken from :cpy-file:`Modules/_io/bufferedio.c` ---
```
AlexWaygood, line 2013:
```suggestion
This example -- taken from :cpy-file:`Modules/_io/bufferedio.c` --
```

development-tools/clinic.rst
```diff
@@ -2001,6 +2001,55 @@ The generated glue code looks like this:
...
+        PyObject *return_value = NULL;
+
+        Py_BEGIN_CRITICAL_SECTION(self);
+        return_value = _io__Buffered_closed_get_impl(self);
+        Py_END_CRITICAL_SECTION();
+
+        return return_value;
+    }
+
+And then the implementation will work the same as the Python method which is
```
AlexWaygood, line 2040:
```suggestion
And then the implementation will work the same as a Python method which is
```

erlend-aasland:
LGTM, given Alex's final remarks are addressed :) Thank you so much, Donghee!

erlend-aasland (approved):

corona10:
Thank you @AlexWaygood, @erlend-aasland and @ezio-melotti :)
Writing in English is quite difficult, but thanks to your feedback, I seem to be improving little by little.

PR #1235, opened 2023-12-01
Use emoji to illustrate good and bad example commit messages

hugovk:
Makes it clear which are the good and bad examples.

✅ Doesn't rely on colour: colour blind people can still see a tick or cross 
✅ We don't rely on them: the good/bad meaning is still clear from the text

<details>
<summary>Details</summary>

image

</details>

<details>
<summary>Details</summary>

image

</details>

Compare, for example, the Google developer docs: https://developers.google.com/style/ui-elements

Also fix some formatting.

----
:books: Documentation preview :books:: https://cpython-devguide--1235.org.readthedocs.build/

### DIFF
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -401,7 +401,7 @@ Solution:
 
 2. Apply the patch::
 
-       git apply /path/to/patch.diff
+       $ git apply /path/to/patch.diff
 
    If there are errors, update to a revision from when the patch was
    created and then try the ``git apply`` again::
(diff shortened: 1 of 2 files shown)


AlexWaygood (approved):

ezio-melotti:
PDF builds used to have issues with emojis, and even other Unicode characters. 
I don't know if that's still the case and we don't seem to build PDFs for the devguide, so in this case it should be ok.
This is something to keep in mind if you are planning to do similar changes to the `cpython` docs though.

PR #1236, opened 2023-12-04
Remove references to quay.io/tiran/cpython_autoconf

brettcannon:
The `make regen-configure` does the right thing via containers already.

----
:books: Documentation preview :books:: https://cpython-devguide--1236.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -513,27 +513,19 @@ After editing :file:`configure.ac`, run ``make regen-configure`` to generate
 When submitting a pull request with changes made to :file:`configure.ac`,
 make sure you also commit the changes in the generated files.
 
-The recommended and by far the easiest way to regenerate :file:`configure` is::
-
-   $ make regen-configure
-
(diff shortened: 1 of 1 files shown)


ezio-melotti (approved):

erlend-aasland (approved):

encukou:
My devguide experience with documenting only the `make` command is that it hurts Windows users. Is this already unusable on Windows? (I don't think you'd want to tweak `configure` there, but perhaps fix up someone else's commit?)

And in this case, the command would also be useful if you have yet another kind of container tooling.

erlend-aasland:
Windows users do not use `configure`; they use their own custom build script.

brettcannon:
> My devguide experience with documenting only the `make` command is that it hurts Windows users. Is this already unusable on Windows? (I don't think you'd want to tweak `configure` there, but perhaps fix up someone else's commit?)
> 
> And in this case, the command would also be useful if you have yet another kind of container tooling.

Windows developers can always use a GitHub Codespace if they needed a way to do this.

PR #1238, opened 2023-12-06
Rework the docs for NEWS/What's New entries

erlend-aasland:
- Add subsubsections instead of keeping all the text in a single large subsection
- Add some text about how to word NEWS entries
- Mention that test and strictly internal changes seldom require NEWS entries

(Some minor wording changes might have snuck in, but largely, I kept the existing text.)

----
:books: Documentation preview :books:: https://cpython-devguide--1238.org.readthedocs.build/core-developers/committing/#updating-news-and-what-s-new-in-python

PR #1239, opened 2023-12-06
Lint: fixup subsubsection markups

erlend-aasland:
Consistently use ^ iso. ' markups for subsubsections

----
:books: Documentation preview :books:: https://cpython-devguide--1239.org.readthedocs.build/

### DIFF
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -34,7 +34,7 @@ and *null* for final releases), and ``N`` is the release serial number.
 Some examples of release tags: ``v3.7.0a1``, ``v3.6.3``, ``v2.7.14rc1``.
 
 Branches
-''''''''
+^^^^^^^^
 
 There is a branch for each *feature version*, whether released or not (e.g.
 3.7, 3.8).
@@ -141,7 +141,7 @@ The latest release for each Python version can be found on the `download page
 .. _stages:
 
 Stages
-''''''
+^^^^^^
 
 Based on what stage the :ref:`in-development <indevbranch>` version of Python
 is in, the responsibilities of a core developer change in regards to commits
@@ -214,7 +214,7 @@ changes to the branch.  After the final release is published, the full
 
 
 Repository administration
-'''''''''''''''''''''''''
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The source code is currently hosted on `GitHub
 <https://github.com/python/cpython>`_ in the `Python organization <https://github.com/python/>`_.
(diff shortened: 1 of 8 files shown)


erlend-aasland:
Discovered over at #1238.

erlend-aasland:
> For this file, are we starting at `=`, then `-` and `^`?

For which file?

ezio-melotti:
> For which file?

This PR updates the headings of 8 files and, if the update follows the convention described above, then it can be approved and merged.  If the other files already follow this convention, then there is nothing else to do.
If they don't, we might as well update all other files while we are at it, either in this PR or, if it turns out to be a major endeavor, we can also do it in batches.

developer-workflow/development-cycle.rst
```diff
@@ -34,7 +34,7 @@ and *null* for final releases), and ``N`` is the release serial number.
 Some examples of release tags: ``v3.7.0a1``, ``v3.6.3``, ``v2.7.14rc1``.
 
 Branches
-''''''''
+^^^^^^^^
```
AlexWaygood, line 37: Should this use `-`? It seems to be the second level (a "subsection"), and according to https://devguide.python.org/documentation/markup/#sections, the second level should use `-`, while the _third_ level (such as lines 46 and 68 in this file) should use `^`

AlexWaygood (commented):
LGTM other than https://github.com/python/devguide/pull/1239/files#r1417581480

hugovk:
> > For this file, are we starting at `=`, then `-` and `^`?
> 
> For which file?

I obviously didn't scroll very far down the diff! 😅

Let's try again: 

For these files, are we starting at `=`, then `-` and `^`?

When would we use the overlines?

> - `#` with overline, for parts
> - `*` with overline, for chapters

If a page wants more than four levels? In that case, it would be a bit of a tricky refactor, and it's probably a good idea to avoid such deep nesting.

developer-workflow/development-cycle.rst
erlend-aasland: Oh my, yeah you are right. This should be `---`, and we'll have to adjust all the following sub-stuff.

erlend-aasland: Fixed in dc285783624e6abc26a7ee5391b37f5b03eccdd3

erlend-aasland:
> For these files, are we starting at `=`, then `-` and `^`?

Starting at `=` it appears.

AlexWaygood (approved):

erlend-aasland:
Thanks for the reviews! I'll land this post house chores 🏠🗑️

ezio-melotti (approved):

PR #1246, opened 2023-12-12
Add explanation about @critical_section target directive

corona10:
See: https://github.com/python/cpython/issues/112213

----
:books: Documentation preview :books:: https://cpython-devguide--1246.org.readthedocs.build/

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -1996,7 +1996,7 @@ The generated glue code looks like this:
       return return_value;
    }
 
-You can lock up to two additional objects
+You can lock one or two additional objects
 by supplying their C variable names as arguments
 to the ``@critical_section`` directive.
 This example from :cpy-file:`Modules/_weakref.c` takes
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -1996,12 +1996,11 @@ The generated glue code looks like this:
       return return_value;
    }
 
-If you want to specify a target object to be locked then you can use a target directive for
-``@critical_section``. Since the ``@critical_section`` locks the object which is located
-as the first argument implicitly, you will need this feature when the ``@critical_section``
-locks an unwanted target object. Note that explicit target declaration only supports up to 2 objects.
-
-Example from :cpy-file:`Modules/_weakref.c`::
+You can lock up to two additional objects
+by supplying their C variable names as arguments
+to the ``@critical_section`` directive.
+This example from :cpy-file:`Modules/_weakref.c` takes
+one additional argument (a C variable named ``object``)::
 
    /*[clinic input]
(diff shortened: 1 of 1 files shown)


development-tools/clinic.rst
```diff
@@ -1996,6 +1996,37 @@ The generated glue code looks like this:
       return return_value;
    }
 
+If you want to specify a target object to be locked then you can use a target directive for
+``@critical_section``. Since the ``@critical_section`` locks the object which is located
+as the first argument implicitly, you will need this feature when the ``@critical_section``
+locks an unwanted target object. Note that explicit target declaration only supports up to 2 objects.
+
+Example from :cpy-file:`Modules/_weakref.c`::
```
erlend-aasland, line 2004: I wonder if this text is overly complex. How about reducing it to something like this?

```suggestion
You can lock up to two additional objects
by supplying their C variable names as arguments
to the ``@critical_section`` directive.
This example from :cpy-file:`Modules/_weakref.c` takes
one additional argument (a C variable named ``object``)::
```

corona10: Good!

erlend-aasland (approved):
LGTM. Let's give the other reviewers a chance to look at this before we land it.

development-tools/clinic.rst
```diff
@@ -1996,6 +1996,36 @@ The generated glue code looks like this:
       return return_value;
    }
 
+You can lock up to two additional objects
```
ezio-melotti, line 1999:
```suggestion
You can lock one or two additional objects
```
Maybe this is clearer since I initially read it as "(lock up) to" instead of "lock (up to)".

development-tools/clinic.rst
```diff
@@ -1996,6 +1996,36 @@ The generated glue code looks like this:
...
+by supplying their C variable names as arguments
+to the ``@critical_section`` directive.
+This example from :cpy-file:`Modules/_weakref.c` takes
+one additional argument (a C variable named ``object``)::
+
+   /*[clinic input]
+   @critical_section object
+   _weakref.getweakrefcount -> Py_ssize_t
+
+   object: object
```
ezio-melotti, line 2009: Maybe it would be better to find an example that uses a different name, in order to avoid confusion between the variable name and its type.

development-tools/clinic.rst
erlend-aasland: Thanks, that's less confusing!

ezio-melotti (approved):

development-tools/clinic.rst
erlend-aasland: Well, it does not have to be a real example :) The Argument Clinic docs are full of code examples that are just made up.

PR #1248, opened 2023-12-16
Add explanation about docstring for PyGetSetDef

corona10:
----
📚 Documentation preview 📚: https://cpython-devguide--1248.org.readthedocs.build/

See: https://github.com/python/cpython/pull/113160

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -2087,6 +2087,8 @@ The generated glue code looks like this:
 
    Getters and setters must be declared as separate functions.
    The *value* parameter for a "setter" is added implicitly by Argument Clinic.
+   The docstring is also supported for :c:type:`PyGetSetDef` functions like other
+   but it must be declared at ``@getter``.
 
 And then the implementation will work the same as a Python method which is
 decorated by :py:class:`property`:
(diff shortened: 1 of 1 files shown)


development-tools/clinic.rst
```diff
@@ -2087,6 +2087,8 @@ The generated glue code looks like this:
 
    Getters and setters must be declared as separate functions.
    The *value* parameter for a "setter" is added implicitly by Argument Clinic.
+   The docstring is also supported for :c:type:`PyGetSetDef` functions like
+   other methods, but it must be declared at ``@getter``.
```
ezio-melotti, line 2091: Maybe this wording is better?
```suggestion
   It is possible to create a docstring for the property by adding it to
   the ``@getter``.
```

I also noticed that there doesn't seem to be any section specifically for docstrings, even though there are some mentions about them in the tutorial and in the other howtos.  It might be useful to link to another part of the docs that explains how to add docstrings, but in order to do that we either need to add a label or create a new howto that explains how to do it and link to that.

corona10:
@AlexWaygood @erlend-aasland 
Would you like to take a look at this PR too? (FYI Not that urgent :) )

AlexWaygood (approved):

erlend-aasland (approved):
Thanks!

PR #1250, opened 2023-12-17
Fix dead links to the tokenizer's source code

MahadMuhammad:
## Fixed `token.c` hyperlink in compiler design

- Currently, the latest build (at the time of raising this PR) has the **typo** in `token.c` file, which can be seen here [`cpython-devguide--1245.org.readthedocs.build/`](https://cpython-devguide--1245.org.readthedocs.build/)
- Updated the name to point to the correct file

----
📚 Documentation preview 📚: https://cpython-devguide--1250.org.readthedocs.build/

### DIFF
--- a/developer-workflow/grammar.rst
+++ b/developer-workflow/grammar.rst
@@ -42,7 +42,7 @@ Note: sometimes things mysteriously don't work.  Before giving up, try ``make cl
   Then run ``make regen-ast`` to regenerate
   :cpy-file:`Include/internal/pycore_ast.h` and :cpy-file:`Python/Python-ast.c`.
 
-* :cpy-file:`Parser/tokenizer.c` contains the tokenization code.
+* :cpy-file:`Parser/lexer/` & :cpy-file:`Parser/tokenizer/` contains the tokenization code.
   This is where you would add a new type of comment or string literal, for example.
 
 * :cpy-file:`Python/ast.c` will need changes to validate AST objects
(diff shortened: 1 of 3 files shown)


AlexWaygood (changes requested):
Thanks for the PR! It's not _quite_ correct currently, however. The tokenizer was refactored into two new directories, `Parser/lexer/` and `Parser/tokenizer/`, in https://github.com/python/cpython/pull/110684. The link here should point to one or both of those directories rather than `token.c`.

AlexWaygood:
@lysnikolaou, does this look okay to you? Should we be linking to both directories in these places, or just one of the two?

ezio-melotti (approved):

developer-workflow/grammar.rst
```diff
@@ -42,7 +42,8 @@ Note: sometimes things mysteriously don't work.  Before giving up, try ``make cl
   Then run ``make regen-ast`` to regenerate
   :cpy-file:`Include/internal/pycore_ast.h` and :cpy-file:`Python/Python-ast.c`.
 
-* :cpy-file:`Parser/tokenizer.c` contains the tokenization code.
+* :cpy-file:`Parser/lexer/` and :cpy-file:`Parser/tokenizer/` contain the
+  tokenization code.
```
lysnikolaou, line 46:
```suggestion
* :cpy-file:`Parser/lexer/` contain the tokenization code.
```

lysnikolaou (commented):
I would just change this one occurrence to point to the lexer, which contains the pure tokenization code. The following sentence makes more sense this way.

Thanks for the ping @AlexWaygood!

developer-workflow/grammar.rst
MahadMuhammad: Done. Thanks @lysnikolaou and @AlexWaygood for the review and suggestions :)

AlexWaygood (approved):
Thanks!

PR #1255, opened 2023-12-30
Document Git worktree

hugovk:
Fixes https://github.com/python/devguide/issues/105.

Based on the instructions in the first post there. I've been using this for a couple of months.

Also add the [.editorconfig from CPython](https://github.com/python/cpython/blob/main/.editorconfig) and add some handy pre-commit hooks.

----
📚 Documentation preview 📚: https://cpython-devguide--1255.org.readthedocs.build/getting-started/git-boot-camp/#git-worktree

### DIFF
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -669,3 +669,65 @@ Examples of useful commands:
 * Set the browser::
 
       $ gh config set browser <browser-path>
+
+
+Git worktree
+------------
+
+When working on several version branches at once, experienced Git users can
+take advantage of Git's worktree to check out several branches at once.
+
+Setting up Git worktree
+^^^^^^^^^^^^^^^^^^^^^^^
+
+With an existing CPython clone (see :ref:`clone-your-fork`), rename the
+``cpython`` directory to ``main`` and move it into a new ``cpython``
+directory, so we have a structure like:
+
+.. Generated with: tree -L 1 -d cpython
+
+.. code-block:: text
+
+   cpython
+   └── main (.git is here)
+
+Next, create worktrees for the other branches::
+
+   $ cd cpython/main
+   $ git worktree add -b 3.11 ../3.11 upstream/3.11
+   $ git worktree add -b 3.12 ../3.12 upstream/3.12
+
+This gives a structure like this, with the code for each branch checked out in
+its own directory:
+
+.. code-block:: text
+
+   cpython
+   ├── 3.11
+   ├── 3.12
+   └── main
+
+Using Git worktree
+^^^^^^^^^^^^^^^^^^
+
+List your worktrees, for example::
+
+   $ git worktree list
+   /Users/my-name/cpython/main  b3d24c40df [main]
+   /Users/my-name/cpython/3.11  da1736b06a [3.11]
+   /Users/my-name/cpython/3.12  cf29a2f25e [3.12]
+
+Change into a directory to work from that branch. For example::
+
+   $ cd ../3.12
+   $ git checkout my-3.12-bugfix-branch
+   ...
+   $ cd ../main
+   $ git checkout my-feature-branch
+   ...
+
+.. seealso::
+
+   * `Git Reference Manual <https://git-scm.com/docs/git-worktree>`_
+   * `"Experiment on your code freely with Git worktree"
+     <https://opensource.com/article/21/4/git-worktree>`_
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -677,7 +677,7 @@ Git worktree
 With Git worktrees, you can have multiple isolated working trees
 associated with a single repository (the `.git` directory).
 This allows you to work simultaneously on different version
-branches, eliminating the need for multiple independent clones 
+branches, eliminating the need for multiple independent clones
 that need to be maintained and updated separately.
 In addition, it reduces cloning overhead and saves disk space.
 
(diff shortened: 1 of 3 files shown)


getting-started/git-boot-camp.rst
```diff
@@ -669,3 +669,65 @@ Examples of useful commands:
...
+
+Change into a directory to work from that branch. For example::
+
+   $ cd ../3.12
+   $ git checkout my-3.12-bugfix-branch
+   ...
+   $ cd ../main
+   $ git checkout my-feature-branch
+   ...
+
```
merwok, line 728: Are more steps needed to configure pulls from upstream and pushed to origin (personal fork)?

getting-started/git-boot-camp.rst
```diff
@@ -669,3 +669,65 @@ Examples of useful commands:
...
+   /Users/my-name/cpython/3.11  da1736b06a [3.11]
+   /Users/my-name/cpython/3.12  cf29a2f25e [3.12]
+
+Change into a directory to work from that branch. For example::
+
+   $ cd ../3.12
+   $ git checkout my-3.12-bugfix-branch
+   ...
+   $ cd ../main
+   $ git checkout my-feature-branch
```
ezio-melotti, line 726: Maybe this example would be more realistic:
```suggestion
   $ cd ../3.12
   $ git switch -c my-3.12-bugfix-branch
   ...
   $ cd ../main
   $ git switch -c my-feature-branch
```

I also assume that at the end one will have to remember to switch back to `3.12` and not `main` (in the `/3.12` dir), right?  Do we need to show a more complete example like:
```suggestion
   $ cd ../3.12
   $ git switch -c my-3.12-bugfix-branch  # create new branch
   $ # make changes, test them, commit
   $ git push origin my-3.12-bugfix-branch
   $ # create PR
   $ git switch 3.12  # switch back to the 3.12 branch
 ```
?

Anything else that I'm missing for this workflow?

File: getting-started/git-boot-camp.rst [hunk 3]
```diff
@@ -669,3 +669,65 @@ Examples of useful commands:
...
+directory, so we have a structure like:
+
+.. Generated with: tree -L 1 -d cpython
+
+.. code-block:: text
+
+   cpython
+   └── main (.git is here)
+
+Next, create worktrees for the other branches::
```
itamaro, line 694: perhaps add a few words about the relationship between the different worktrees? (e.g. do they share remotes? git config? if you fetch in one is it visible in others? if you add/rename remotes in one is it visible in others?)

itamaro (commented):
the git worktree addition and the editorcofig addition seem unrelated - any reason to do both in the same PR?

hugovk:
> the git worktree addition and the editorcofig addition seem unrelated - any reason to do both in the same PR?

I added them to help whilst making this PR, and seemed easier to keep them together, but I can split it to another PR if you like?

getting-started/git-boot-camp.rst
hugovk: Yeah, I should have used `checkout -b` instead of `git checkout`. I use `checkout` because my fingers have learned it, but I guess `switch` is the better recommendation.

And yes, that looks good to switch back to `3.12` at the end. Let's go with your second suggestion. Thanks!

getting-started/git-boot-camp.rst
hugovk: I've committed Ezio's suggestion that shows a more concrete example of pushing and switching back, how does [`3b26368` (#1255)](https://github.com/python/devguide/pull/1255/commits/3b2636881b40ac42246a70fee975a6957b216a85) look?

getting-started/git-boot-camp.rst
hugovk: In general this stuff is shared. Yes, remotes are shared, and (by default) repo config. You can add/rename a remote from one and cd to another, and still see them all.

https://git-scm.com/docs/git-worktree details some things which are not shared, which is likely way too detailed to include here. There's a link to the manual in the "See also" at the end.

ezio-melotti: Indeed the first paragraph could do a better job at clarifying *why* I would want to use worktrees.  The ability to work on multiple branches in parallel without switching back and forth from the same clone/dir is one.  AFAIU, other advantages should be:
* there's only one copy of the repo (`.git/*`)
* you only need to set up one repo (e.g. create remotes)
* you only need to update one repo (i.e. pull changes)
* less disk space is needed

itamaro:
> I added them to help whilst making this PR, and seemed easier to keep them together, but I can split it to another PR if you like?

My personal preference is one-theme-per-commit, but this change is small enough that it's not critical, so up to you!

getting-started/git-boot-camp.rst
hugovk: This sounds good, would you like to suggest a paragraph?

ezio-melotti: Done!

getting-started/git-boot-camp.rst
```diff
@@ -669,3 +669,70 @@ Examples of useful commands:
...
       $ gh config set browser <browser-path>
+
+
+Git worktree
+------------
+
+With Git worktrees, you can have multiple isolated working trees
+associated with a single repository (the `.git` directory).
+This allows you to work simultaneously on different version
+branches, eliminating the need for multiple independent clones
```
itamaro, line 680: remove trailing space

```suggestion
branches, eliminating the need for multiple independent clones
```

itamaro (approved):
LG!

erlend-aasland (approved):
Nice; thanks, Hugo.

PR #1259, opened 2024-01-05
Document that type-feature labels do not need to be paired with version labels

erlend-aasland:
----
📚 Documentation preview 📚: https://cpython-devguide--1259.org.readthedocs.build/

### DIFF
--- a/triage/labels.rst
+++ b/triage/labels.rst
@@ -26,6 +26,8 @@ These labels are used to specify the type of issue:
 * :gh-label:`type-crash`: for hard crashes of the interpreter, possibly with a
   core dump.
 * :gh-label:`type-feature`: for feature requests or enhancements.
+  Feature requests do not need :ref:`Version labels`;
+  it is implicit that features are added to the ``main`` branch only.
   The `Ideas Discourse category`_ can be used to discuss enhancements
   before filing an issue.
 * :gh-label:`type-security`: for security issues.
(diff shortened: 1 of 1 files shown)


triage/labels.rst
```diff
@@ -26,6 +26,8 @@ These labels are used to specify the type of issue:
 * :gh-label:`type-crash`: for hard crashes of the interpreter, possibly with a
   core dump.
 * :gh-label:`type-feature`: for feature requests or enhancements.
+  Feature requests do not need :ref:`Version labels`;
```
hugovk, line 29: Avoid the capital V in the middle of a sentence:

```suggestion
  Feature requests do not need :ref:`version labels <Version labels>`;
```

erlend-aasland: Good catch! BTW; I found the capitalised anchor names weird. Is this a new thing? I can't remember seeing it in the docs before.

hugovk: Hmm, I'm not sure, let's ask @AA-Turner.

hugovk (approved):

triage/labels.rst
erlend-aasland: Well, if it is ok for Adam, I'd like to add the anchor in lower-case 🙂

erlend-aasland:
Thanks for the reviews!

PR #1260, opened 2024-01-10
Add governance contact

hugovk:
I wasn't sure which page to put this, this one seems as good as any because there's also info about GitHub org owners and admins.

Maybe we should move it up the page too?

image

---

Related: https://github.com/capi-workgroup/decisions/ sticks out -- would the C API Working Group like to move inside the https://github.com/python/ org? (re: https://devguide.python.org/developer-workflow/development-cycle/#organization-owner-policy)

----
📚 Documentation preview 📚: https://cpython-devguide--1260.org.readthedocs.build/developer-workflow/development-cycle/#governance

### DIFF
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -373,7 +373,7 @@ Multi-Factor Authentication must be enabled by the user in order to retain
 access as a Release Manager of the branch.
 
 Governance
-==========
+----------
 
 The Python Steering Council has overall authority over Python and has delegated
 some of its responsibilities to other groups.
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -371,3 +371,35 @@ administrators) to merge changes to that branch.
 
 Multi-Factor Authentication must be enabled by the user in order to retain
 access as a Release Manager of the branch.
+
+Governance
+==========
+
+The Python Steering Council has overall authority over Python and has delegated
+some of its responsibilities to other groups.
+
+This table lists the PEPs defining each group's responsibilities,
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

Mariatta:
Cc @python/editorial-board

developer-workflow/development-cycle.rst
```diff
@@ -371,3 +371,35 @@ administrators) to merge changes to that branch.
...
+     - :github:`python/steering-council`
+   * - C API Working Group
+     - :pep:`731`
+     - :github:`capi-workgroup/decisions`
+   * - Documentation Editorial Board
+     - :pep:`732`
+     - :github:`python/editorial-board`
+   * - Typing Council
+     - :pep:`729`
+     - :github:`python/typing-council`
```
gvanrossum, line 401: Just a silly nit to prove I received the notification:

How did you decide to order these? Maybe in PEP order (i.e., roughly chronologically :-) makes sense?

JelleZijlstra: I guess "Steering Council" is first because it's primary, and the rest is in roughly alphabetical order. Seems like a reasonable order to me. PEP order doesn't seem that useful as the PEPs were all so close together.

hugovk: Yes, I did consider PEP order but thought SC followed by alphabetical was a bit more useful.

willingc (approved):
Thanks @hugovk. Content looks great and a good addition. You may consider moving the content to the Communications channel page at the top. developer-workflow/communication-channels.rst 

My thinking is that someone wanting to contact folks in these groups may look there first.

PR #1261, opened 2024-01-10
Add big O notation to the style guide

hugovk:
Following https://github.com/python/cpython/pull/113695, document how to add big O notation.

image

----
📚 Documentation preview 📚: https://cpython-devguide--1261.org.readthedocs.build/documentation/style-guide/#big-o-notation

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -233,18 +233,12 @@ Big *O* notation is used to describe the performance of algorithms.
 
 Use italics for the big *O* and variables. For example:
 
-.. list-table::
-   :header-rows: 1
-
-   * - reStructuredText
-     - Rendered
-   * - ``*O*\ (1)``
-     - *O*\ (1)
-   * - ``*O*\ (log *n*)``
-     - *O*\ (log *n*)
-   * - ``*O*\ (*n*)``
-     - *O*\ (*n*)
-   * - ``*O*\ (*n* log *n*)``
-     - *O*\ (*n* log *n*)
-   * - ``*O*\ (*n*\ :sup:`2`)``
-     - *O*\ (*n*\ :sup:`2`)
+======================== ====================
+reStructuredText         Rendered
+======================== ====================
+``*O*\ (1)``             *O*\ (1)
+``*O*\ (log *n*)``       *O*\ (log *n*)
+``*O*\ (*n*)``           *O*\ (*n*)
+``*O*\ (*n* log *n*)``   *O*\ (*n* log *n*)
+``*O*\ (*n*\ :sup:`2`)`` *O*\ (*n*\ :sup:`2`)
+======================== ====================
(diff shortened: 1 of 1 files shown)


ezio-melotti (commented):
In this case a regular reST table (compact form) would probably be more readable then a list table (and save a few lines).

Somewhat unrelated, but I was wondering whether using (and recommending) `²` instead of `` :sup:`2` `` would be better, especially since the latter tends to become a regular `2` once copy-pasted.

hugovk:
> Somewhat unrelated, but I was wondering whether using (and recommending) `²` instead of `` :sup:`2` `` would be better, especially since the latter tends to become a regular `2` once copy-pasted.

Hmm, could be. We'd need to check epub/PDF/accessibility etc.

doerwalter:
This feels wrong to me: The "O" isn't an emphasized "O". This is a mathematical formula, so IMHO the `:math:` role should be used:

```
:math:`O(1)`
:math:`O(\log n)`
:math:`O(n)`
:math:`O(n \log n)`
:math:`O(n^2)`
```

This should work in both the PDF/TeX and the HTML output.

hugovk:
`:math:` renders as `O(n\log{n})` in plain text and `\(O\log{n}\)` in EPUB:

https://github.com/python/cpython/pull/113724#issuecomment-1878444909

ezio-melotti:
> Hmm, could be. We'd need to check epub/PDF/accessibility etc.

I did some research but I didn't find anything too conclusive.

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup#usage_notes) says:
> Appropriate use cases for `<sup>` include (but aren't necessarily limited to):
> * Displaying exponents, such as "x3". It may be worth considering the use of [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) for these, especially in more complex cases. See [Exponents](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup#exponents) under [Examples](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup#examples) below.

[Wikipedia](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts) says:
> The intended use when these characters were added to Unicode was to produce true superscripts and subscripts so that chemical and algebraic formulas could be written without markup.

and links to a [Unicode document](https://www.w3.org/TR/unicode-xml/#Superscripts).

Some screen readers seem to have options to handle `<sup>`.  Both `<sup>` and Unicode superscript chars seem valid choices, so unless someone finds something more conclusive, I'd say to just stick with `:sup:`/`<sup>`.

hugovk:
https://stackoverflow.com/a/12224471/724176 says to use the Unicode version where available instead of `<sup>`.

Similarly https://stackoverflow.com/a/76931597/724176 says ² is better for accessibility and mentions some screen readers that have problems with `<sup>`.

It also links to https://adrianroselli.com/2022/09/brief-note-on-super-and-subscript-text.html by an a11y expert, showing problems some screen readers have  with  `<sup>`.

erlend-aasland (approved):

hugovk:
Let's merge this, it reflects what we're using now. 

If we update the docs to Unicode ² instead of `sup` (and I'm not opposed, seem like an improvement), then we can update the devguide to match too.

PR #1263, opened 2024-01-24
Describe GC for free-threaded build in the GC design doc

colesbury:
This updates the garbage collector design page to describe the implementation of the GC for Python 3.13's free-threaded build.

----
📚 Documentation preview 📚: https://cpython-devguide--1263.org.readthedocs.build/

### DIFF
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -53,8 +53,23 @@ is needed to clean these reference cycles between objects once they become
 unreachable. This is the cyclic garbage collector, usually called just Garbage
 Collector (GC), even though reference counting is also a form of garbage collection.
 
+CPython contains two GC implementations starting in version 3.13. One implementation
+is used in the default build and relies on the global interpreter lock for
+thread-safety. The other implementation is used in the free-threaded build. Both
+implementations use the same basic algorithms, but operate on different data
+structures.
+
+
 Memory layout and object structure
 ==================================
+
+The garbage collector requires additional fields in Python objects to support
+garbage collection.  These extra fields are different in the default and the
+free-threaded builds.
+
+
+GC for the default build
+------------------------
 
 Normally the C structure supporting a regular Python object looks as follows:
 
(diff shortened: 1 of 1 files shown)


colesbury:
I think there's still some more I should add to the document, but I wanted to first get feedback on the general approach of mixing the descriptions of the two designs.

internals/garbage-collector.rst
```diff
@@ -53,9 +53,24 @@ is needed to clean these reference cycles between objects once they become
 unreachable. This is the cyclic garbage collector, usually called just Garbage
 Collector (GC), even though reference counting is also a form of garbage collection.
 
+CPython contains two GC implementations starting in version 3.13. One implementation
+is used in the default build and relies on the global interpreter lock for
+thread-safety. The other implementation is used in the free-threaded build. Both
+implementations use the same basic algorithms, but operate on different data
+structures.
```
ezio-melotti, line 60: This could use a list to make the difference between the two builds stand out more, and to mirror the two sections below.  Something like:
```suggestion
Starting in version 3.13, CPython contains two GC implementations starting in version 3.13:
* the default build implementation relies on the global interpreter lock for
thread-safety;
* the free-threaded build ...
 
Both implementations use the same basic algorithms, but operate on different data
structures.
```

Here you could also add link to `:term:`s like GC and GIL using intersphinx.

internals/garbage-collector.rst
```diff
@@ -107,6 +122,38 @@ isn't running at all!), and merging partitions, all with a small constant number
...
+garbage collection.  The field is used to identify objects that are tracked
+by the collector, ensure that finalizers are called only once per object,
+and, during garbage collection, differentiate reachable vs. unreachable objects.
+
+.. code-block:: none
+
+    object -----> +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+ \
+                  |                     ob_tid                    | |
+                  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+ |
+                  | pad | ob_mutex | ob_gc_bits |  ob_ref_local   | |
```
ezio-melotti, line 140: Do each `--+` mark one bit?  If so, `ob_tid`, `ob_ref_shared` and `*ob_type` are 2 bytes (16 bits), but this fields don't seem to match the description, since `ob_gc_bits` takes ~4 bits and `ob_ref_local` 6.

internals/garbage-collector.rst
```diff
@@ -107,6 +122,38 @@ isn't running at all!), and merging partitions, all with a small constant number
...
+                  |                  ob_ref_shared                | |
+                  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+ |
+                  |                    *ob_type                   | |
+                  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+ /
+                  |                      ...                      |
+
+
+The garbage collector also temporarily repurposes the ``ob_tid`` (thread ID)
+and ``ob_ref_local`` (local reference count) fields temporarily for other
+purposes during collections.
```
ezio-melotti, line 151:
```suggestion
The garbage collector also temporarily repurposes the ``ob_tid`` (thread ID)
and ``ob_ref_local`` (local reference count) fields for other purposes during
collections.
```

internals/garbage-collector.rst
```diff
@@ -171,10 +215,17 @@ is completely unreachable:
     >>> gc.collect()
     2
 
-When the GC starts, it has all the container objects it wants to scan
-on the first linked list. The objective is to move all the unreachable
-objects. Since most objects turn out to be reachable, it is much more
-efficient to move the unreachable as this involves fewer pointer updates.
+The GC starts with a set of candidate objects it wants to scan.  In the
+default build, theese "objects to scan" might be all container objects or a
```
ezio-melotti, line 219:
```suggestion
default build, these "objects to scan" might be all container objects or a
```

internals/garbage-collector.rst
colesbury: No -- each row is 64-bits (on 64-bit platforms), but otherwise the fields `pad`, `ob_mutex`, `ob_gc_bits`, and `ob_ref_local` are not to scale.

I'm not sure what to do about this. It'd be nice if the diagram were to scale, but I think that would require the diagram to be very wide -- `ob_gc_bits` is only 1 byte, but 10 characters.

ezio-melotti: It might be enough to just add a note after the diagram saying that e.g. "Note that `pad`, `ob_mutex`, `ob_gc_bits`, and `ob_ref_local` are not to scale and take X, Y, Z, W bytes respectively."

Another option, if those 4 values are related, could be to use a single name (like "free-threaded info bytes") and then break that down afterwards, either in text or with a separate "zoomed in" diagram.

internals/garbage-collector.rst
colesbury: I added a section at the end to summarize the differences between the two implementations. I was hesitant to add it here because we introduce a number of the concepts later on in the document.

ambv:
Let's get this in and iterate on the live document.

PR #1264, opened 2024-01-26
Clarify when to use versionadded and versionchanged

erlend-aasland:
----
📚 Documentation preview 📚: https://cpython-devguide--1264.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -1019,8 +1019,9 @@ units as well as normal text:
    feature, or a part of it, to the library or C API.  When this applies to an
    entire module, it should be placed at the top of the module section before
    any prose.
-   When a new attribute, function, or method is added, make sure a
-   ``versionadded`` directive is added.
+   When a new attribute, function, or method is added, a
+   ``versionadded`` directive should be present at the end of its
+   documentation block.
 
    The first argument must be given and is the version in question.  The second
    argument is optional and can be used to describe the details of the feature.
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -373,6 +373,8 @@ As you can see, the module-specific markup consists of two directives, the
    in overview files.
 
 
+.. _information-units:
+
 Information units
 -----------------
 
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -1019,9 +1019,9 @@ units as well as normal text:
    feature, or a part of it, to the library or C API.  When this applies to an
    entire module, it should be placed at the top of the module section before
    any prose.
-   When a new attribute, function, or method is added, a
-   ``versionadded`` directive should be present at the end of its
-   documentation block.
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Thank you!

erlend-aasland:
Thanks for the review; I'll wait for CAM to chime in before landing!

documentation/markup.rst
```diff
@@ -1019,13 +1019,19 @@ units as well as normal text:
    feature, or a part of it, to the library or C API.  When this applies to an
    entire module, it should be placed at the top of the module section before
    any prose.
+   When a new attribute, function, or method is added, make sure a
+   ``versionadded`` directive is added.
```
JelleZijlstra, line 1023:
```suggestion
   When a new attribute, function, or method is added, a
   ``versionadded`` directive should be present at the end of its
   documentation block.
```
Also describe explicitly where to put the directive.

erlend-aasland: That's an improvement! Thanks.

documentation/markup.rst
```diff
@@ -1019,13 +1019,20 @@ units as well as normal text:
    feature, or a part of it, to the library or C API.  When this applies to an
    entire module, it should be placed at the top of the module section before
    any prose.
+   When a new attribute, function, or method is added, a
+   ``versionadded`` directive should be present at the end of its
+   documentation block.
```
CAM-Gerlach, line 1024: **Note**: _Requires adding a new ref target label, `.. _information-units`, above the eponymous section above.

```suggestion
   When adding a new API :ref:`with a directive <information-units>`
   (``class``, ``attribute``, ``function``, ``method``, ``c:type``, etc),
   a ``versionadded`` should be included at the end of its description block.
```

* More broadly construe the types of entities we're discussing here (not just attributes, functions and methods)
* Be more specific about which get `versionadded` (any object/API that is represented by a directive, vs. e.g. callable parameters, behaviors, etc. that are not).
* Crosslink the relevant section
* Use active voice
* Describe the location a little more precisely/less ambiguously

documentation/markup.rst
```diff
@@ -1035,8 +1042,12 @@ units as well as normal text:
 
    Example::
 
-      .. versionchanged:: 3.1
-         The *spam* parameter was added.
+      .. function:: func(spam)
+
+         Return foo and bar with *spam* applied.
```
CAM-Gerlach, line 1047:
```suggestion
      .. function:: func(spam=False)

         Return foo and bar, optionally with *spam* applied.
```

Be a little more realistic and consistent with the previous (avoiding a compat break, which in this case would usually mean a new function).

CAM-Gerlach (commented):
Thanks! A few suggestions, one of which requires adding a new ref target label, `.. _information-units`, above the eponymous section.

CAM-Gerlach (approved):
LGTM, thanks @erlend-aasland !

PR #1271, opened 2024-02-06
Add Sam Gross & Nikita Sobolev, clarify how to announce

ambv:
----
📚 Documentation preview 📚: https://cpython-devguide--1271.org.readthedocs.build/

### DIFF
--- a/core-developers/become-core-developer.rst
+++ b/core-developers/become-core-developer.rst
@@ -54,8 +54,10 @@ are granted through these steps:
    - Add their details to `🔒 python/voters <https://github.com/python/voters>`_.
    - Update the devguide to publicly list their team membership
      at :ref:`developers`.
-   - Send an announcement email to the `Committers Discourse category
-     <https://discuss.python.org/c/committers/5>`_.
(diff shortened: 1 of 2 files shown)


core-developers/become-core-developer.rst
```diff
@@ -54,8 +54,10 @@ are granted through these steps:
    - Add their details to `🔒 python/voters <https://github.com/python/voters>`_.
    - Update the devguide to publicly list their team membership
      at :ref:`developers`.
-   - Send an announcement email to the `Committers Discourse category
-     <https://discuss.python.org/c/committers/5>`_.
+   - Post an announcement in the `Committers Discourse category
+     <https://discuss.python.org/c/committers/5>`_.  The past few announcements
```
hugovk, line 58: Thanks for updating this too.

"The past few" will go out of date -- unless we make sure to keep doing it this way!

Perhaps word as "This can be in the form of..."?

ambv: How about we'll update it if it goes out of date :D

PR #1274, opened 2024-02-12
Remove reference to core-workflow

eltoder:
cherry_picker is in its own repo now.

----
📚 Documentation preview 📚: https://cpython-devguide--1274.org.readthedocs.build/


Mariatta (approved):
Thanks. Great catch.

PR #1277, opened 2024-02-23
Document how to use varargs in Argument Clinic

aisk:
This is related to https://github.com/python/cpython/issues/64490

----
📚 Documentation preview 📚: https://cpython-devguide--1277.org.readthedocs.build/


ezio-melotti (commented):
This might be higher up in the list.
Should `**kwargs`` be documented too, possibly together with `*args`?

aisk:
Thanks for the review!

> This might be higher up in the list.

Is it ok to move it below [How to convert METH_O and METH_NOARGS functions](https://devguide.python.org/development-tools/clinic/#how-to-convert-meth-o-and-meth-noargs-functions)?

> Should `**kwargs` be documented too, possibly together with `*args`?

According to https://github.com/python/cpython/issues/64490, `**kwargs` is not supported by Argument Clinic.

willingc:
@ezio-melotti You have more experience with this tool than I do. Is this PR ready for merge?

ezio-melotti:
LGTM. @erlend-aasland?

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,23 @@ You can still use a self converter, a return converter, and specify
 a *type* argument to the object converter for :c:macro:`METH_O`.
 
 
+How to convert var-positional parameter functions
+-------------------------------------------------
+
+To convert a var-positional parameter function,
+prepending the parameter name with ``*`` ,
+and the parameter should use the ``object`` converter::
```
JelleZijlstra, line 1503: What happens if I use something else? E.g., what if I want to accept an arbitrary number of ints?

Also, what type does the implementation function receive?

aisk: Thank you for the review! I have added a new paragraph to explain more about it.

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,27 @@ You can still use a self converter, a return converter, and specify
...
+   /*[clinic input]
+   var_positional_sample
+
+       foo: int
+       *args: object
+   [clinic start generated code]*/
+
+The implementation function will receive var-positional arguments
+as a tuple, you can either use the tuple directly
+or parsing them into C types using :c:func:`!PyArg_Parse*` functions.
```
picnixz, line 1514:
```suggestion
The implementation function will receive var-positional arguments
as a tuple; you can either use the tuple directly or parse it into
C types via :c:func:`PyArg_ParseTuple` or :c:func:`PyArg_UnpackTuple`.
```

Am I right concerning the functions to use? I don't think we have other
functions that would be relevant in this specific case.

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,27 @@ You can still use a self converter, a return converter, and specify
...
+
+.. versionadded:: 3.11
```
picnixz, line 1516: Argument Clinic is somewhat an internal tool so I'm not sure we should mention when it was added, or do we?

willingc: Since this is the devguide @picnixz, I think it is fine to mention or leave out.

development-tools/clinic.rst
JelleZijlstra: I think it's OK to skip this part. It's important to mention that it's a tuple; for figuring out what to do with the tuple we can point to the C API documentation.
```suggestion
The implementation function will receive var-positional arguments
as a tuple.
```

picnixz: I'm also ok if we omit the C API mention!

development-tools/clinic.rst
picnixz: I'll leave it at aisk's discretion!

willingc (approved):

Issue #1278, opened 2024-02-23
Replace GDB instructions with a link to the new HOWTO


encukou:
GDB instructions have moved to the main CPython docs. We shouldn't keep a duplicate here.

I have a PR almost ready, just waiting for the docs rebuild so I can copy the URL. But I'll be offline during the weekend, so I'll probably send it Monday.

PR #1279, opened 2024-02-26
Replace most of "GDB Support" by a link to the main CPython docs

encukou:
GDB instructions have moved to the main CPython docs. We shouldn't keep a duplicate here.

This leaves two tips that aren't all that useful for extensions.

Fixes: #1278

----
📚 Documentation preview 📚: https://cpython-devguide--1279.org.readthedocs.build/

### DIFF
--- a/development-tools/gdb.rst
+++ b/development-tools/gdb.rst
@@ -6,345 +6,19 @@ GDB support
 
 .. highlight:: none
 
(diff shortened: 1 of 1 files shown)


willingc (approved):
Looks great. Thanks @encukou.

PR #1281, opened 2024-02-28
Document the OS-Unsupported label and Platforms experts list

encukou:
The experts list is perhaps not the best place for the Platforms table, but moving it is left to another PR.

I've removed the blank Android line in anticipation of PEP-738 getting accepted; if it doesn't make it I'll ask Malcolm to add it back with his name.

----
📚 Documentation preview 📚: https://cpython-devguide--1281.org.readthedocs.build/

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -289,8 +289,9 @@ Platforms
 
 For official contacts for supported platforms, see :pep:`11`.
 
-Platforms listed here are not necessarily supported by CPython, and
-the maintainers are not necessarily core CPython developers.
+Platforms listed here are not necessarily supported by CPython.
+Some of the experts listed here maintain and distribute Python
+for “their” platform as a third-party project.
 
 ===================   ===========
 Platform              Maintainers
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -282,14 +282,20 @@ PEG Generator       gvanrossum, pablogsal, lysnikolaou
 ==================  ===========
 
 
+.. _platform-experts:
+
 Platforms
 =========
+
+For official contacts for supported platforms, see :pep:`11`.
+
+Platforms listed here are not necessarily supported by CPython, and
+the maintainers are not necessarily core CPython developers.
 
 ===================   ===========
 Platform              Maintainers
 ===================   ===========
 AIX                   David.Edelsohn^
-Android
 Cygwin                jlt63^, stutzbach^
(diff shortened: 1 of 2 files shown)


core-developers/experts.rst
```diff
@@ -282,14 +282,20 @@ PEG Generator       gvanrossum, pablogsal, lysnikolaou
...
 
+.. _platform-experts:
+
 Platforms
 =========
 
+For official contacts for supported platforms, see :pep:`11`.
+
+Platforms listed here are not necessarily supported by CPython, and
+the maintainers are not necessarily core CPython developers.
```
hugovk, line 293: > ... the maintainers are not necessarily core CPython developers.

This echoes the first sentence of this page:

> ... names for each item that indicate a maintainer or an expert in the field.

But fine to keep if you wish to re-emphasise it here?

hugovk (approved):

triage/labels.rst
```diff
@@ -61,9 +61,14 @@ OS labels
...
 Since most issues either affect all systems or are specific to Unix,
-the only available labels are :gh-label:`OS-windows`, :gh-label:`OS-mac`,
-and :gh-label:`OS-freebsd`.
-
+we don't have a dedicated Unix label.
+Use :gh-label:`OS-windows`, :gh-label:`OS-mac`, and :gh-label:`OS-freebsd`.
+
+Use the :gh-label:`OS-unsupported` label for issues on platforms outside the
+support tiers defined in :pep:`11`. Applying this label adds the issue to
+`a project <https://github.com/orgs/python/projects/27/views/1>`_ where
```
erlend-aasland, line 69:
```suggestion
`a GitHub project <https://github.com/orgs/python/projects/27/views/1>`_ where
```

erlend-aasland (approved):

core-developers/experts.rst
encukou: I do wish to re-emphasize, because this is more prominent in this table. Core devs listed here should ideally be PEP-11 contacts, and there's not much use in repeating them here. (That is, I don't think this is a good place for macOS and Windows, but removing them isn't worth the churn at this point.)

But maybe it's not important to say that here. My latest commit adds this instead:

> Platforms listed here are not necessarily supported by CPython.
Some of the experts listed here maintain and distribute Python
for “their” platform as a third-party project.

IMO that's a more interesting fact :)

erlend-aasland (approved):

PR #1284, opened 2024-03-05
removing `.exe` extension from mac tab

ryanparsa:
# Removing `.exe` extension from Mac tab

in the setup guide section, for Mac users, wrongly used the `.exe` extension for `python`. I just removed that.

----
📚 Documentation preview 📚: https://cpython-devguide--1284.org.readthedocs.build/

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -72,10 +72,10 @@ instructions please see the :ref:`setup guide <setup>`.
 
       .. code-block:: shell
 
-         ./python.exe -m test -j3
+         ./python -m test -j3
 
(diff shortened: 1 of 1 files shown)


hugovk:
This is not wrong, it really does have .exe when you build from source for the reason given in the note.

For example:

```console
❯ ./python.exe
Python 3.13.0a4+ (heads/main:87faec28c7, Mar  3 2024, 16:25:51) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
```

PR #1287, opened 2024-03-08
Remove HTTPS-based suggestion for cloning cpython

davidlowryduda:
Github removed support for password authentication through HTTPS on August 13, 2021. We shouldn't recommend that users use an HTTPS clone when an SSH-based `git@github.com:<username>/cypthon.git` connection is necessary.

This small change removes a comment that suggests that one *could* use HTTPS-based URLs.

----
📚 Documentation preview 📚: https://cpython-devguide--1287.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -82,8 +82,6 @@ You will only need to execute these steps once per machine:
 
       $ git clone git@github.com:<username>/cpython.git
 
-   (You can use both SSH-based or HTTPS-based URLs.)
-
 .. Step 6 and 7 are are duplicated in bootcamp as well.
    Please update these steps in both places.
 


hugovk:
I have an HTTPS-based clone:

```console
❯ git remote -v
origin  https://github.com/hugovk/cpython.git (fetch)
origin  https://github.com/hugovk/cpython.git (push)
upstream        https://github.com/python/cpython.git (fetch)
upstream        https://github.com/python/cpython.git (push)
```

And it's still in the UI:

image

The question mark links to:

https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories

Which explains how HTTPS-based clones are supported.

davidlowryduda:
Reading through the docs again, I see now that it is possible to use HTTPS if you set up personal access tokens with github. I forgot that this was possible.

I should close this then.

PR #1289, opened 2024-03-14
Revise specific guidance about avoiding 'CPU'

hugovk:
Fixes https://github.com/python/docs-community/issues/109.

----
📚 Documentation preview 📚: https://cpython-devguide--1289.org.readthedocs.build/


willingc (commented):
@hugovk In case it is helpful. Here's the language that I shared with the Editorial Board last month:

Allow use of CPU in text. It is common in today's language. There is no need to expand to "Central Processing Unit" or "processor or define the term.

hugovk:
Thanks, I condensed those down to:

```rst
CPU
   Central processing unit. No need to spell out.
```

PR #1290, opened 2024-03-19
Footnotes are discouraged

nedbat:
This is a better way to explain more about `literalinclude`

----
📚 Documentation preview 📚: https://cpython-devguide--1290.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -661,13 +661,20 @@ Syntax highlighting is handled in a smart way:
 
 Longer displays of verbatim text may be included by storing the example text in
 an external file containing only plain text.  The file may be included using the
-``literalinclude`` directive. [1]_ For example, to include the Python source
+``literalinclude`` directive. For example, to include the Python source
 file :file:`example.py`, use::
 
    .. literalinclude:: example.py
 
 The file name is relative to the current file's path.  Documentation-specific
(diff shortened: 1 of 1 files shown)


documentation/markup.rst
```diff
@@ -661,14 +661,20 @@ Syntax highlighting is handled in a smart way:
...
 
    .. literalinclude:: example.py
 
 The file name is relative to the current file's path.  Documentation-specific
 include files should be placed in the ``Doc/includes`` subdirectory.
 
+.. note::
+   There is a standard ``include`` directive, but it raises errors if the
+   file is not found.  ``literalinclude`` is preferred because it only emits a
+   warning in that case.
```
willingc, line 675: @nedbat Can you clarify "in that case" please? Perhaps "...because it emits a warning instead of an error."

willingc (approved):

hugovk (approved):

documentation/markup.rst
nedbat: done.

ezio-melotti (commented):
The change from footnote to regular note is ok, but why would a warning be preferred over an error?

nedbat:
I assumed it was because it doesn't stop the build, but I could be wrong.

PR #1291, opened 2024-03-19
Explain that a venv is being created

nedbat:
The first time I used `make html` I panicked when I saw lots of packages being installed and I didn't know where they were going.


willingc (approved):

hugovk (approved):
Want to update the one in the CPython Doc dir and PEPs repo too?

PR #1292, opened 2024-03-19
Move markup concerns from style to markup

nedbat:
I'm preparing to add more linguistic style to the style guide.  These two sections are about the logistics of .rst files, so I'm moving them to the markup page.

----
📚 Documentation preview 📚: https://cpython-devguide--1292.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -52,6 +52,22 @@ language, this will not take too long.
     Documentation <https://docutils.sourceforge.io/rst.html>`_.
 
 
+Use of whitespace
+-----------------
+
+All reST files use an indentation of 3 spaces; no tabs are allowed.  The
+maximum line length is 80 characters for normal text, but tables, deeply
+indented code samples and long links may extend beyond that.  Code example
+bodies should use normal Python 4-space indentation.
+
+Make use of multiple blank lines where applicable to clarify the structure of
+the reST file.  Extra blank lines help group sections together to make the
+organization of the file clearer.
+
+A sentence-ending period may be followed by one or two spaces. While reST
+ignores the second space, it is customarily put in by some users, for example
+to aid Emacs' auto-fill mode.
+
 Paragraphs
 ----------
 
@@ -288,6 +304,28 @@ There are some problems one commonly runs into while authoring reST documents:
 * **Separation of inline markup:** As said above, inline markup spans must be
   separated from the surrounding text by non-word characters, you have to use
   an escaped space to get around that.
+
+
+Typographic conventions
+=======================
+
+Big *O* notation
+----------------
+
+Big *O* notation is used to describe the performance of algorithms.
+
+Use italics for the big *O* and variables. For example:
+
+======================== ====================
+reStructuredText         Rendered
+======================== ====================
+``*O*\ (1)``             *O*\ (1)
+``*O*\ (log *n*)``       *O*\ (log *n*)
+``*O*\ (*n*)``           *O*\ (*n*)
+``*O*\ (*n* log *n*)``   *O*\ (*n* log *n*)
+``*O*\ (*n*\ :sup:`2`)`` *O*\ (*n*\ :sup:`2`)
+======================== ====================
+
 
 .. _additional-markup-constructs:
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -4,25 +4,11 @@
 Style guide
 ===========
 
-.. highlight::  rest
+.. highlight:: rest
 
-This document describes the style guide for our documentation.
+This page describes the linguistic style guide for our documentation.
+For markup details in reST files, see :ref:`markup`.
 
-
-Use of whitespace
-=================
-
-All reST files use an indentation of 3 spaces; no tabs are allowed.  The
-maximum line length is 80 characters for normal text, but tables, deeply
-indented code samples and long links may extend beyond that.  Code example
-bodies should use normal Python 4-space indentation.
-
-Make generous use of blank lines where applicable; they help group things
-together.
-
-A sentence-ending period may be followed by one or two spaces; while reST
-ignores the second space, it is customarily put in by some users, for example
-to aid Emacs' auto-fill mode.
 
 Footnotes
 =========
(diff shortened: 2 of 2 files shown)


jablonskidev (approved):
Agreed, splitting the two is overdue.

documentation/markup.rst
```diff
@@ -52,6 +52,21 @@ language, this will not take too long.
...
+maximum line length is 80 characters for normal text, but tables, deeply
+indented code samples and long links may extend beyond that.  Code example
+bodies should use normal Python 4-space indentation.
+
+Make generous use of blank lines where applicable; they help group things
+together.
+
+A sentence-ending period may be followed by one or two spaces. While reST
+ignores the second space, it is customarily put in by some users, for example
+to aid Emacs' auto-fill mode.
```
hugovk, line 68: The first paragraph is indeed about the technical rules of reST markup. However, the second is about writing style. The last seems a bit of a mix, but really about reST.

I don't mind them all being moved here, if there's no good place for the second one in the style guide.

documentation/markup.rst
```diff
@@ -291,6 +306,28 @@ There are some problems one commonly runs into while authoring reST documents:
 
 .. _additional-markup-constructs:
```
hugovk, line 307: This should be moved from here:
```suggestion
```

documentation/markup.rst
```diff
@@ -291,6 +306,28 @@ There are some problems one commonly runs into while authoring reST documents:
...
+======================== ====================
+``*O*\ (1)``             *O*\ (1)
+``*O*\ (log *n*)``       *O*\ (log *n*)
+``*O*\ (*n*)``           *O*\ (*n*)
+``*O*\ (*n* log *n*)``   *O*\ (*n* log *n*)
+``*O*\ (*n*\ :sup:`2`)`` *O*\ (*n*\ :sup:`2`)
+======================== ====================
+
+
 Additional markup constructs
```
hugovk, line 331: ... to here:

```suggestion
.. _additional-markup-constructs:

Additional markup constructs
```

hugovk (changes requested):

documentation/markup.rst
nedbat: The second and third paragraphs are only about the .rst files: blank lines and extra spaces aren't rendered into the output.  These guidelines don't affect what the reader sees, so they belong in the markup concerns.

documentation/markup.rst
nedbat: oops, fixed.

documentation/markup.rst
hugovk: Ah, does the second just mean single line breaks (`\n`)?

Like this:

```rst
An expression-less :keyword:`!except` clause, if present, must be last; it matches any 
exception. For an :keyword:`!except` clause with an expression, that expression is 
evaluated, and the clause matches the exception if the resulting object is "compatible" 
with the exception.  An object is compatible with an exception if the object is the 
class or a :term:`non-virtual base class <abstract base class>` of the exception object, 
or a tuple containing an item that is the class or a non-virtual base class of the 
exception object.
```

Is indeed rendered the same as:

```rst
An expression-less :keyword:`!except` clause, if present, must be last;
it matches any exception.
For an :keyword:`!except` clause with an expression,
that expression is evaluated, and the clause matches the exception
if the resulting object is "compatible" with the exception.  An object is
compatible with an exception if the object is the class or a
:term:`non-virtual base class <abstract base class>` of the exception object,
or a tuple containing an item that is the class or a non-virtual base class
of the exception object.
```

Or is it talking about new empty lines (`\n\n`)? Because if I add extra empty lines here, this is rendered as multiple paragraphs, which is style:

```rst
An expression-less :keyword:`!except` clause, if present, must be last;
it matches any exception.

For an :keyword:`!except` clause with an expression,
that expression is evaluated, and the clause matches the exception
if the resulting object is "compatible" with the exception.

An object is
compatible with an exception if the object is the class or a
:term:`non-virtual base class <abstract base class>` of the exception object,
or a tuple containing an item that is the class or a non-virtual base class
of the exception object.
```

willingc: I think the second paragraph could be in the style guide. The first and third belong in markup. If we want to mention space after a sentence in the style guide, we should recommend one space for any new docs.

nedbat: Because of `they help group things together,` I interpreted this advice as encouraging people to sometimes use two blank lines instead of just one if it would make the structure of the document clearer.

willingc: We could call it something different than whitespace.

nedbat: I still think this paragraph is about the reST file, not the text as visible to the reader.  I've changed it to:
```
Make use of multiple blank lines where applicable to clarify the structure of
the reST file.  Extra blank lines help group sections together to make the
organization of the file clearer.
```

hugovk: Multiple blank lines do make a difference to the text as visible to the reader.

Here's the three chunks above, rendered. 

* ONE and TWO are the same except for single newlines, and are the same for the reader.

* But THREE has multiple newlines and they become paragraphs.


image

nedbat: Yes, but one and two don't differ in the number of blank lines. They differ in the number of newlines. They have no blank lines.

I understand the value of wisely choosing paragraph breaks. I don't think that's what this text was about.

documentation/markup.rst
ezio-melotti, line 1: FTR this page was initially copied from <https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html>.  I don't think your additions are a problem though, since it has already diverged from the original over the years.

documentation/markup.rst
hugovk: Okay, if that's the case, then we agree.

A bit of archaeology, seems this wasn't in [2.5](https://docs.python.org/release/2.5/doc/style-guide.html) and was added in [2.6](https://github.com/python/cpython/commit/8ec7f656134b1230ab23003a94ba3266d7064122), which incidentally is when the docs were converted from Latex to reST.

---

As it happens, the CPython style guide used to recommend primarily following the Apple Publications Style Guide, but that advice was removed when the link broke (https://github.com/python/cpython/issues/62221). Here's an [archived copy](https://web.archive.org/web/20030630102016/http://developer.apple.com/documentation/UserExperience/Conceptual/APStyleGuide/AppleStyleGuide2003.pdf), and [Apple's current style guide](https://support.apple.com/en-gb/guide/applestyleguide/welcome/web). Not suggesting we reinstate it, just sharing some history :)

hugovk (approved):

Issue #1293, opened 2024-03-19
Document only adding links once per paragraph


willingc:
https://github.com/python/docs-community/issues/52#issuecomment-1925884904

There is general consensus on this. We should add to the markup section of the doc page.

PR #1294, opened 2024-03-20
Avoid over-linking #1293

nedbat:
Summarizing and codifying the advice discussed in https://github.com/python/docs-community/issues/52

----
📚 Documentation preview 📚: https://cpython-devguide--1294.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -754,7 +754,7 @@ versatile:
   tool-tip on mouse-hover) will always be the full target name.
 
 * Combining ``~`` and ``!`` (for example, ``:meth:`~!Queue.Queue.get```) is not
-  supported.  You can obtain the same result by simply using `!` and the last
+  supported.  You can obtain the same result by simply using ``!`` and the last
   component of the target (for example, ``:meth:`!get```).
 
 The following roles refer to objects in modules and are possibly hyperlinked if
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -30,20 +30,10 @@ external links          ```Link text <https://example.com>`_``      :ref:`hyperl
 roles w/ custom text    ``:role:`custom text <target>```            :ref:`roles`
 roles w/ only last part ``:role:`~hidden.hidden.visible```          :ref:`roles`
 roles w/o link          ``:role:`!target```                         :ref:`roles`
-roles w/o link, short   ``:role:`!visible``` (Note 1)               :ref:`roles`
 issues                  ``:gh:`ID```, ``:issue:`ID```               :ref:`roles`
 CPython source          ``:source:`PATH```                          :ref:`roles`
 comments                ``.. a comment``                            :ref:`comments`
 ======================= =========================================== ====================
-
-Notes:
-
-(1)
-   For an only-last-part reference with a suppressed link,
-   ``:role:`~!hidden.visible``` makes more sense, but it causes
-   a warning as Sphinx tries to look up the reference ``!hidden.visible``
-   which does not exist.  The shorter form ``:role:`!visible`` renders as
-   desired and will build successfully.
 
 
 .. _rst-primer:
@@ -762,6 +752,10 @@ versatile:
 
   In HTML output, the link's ``title`` attribute (that is e.g. shown as a
   tool-tip on mouse-hover) will always be the full target name.
+
+* Combining ``~`` and ``!`` (for example, ``:meth:`~!Queue.Queue.get```) is not
+  supported.  You can obtain the same result by simply using `!` and the last
+  component of the target (for example, ``:meth:`!get```).
 
 The following roles refer to objects in modules and are possibly hyperlinked if
 a matching identifier is found:
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -30,10 +30,20 @@ external links          ```Link text <https://example.com>`_``      :ref:`hyperl
 roles w/ custom text    ``:role:`custom text <target>```            :ref:`roles`
 roles w/ only last part ``:role:`~hidden.hidden.visible```          :ref:`roles`
 roles w/o link          ``:role:`!target```                         :ref:`roles`
+roles w/o link, short   ``:role:`!visible``` (Note 1)               :ref:`roles`
 issues                  ``:gh:`ID```, ``:issue:`ID```               :ref:`roles`
 CPython source          ``:source:`PATH```                          :ref:`roles`
 comments                ``.. a comment``                            :ref:`comments`
 ======================= =========================================== ====================
+
+Notes:
+
+(1)
+   For an only-last-part reference with a suppressed link,
+   ``:role:`~!hidden.visible``` makes more semantic sense, but it causes
+   a warning as Sphinx tries to look up the reference ``!hidden.visible``
+   which does not exist.  The shorter form ``:role:`!visible`` renders as
+   desired and will build successfully.
 
 
 .. _rst-primer:
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -754,7 +754,7 @@ versatile:
   tool-tip on mouse-hover) will always be the full target name.
 
 * Combining ``~`` and ``!`` (for example, ``:meth:`~!Queue.Queue.get```) is not
-  supported.  You can obtain the same result by simply using ``!`` and the last
+  supported.  You can obtain the same result by using ``!`` and the last
   component of the target (for example, ``:meth:`!get```).
 
 The following roles refer to objects in modules and are possibly hyperlinked if
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -40,7 +40,7 @@ Notes:
 
 (1)
    For an only-last-part reference with a suppressed link,
-   ``:role:`~!hidden.visible``` makes more semantic sense, but it causes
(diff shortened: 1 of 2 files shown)


documentation/style-guide.rst
```diff
@@ -124,11 +124,37 @@ explanation.
...
+=====
+
+Linking words to more information about those words is a powerful tool for
+helping people navigate documentation, but links can be over-used.
+
+Links should be provided only if they help the user.  If a link will not help
+the user, do not create a link.  You may need to suppress a link that Sphinx
+would have created automatically.
+
+Generally, a link should be provided for the first use of a term in a unit This
```
willingc, line 142:
```suggestion
Generally, a link should be provided for the first use of a term in a unit, such as a section or paragraph. This
```

documentation/style-guide.rst
```diff
@@ -124,11 +124,37 @@ explanation.
...
+is not a hard and fast rule.  Sometimes the second mention is a more
+appropriate jumping off point.  Some units are long enough to have a few
```
willingc, line 144:
```suggestion
appropriate for a link.  Some units are long enough to have a few
```

willingc: @nedbat I'm not sure that "jumping off point" is commonly used globally.

documentation/style-guide.rst
```diff
@@ -124,11 +124,37 @@ explanation.
...
 Please consult the `Diátaxis <https://diataxis.fr/>`_ guide for more
 detail.
 
+Links
+=====
+
+Linking words to more information about those words is a powerful tool for
+helping people navigate documentation, but links can be over-used.
+
+Links should be provided only if they help the user.  If a link will not help
```
willingc, line 138:
```suggestion
Links should be provided only if they help the reader.  If a link will not help
```

documentation/style-guide.rst
```diff
@@ -124,11 +124,37 @@ explanation.
...
+the user, do not create a link.  You may need to suppress a link that Sphinx
```
willingc, line 139: I'm unclear on suppressing a link. Perhaps omit from this paragraph and move further down the section.

willingc (commented):
Thanks Ned!

jablonskidev (approved):

documentation/style-guide.rst
nedbat: Done.

documentation/style-guide.rst
nedbat: Done.

nedbat:
Changes made, and I added another rule: no links in section titles.

hugovk (approved):

documentation/markup.rst
```diff
@@ -30,11 +30,21 @@ external links          ```Link text <https://example.com>`_``      :ref:`hyperl
...
 issues                  ``:gh:`ID```, ``:issue:`ID```               :ref:`roles`
 CPython source          ``:source:`PATH```                          :ref:`roles`
 comments                ``.. a comment``                            :ref:`comments`
 ======================= =========================================== ====================
 
+Notes:
+
+(1)
+   For an only-last-part reference with a suppressed link,
+   ``:role:`~!hidden.visible``` makes more semantic sense, but it causes
```
merwok, line 43: semantic (adj): related to meaning, or to the study of meaning

Therefore semantic sense or semantic meaning is redundant 🙂 

```suggestion
   ``:role:`~!hidden.visible``` makes more sense, but it causes
```

documentation/markup.rst
```diff
@@ -30,11 +30,21 @@ external links          ```Link text <https://example.com>`_``      :ref:`hyperl
...
+   a warning as Sphinx tries to look up the reference ``!hidden.visible``
+   which does not exist.  The shorter form ``:role:`!visible`` renders as
+   desired and will build successfully.
```
ezio-melotti, line 46: I'm not sure this should be included here.  The quick reference lists commonly used markup, and what's being described here is not very common.

I would move this paragraph in the linked ["roles" section](https://devguide.python.org/documentation/markup/#roles), where the nuances of using `!` and `~` are covered in more detail.

documentation/style-guide.rst
```diff
@@ -124,11 +124,39 @@ explanation.
...
-  :ref:`python:unicode-howto`
+  :ref:`python:unicode-howto`.
 
 Please consult the `Diátaxis <https://diataxis.fr/>`_ guide for more
 detail.
 
+Links
+=====
+
+Linking words to more information about those words is a powerful tool for
```
ezio-melotti, line 135: I would rephrase this sentence.

documentation/style-guide.rst
```diff
@@ -124,11 +124,39 @@ explanation.
...
+
+Do not use links in section headers.  They distract from the title of the
+section.  The term will be mentioned in the paragraph text and can be linked
+from there.
+
+Sphinx provides ways to automatically add links to references, and a way to
+suppress the link.  Using roles like ``:func:`map``` will link to the
+documentation for ``map``.  You can suppress the link while keeping the
+semantic presentation of the function name by adding an exclamation point
+prefix: ``:func:`!map```.
```
ezio-melotti, line 158: Here you could link to the "roles" section of `markup.rst`.

documentation/markup.rst
nedbat: It took a few developers a while to work out what happened when trying to combine `~!`, so I think it will be helpful to leave here.

documentation/style-guide.rst
nedbat: Done

documentation/style-guide.rst
nedbat: Done

documentation/markup.rst
ezio-melotti: What I'm thinking is:
* the quick reference table is meant mostly for readers that already know about the existence and functioning of those constructs, but don't quite remember the syntax
* all further details and explanations shouldn't (and aren't) covered here, but in the sections linked in the third column
* if users look at the "roles" section (where `~` and `!` are explained) for ways to combine `~`/`!` they won't find it there, unless you duplicate the note there
* the wording of `roles w/o link, short` is not too clear, since `short` doesn't mirror `only last part` (understandably, due to lack of space)
* a reader might be wondering how to combine `~` and `!` together, but `` :role:`!visible` `` doesn't seem an obvious answer to the question (unless they keep scrolling and read the note), especially since it doesn't have any apparent difference from  `` :role:`!target` ``
* the consequences of combining `~` and `!` are a Sphinx implementation detail that shouldn't matter to the reader, so I would omit them from the note

My suggestion is to add:
```rst
* Combining ``~`` and ``!`` (e.g. ``:meth:`~!Queue.Queue.get```) is not supported.
  You can obtain the same result by simply using `!` and the last component of
  the target (e.g. ``:meth:`!get```).
```
in the "roles" section, after the bullet points that already explain `~` and `!` and use `Queue.Queue.get` as an example.

The entry in the table could be removed altogether, since it's not a very common construct (trusting that the reader will follow the link looking for more information).  If you prefer to keep it, the table should be updated by:
* widening the first column and using `roles w/o link, only last part` to mirror the text of `roles w/ only last part`
* reordering the entries so that `roles w/o link` comes first, `roles w/ only last part` next, and `roles w/o link, only last part` just after, so that the use of `!visible` is more evident.

willingc: @nedbat Let me take some time to read @ezio-melotti's comment. Ezio makes some good points especially considering the original intent of the chart. There needs to be a new doc contributor cheat sheet (that is useful to current contributors) to guide folks on many of these markup constructs.

nedbat: @ezio-melotti I like your suggestion, and have made it in the latest commit.

ezio-melotti (approved):

documentation/markup.rst
```diff
@@ -753,6 +753,10 @@ versatile:
   In HTML output, the link's ``title`` attribute (that is e.g. shown as a
   tool-tip on mouse-hover) will always be the full target name.
 
+* Combining ``~`` and ``!`` (for example, ``:meth:`~!Queue.Queue.get```) is not
+  supported.  You can obtain the same result by simply using ``!`` and the last
```
hugovk, line 757: Don't say simply: 
* https://www.youtube.com/watch?v=gsT2BBWBVmM
* https://blog.knowledgeowl.com/blog/posts/dont-say-simply-jim-fisher/
* https://jameshfisher.com/2017/02/22/dont-use-simply/


```suggestion
  supported.  You can obtain the same result by using ``!`` and the last
```

documentation/style-guide.rst
```diff
@@ -124,11 +124,39 @@ explanation.
...
+
+Generally, a link should be provided for the first use of a term in a unit,
+such as a section or paragraph. This is not a hard and fast rule.  Sometimes
+the second mention is more appropriate for a link.  Some units are long enough
+to have a few repeated links.  Use judgement to decide when a link will help
+the reader.
+
+Do not use a link when the link would point to the current unit.  It's natural
+to use the name of a function in the documentation for the function, but a link
+on that function name that simply reloads the section the user is already
```
hugovk, line 147: (I think this "simply" is fine, it's describing the redundant behaviour.)

hugovk:
Thanks!

willingc:
Thanks @nedbat

Issue #1285, opened 2024-03-05
Document env var style


hugovk:
In https://discuss.python.org/t/change-environment-variable-style/35180 we decided to use underscores in new environment variable names introduced in Python 3.13+.

For example, `PYTHON_CPU_COUNT` instead of `PYTHONCPUCOUNT`.

Let's document it somewhere in the devguide.

hugovk:
Please see PR https://github.com/python/devguide/pull/1300.

ezio-melotti:
Are there any mechanisms in place to e.g. ignore the `_`s so that both `PYTHON_CPU_COUNT` and `PYTHONCPUCOUNT` work?

hugovk:
Inside Python? Maybe, but we'd need to decide what to do when both exist, with different values, which could be set by different groups. Who wins? See https://discuss.python.org/t/change-environment-variable-style/35180/7, and the Discourse thread is the place to discuss this :)

PR #1300, opened 2024-03-25
Document style on naming env vars

hugovk:
Fixes https://github.com/python/devguide/issues/1285.

I wasn't sure where to put it, so went for `developer-workflow/stdlib.html`.

Is there a better place?

----
📚 Documentation preview 📚: https://cpython-devguide--1300.org.readthedocs.build/

### DIFF
--- a/developer-workflow/stdlib.rst
+++ b/developer-workflow/stdlib.rst
@@ -142,8 +142,8 @@ once the authors of the module sign
 Adding a new environment variable
 ---------------------------------
 
-From Python 3.13 onwards, name environment variables with underscores for
-readability and accessibility.
+Names of environment variables should be uppercase and, from Python 3.13
+onwards, use underscores for readability and accessibility.
 
 For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
 
--- a/developer-workflow/stdlib.rst
+++ b/developer-workflow/stdlib.rst
@@ -139,8 +139,8 @@ once the authors of the module sign
 
 .. _issue tracker: https://github.com/python/cpython/issues
 
-Adding a new env var
---------------------
+Adding a new environment variable
+---------------------------------
 
 From Python 3.13 onwards, name environment variables with underscores for
 readability and accessibility.
@@ -150,5 +150,5 @@ For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
 See also:
 
 * :ref:`python:using-on-envvars`
-* `Discourse discussion
+* `"Change environment variable style" Discourse discussion
   <https://discuss.python.org/t/change-environment-variable-style/35180>`__
--- a/developer-workflow/stdlib.rst
+++ b/developer-workflow/stdlib.rst
@@ -52,6 +52,7 @@ useful.
 
 Adding a new module
 -------------------
+
 It must be stated upfront that getting a new module into the stdlib is very
 difficult. Adding any significant amount of code to the stdlib increases the
 burden placed upon core developers. It also means that the module somewhat
@@ -137,3 +138,17 @@ once the authors of the module sign
 :ref:`contributor agreements <contributor_agreement>`.
 
 .. _issue tracker: https://github.com/python/cpython/issues
+
+Adding a new env var
+--------------------
+
+From Python 3.13 onwards, name environment variables with underscores for
+readability and accessibility.
+
+For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
+
+See also:
+
+* :ref:`python:using-on-envvars`
+* `Discourse discussion
+  <https://discuss.python.org/t/change-environment-variable-style/35180>`__


developer-workflow/stdlib.rst
```diff
@@ -137,3 +138,17 @@ once the authors of the module sign
...
+From Python 3.13 onwards, name environment variables with underscores for
+readability and accessibility.
+
+For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
+
+See also:
+
+* :ref:`python:using-on-envvars`
+* `Discourse discussion
+  <https://discuss.python.org/t/change-environment-variable-style/35180>`__
```
ezio-melotti, line 154:
```suggestion
* `"Change environment variable style" Discourse discussion
  <https://discuss.python.org/t/change-environment-variable-style/35180>`__
```

Even if in context it's understandable, it's better to avoid generic link text.

developer-workflow/stdlib.rst
```diff
@@ -137,3 +138,17 @@ once the authors of the module sign
 :ref:`contributor agreements <contributor_agreement>`.
 
 .. _issue tracker: https://github.com/python/cpython/issues
+
+Adding a new env var
+--------------------
+
+From Python 3.13 onwards, name environment variables with underscores for
+readability and accessibility.
```
ezio-melotti, line 146: I initially failed to parse ["name" as a verb](https://en.wikipedia.org/wiki/Stroop_effect).  I suggest rephrasing.
```suggestion
From Python 3.13 onwards, environment variables name should use
underscores for readability and accessibility.
```

This sentence also assumes the reader is aware that the names are already uppercase, but it might be worth mentioning it explicitly:
```suggestion
Environment variables name should always be uppercase and, from Python 3.13 onwards,
use underscores for readability and accessibility.
```

developer-workflow/stdlib.rst
```diff
@@ -137,3 +138,17 @@ once the authors of the module sign
 :ref:`contributor agreements <contributor_agreement>`.
 
 .. _issue tracker: https://github.com/python/cpython/issues
+
+Adding a new env var
+--------------------
```
ezio-melotti, line 143:
```suggestion
Adding a new environment variable
---------------------------------
```

developer-workflow/stdlib.rst
```diff
@@ -137,3 +138,17 @@ once the authors of the module sign
...
+
+From Python 3.13 onwards, name environment variables with underscores for
+readability and accessibility.
+
+For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
```
ezio-melotti, line 148:
```suggestion
For example, use ``PYTHON_CPU_COUNT`` instead of ``PYTHONCPUCOUNT``.
```
Not sure this deserves a separate paragraph.

ezio-melotti:
After reviewing the rest of the page, I think this new section should also cover when is it ok to add new envvars and how to do it, even though you might want to do that in a separate PR.

developer-workflow/stdlib.rst
hugovk: Good idea, I started off with something like yours, and tried to be clever and make it shorter but made it less clear 🙃

How about something like this?

```suggestion
Environment variables should be uppercase, and from Python 3.13 onwards, use
underscores for readability and accessibility.
```

developer-workflow/stdlib.rst
hugovk: I think it's clearer like this. If we expand the section, we can adjust as needed.

developer-workflow/stdlib.rst
ezio-melotti: Technically it's the names -- not the variables themselves -- that are uppercase, but the sentence is clear even if "names" is dropped.  Dropping "always" is also ok.
In my suggestion I was going for a [parenthetical expression](https://en.wikipedia.org/wiki/Parenthesis_(rhetoric)), hence the comma after the "and".
```suggestion
Environment variables should be uppercase and, from Python 3.13 onwards,
use underscores for readability and accessibility.
```
or (with "names" added back):
```suggestion
Environment variables names should be uppercase and, from Python 3.13 onwards,
use underscores for readability and accessibility.
```
Not quite sure if a [possessive apostrophe](https://en.wikipedia.org/wiki/Apostrophe#Possessive_apostrophe) is needed here ("variables' names") -- @CAM-Gerlach?

JelleZijlstra: "Names of environment variables" is probably better

developer-workflow/stdlib.rst
```diff
@@ -137,3 +138,17 @@ once the authors of the module sign
 :ref:`contributor agreements <contributor_agreement>`.
 
 .. _issue tracker: https://github.com/python/cpython/issues
+
+Adding a new environment variable
+---------------------------------
+
+From Python 3.13 onwards, name environment variables with underscores for
+readability and accessibility.
```
hugovk: Thanks both, I've combined those into:
```
Names of environment variables should be uppercase and, from Python 3.13
onwards, use underscores for readability and accessibility.
```

JacobCoffee (approved):

PR #1303, opened 2024-04-02
Python 3.11 is now security-only

hugovk:
3.11.9 was the last bugfix with binary installers.

* https://peps.python.org/pep-0664/

----
📚 Documentation preview 📚: https://cpython-devguide--1303.org.readthedocs.build/


encukou:
Thank you!

Issue #1308, opened 2024-04-18
Document `attr` and `attribute`


hugovk:
Let's update https://devguide.python.org/documentation/markup/ to document that:

*  `.. attribute:: my-attr` should be used for defining attributes, and 
* ``` :attr:`my-attr` ``` used for referring to it.

hugovk:
Yes, you can work on it :) We don't normally assign issues, and don't have a deadline, but someone else can have a go if this is still open after a few weeks.

Let me know if you've any questions.

Thank you!

kokoavailable:
@hugovk 

I've submitted a PR.

https://github.com/python/devguide/pull/1311

Please let me know If there are any inaccuracies or misunderstanding in my request.

Thank you!

PR #1311, opened 2024-04-18
Document attr and attribute #1308

kokoavailable:

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -564,6 +564,16 @@ The directives are:
          .. attribute:: ham
 
             Description of the attribute.
+   
+   This directive should be used for defining attributes.::
+
+      .. attribute:: my-attr
+         This attribute `my-attr` should be defined with this directive.
+
+   When referring to an attribute, use the ``:attr:`my-attr``` role::
+      
+      :attr:`my-attr`
+         Use this role to refer to the attribute `my-attr`
 
    If is also possible to document an attribute outside of a class directive,
    for example if the documentation for different attributes and methods is


Mariatta:
Can you mention the Issue number on the PR description (instead of in the title?)

E.g. 
```
Closes #1308
```

kokoavailable:
> First, of all, please could you sign the CLA?

Done! Thank you.

kokoavailable:
> Can you mention the Issue number on the PR description (instead of in the title?)
> 
> E.g.
> 
> ```
> Closes #1308
> ```

Done!!

documentation/markup.rst
```diff
@@ -565,6 +565,16 @@ The directives are:
 
             Description of the attribute.
 
+   This directive should be used for defining attributes.::
+
+      .. attribute:: my-attr
+         This attribute `my-attr` should be defined with this directive.
+
```
hugovk, line 572: We can delete this bit, it's essentially duplicating the "ham" example just before:
```suggestion
```

documentation/markup.rst
```diff
@@ -565,6 +565,16 @@ The directives are:
...
+   When referring to an attribute, use the ``:attr:`my-attr``` role::
+
+      :attr:`my-attr`
+         Use this role to refer to the attribute `my-attr`
+
```
hugovk, line 577: And let's modify this to use "ham", and roles are usually used inline:
```suggestion
   Refer to attributes with ``:attr:`` role::

      Use the :attr:`ham` attribute to spam the eggs.
```
(Or something else [Pythonesque](https://www.oxfordreference.com/display/10.1093/oi/authority.20110803100356667) :)

hugovk (changes requested):
Thanks!

Let's compare "functions definitions" and "functions references" in the quick reference table:

image

Let's add "attribute definitions" and "attribute references" to the table. The attribute definition are already in the "Information units" section, and let's add the references to the "Roles" section.

hugovk:
@python/organization-owners Please could you enable https://pre-commit.ci for this repo?

kokoavailable:
> @python/organization-owners Please could you enable https://pre-commit.ci for this repo?

done!

documentation/markup.rst
```diff
@@ -565,6 +567,11 @@ The directives are:
 
             Description of the attribute.
 
+
+   Refer to an attribute, use the ``:attr:`` role::
```
hugovk, line 571:
```suggestion
   Refer to an attribute using the ``:attr:`` role::
```

hugovk (changes requested):

documentation/markup.rst
```diff
@@ -24,6 +24,8 @@ variables/literals/code ````foo````, ````42````, ````len(s) - 1```` :ref:`inline
 True/False/None         ````True````, ````False````, ````None````   :ref:`inline-markup`
 functions definitions   ``.. function:: print(*args)``              :ref:`directives`
 functions references    ``:func:`print```                           :ref:`roles`
+attribute definitions   ``.. attribute: `attr-name```               :ref:`directives`
```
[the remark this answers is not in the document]
ezio-melotti: I'm not sure this should be added.  In the table I used `.. function:`/`:func:` as an example, but the same applies for classes, methods, attributes and several other directives/roles pairs.

Perhaps a separate table associating directives with the roles that can be used to link to them would be better?  (If this is done, it should be a separate PR.).

documentation/markup.rst
ezio-melotti: While this is useful here, the same could be done for other directives too, like the `class` just above or the `method` just below.  It would be nice if the documentation was consistent.

hugovk (approved):
Thank you!

documentation/markup.rst
kokoavailable: > While this is useful here, the same could be done for other directives too, like the `class` just above or the `method` just below. It would be nice if the documentation was consistent.

should it be seperate pr?

hugovk: Either way is fine, but smaller PRs are easier to review and merge.

Would you like to make a separate follow-up PR? :)

kokoavailable: > Either way is fine, but smaller PRs are easier to review and merge.
> 
> Would you like to make a separate follow-up PR? :)

sure. and when will this pr be merged ? #1311

hugovk: Let's do it now :)

Thank you for the PR!

PR #1313, opened 2024-04-29
Update deps and build instructions wrt. libmpdecimal

erlend-aasland:
- python/cpython#115119

----
📚 Documentation preview 📚: https://cpython-devguide--1313.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -724,9 +724,16 @@ on Linux, macOS and iOS.
 
    Alternatively, with **MacPorts**::
 
-      $ sudo port install pkgconfig openssl xz gdbm tcl tk +quartz
+      $ sudo port install pkgconfig openssl xz gdbm tcl tk +quartz mpdecimal
 
-   Then, for Python 3.11 and newer, run ``configure``::
+   Then, for Python 3.13 and newer, run ``configure``::
+
+      $ GDBM_CFLAGS="-I$(dirname $(dirname $(which port)))/include" \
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -692,9 +692,17 @@ on Linux, macOS and iOS.
 
    For example, with **Homebrew**, install the dependencies::
 
-      $ brew install pkg-config openssl@3.0 xz gdbm tcl-tk
+      $ brew install pkg-config openssl@3.0 xz gdbm tcl-tk mpdecimal
```
ned-deily, line 695: Similar changes should be made to the MacPorts recipe further down.

erlend-aasland: Yes; addressed with 1a9eb7d0e723f75a8639cd5bcd08edb723f86fe3.

ned-deily (approved):
LGTM, although, once we change the configure default for ``--with-system-libmpdec`` there really wouldn't need to be a separate recipe for 3.13, as long as people didn't care which version of libmpdec is being used.

erlend-aasland:
Thanks for the review!

> LGTM, although, once we change the configure default for `--with-system-libmpdec` there really wouldn't need to be a separate recipe for 3.13, [...]

Yep.

> [...] as long as people didn't care which version of libmpdec is being used.

I don't think the devguide recipe is the place for that user group; for people who care about a particular version of a build dependency, we have https://docs.python.org/3/using/configure.html.

PR #1317, opened 2024-05-08
macOS: Add tabs for Homebrew/MacPorts and for 3.13+/3.11-3.12/3.8-3.10

hugovk:
For example:

image

----
📚 Documentation preview 📚: https://cpython-devguide--1317.org.readthedocs.build/getting-started/setup-building/#install-dependencies

PR #1318, opened 2024-05-08
Linux: Remove outdated `yum` instructions

hugovk:
Similar to https://github.com/python/devguide/pull/1317.

For example, `yum` has short instructions:

image

As does `DNF`:

image

But `apt` is much longer, so it's good to tidy it away when not relevant:

image

----
📚 Documentation preview 📚: https://cpython-devguide--1318.org.readthedocs.build/getting-started/setup-building/#install-dependencies

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -619,49 +619,55 @@ on Linux, macOS and iOS.
    distribution, but the appropriate commands for some popular distributions
    are below.
 
-   On **Fedora**, **Red Hat Enterprise Linux** and other ``yum`` based systems::
+   .. tab:: ``yum``
 
-      $ sudo yum install yum-utils
-      $ sudo yum-builddep python3
+      On **Fedora**, **Red Hat Enterprise Linux** and other ``yum`` based systems::
 
-   On **Fedora** and other ``DNF`` based systems::
+         $ sudo yum install yum-utils
+         $ sudo yum-builddep python3
 
-      $ sudo dnf install dnf-plugins-core  # install this to use 'dnf builddep'
-      $ sudo dnf builddep python3
+   .. tab:: ``DNF``
 
-   On **Debian**, **Ubuntu**, and other ``apt`` based systems, try to get the
-   dependencies for the Python you're working on by using the ``apt`` command.
+      On **Fedora** and other ``DNF`` based systems::
 
-   First, make sure you have enabled the source packages in the sources list.
-   You can do this by adding the location of the source packages, including
-   URL, distribution name and component name, to ``/etc/apt/sources.list``.
-   Take Ubuntu 22.04 LTS (Jammy Jellyfish) for example::
+         $ sudo dnf install dnf-plugins-core  # install this to use 'dnf builddep'
+         $ sudo dnf builddep python3
 
-      deb-src http://archive.ubuntu.com/ubuntu/ jammy main
+   .. tab:: ``apt``
 
-   Alternatively, uncomment lines with ``deb-src`` using an editor, e.g.::
+      On **Debian**, **Ubuntu**, and other ``apt`` based systems, try to get the
+      dependencies for the Python you're working on by using the ``apt`` command.
 
-      sudo nano /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -619,49 +619,55 @@ on Linux, macOS and iOS.
...
 
-   For other distributions, like Debian, change the URL and names to correspond
-   with the specific distribution.
+         deb-src http://archive.ubuntu.com/ubuntu/ jammy main
 
-   Then you should update the packages index::
+      Alternatively, uncomment lines with ``deb-src`` using an editor, e.g.::
 
-      $ sudo apt-get update
+         sudo nano /etc/apt/sources.list
```
erlend-aasland, line 650:
```suggestion
         $ sudo nano /etc/apt/sources.list
```

getting-started/setup-building.rst
```diff
@@ -619,49 +619,55 @@ on Linux, macOS and iOS.
...
 
-      sudo nano /etc/apt/sources.list
+      First, make sure you have enabled the source packages in the sources list.
+      You can do this by adding the location of the source packages, including
+      URL, distribution name and component name, to ``/etc/apt/sources.list``.
+      Take Ubuntu 22.04 LTS (Jammy Jellyfish) for example::
 
-   For other distributions, like Debian, change the URL and names to correspond
-   with the specific distribution.
+         deb-src http://archive.ubuntu.com/ubuntu/ jammy main
```
hugovk, line 646:
```suggestion
         $ deb-src http://archive.ubuntu.com/ubuntu/ jammy main
```

getting-started/setup-building.rst
```diff
@@ -619,49 +619,55 @@ on Linux, macOS and iOS.
    distribution, but the appropriate commands for some popular distributions
    are below.
 
-   On **Fedora**, **Red Hat Enterprise Linux** and other ``yum`` based systems::
+   .. tab:: ``yum``
 
-      $ sudo yum install yum-utils
-      $ sudo yum-builddep python3
+      On **Fedora**, **Red Hat Enterprise Linux** and other ``yum`` based systems::
```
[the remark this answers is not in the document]
encukou: I think this section could be dropped entirely. The last release of `yum` was 2011. The last `yum`-based Fedora went EoL around 2016 AFAIK. Some `yum`-based systems are still supported, but their users can generally install `dnf` or translate the instructions.

encukou:
I think we should not hide **Ubuntu** under “apt” -- there are people who don't know their system's package manager, and they're likely to be using Ubuntu :)

hugovk:
Sure, shall we remove `yum` and rename the other tabs "Fedora & RHEL" and "Debian & Ubuntu"?

hugovk:
And which tab should come first? Do we have an idea of which distros our readers use most?

Random website stats https://truelist.co/blog/linux-statistics/ say:

* apt: Ubuntu + Debian + CentOS = 49.9%
* DNF: CentOS + Red Hat + Fedora = 10.3%

Suggesting we should flip the order.

encukou:
If you remove `yum`, there are two left: one that's very short, and one that's more widely used.
IMO, the best presentation for that is regular sections rather than tabs, with the short one first so it's obvious that another section follows.

hugovk:
Updated!

getting-started/setup-building.rst
```diff
@@ -619,29 +619,25 @@ on Linux, macOS and iOS.
    distribution, but the appropriate commands for some popular distributions
    are below.
 
-   On **Fedora**, **Red Hat Enterprise Linux** and other ``yum`` based systems::
-
-      $ sudo yum install yum-utils
-      $ sudo yum-builddep python3
-
-   On **Fedora** and other ``DNF`` based systems::
+   On **Fedora** and other ``dnf``-based systems::
```
encukou, line 622:
```suggestion
   On **Fedora**, **RHEL**, **CentOS** and other ``dnf``-based systems::
```

encukou (approved):
Sometimes, simple is best :)

hugovk:
Yes, reverted the original change but we nevertheless end up with better docs :)

PR #1322, opened 2024-05-11
Group sphinx definitions

nineteendo:
Issue: https://github.com/python/cpython/issues/118030

----
📚 Documentation preview 📚: https://cpython-devguide--1322.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -432,7 +432,7 @@ following example shows all of the features of this directive type::
                   ham(eggs)
        :noindex:
 
-       Spam or ham the eggs.
+       Spam or ham the foo.
 
 The signatures of object methods or data attributes should not include the
 class name, but be nested in a class directive.  The generated files will
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -246,16 +246,17 @@ directives.)  Looking at this example,
 
 ::
 
-   .. function:: remove(path, *, dir_fd=None)
-                 unlink(path, *, dir_fd=None)
+   .. function:: foo(x)
+                 foo(y, z)
       :bar: no
 
-      Remove (delete) the file *path*.
+      Return a line of text input from the user.
 
 ``function`` is the directive name.  It is given two arguments here, the
 remainder of the first line and the second line, as well as one option ``bar``
-By providing multiple lines, the definitions will be grouped instead of
-separated by a empty line.
+By providing multiple lines, the signatures will be grouped instead of
+separated by a empty line. You can also use this for definitions with a
+different name.
 (as you can see, options are given in the lines immediately following the
 arguments and indicated by the colons).
 
@@ -431,7 +432,7 @@ following example shows all of the features of this directive type::
                   ham(eggs)
        :noindex:
 
-       Spam or ham the foo.
+       Spam or ham the eggs.
 
 The signatures of object methods or data attributes should not include the
 class name, but be nested in a class directive.  The generated files will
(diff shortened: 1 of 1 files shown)


nineteendo:
@hugovk, should we add it here?

hugovk:
For reference:

- https://github.com/python/cpython/issues/118030
- https://discuss.python.org/t/grouping-sphinx-object-definition-directives/52745

documentation/markup.rst
```diff
@@ -246,14 +246,16 @@ directives.)  Looking at this example,
 
 ::
 
-   .. function:: foo(x)
-                 foo(y, z)
+   .. function:: remove(path, *, dir_fd=None)
+                 unlink(path, *, dir_fd=None)
```
ezio-melotti, line 250: It seems there are at least two use cases for this:
1. showing different names/aliases for the same function (like you did)
2. showing multiple signatures for the same function (the original example)

I think the second one is much more common and might be a better example.  Just looking at the [built-in functions](https://docs.python.org/3/library/functions.html) there are several -- such as [`int`](https://docs.python.org/3/library/functions.html#int) or [`max`](https://docs.python.org/3/library/functions.html#max) -- that could be used if we wanted a real-world example.

documentation/markup.rst
```diff
@@ -246,14 +246,16 @@ directives.)  Looking at this example,
...
       :bar: no
```
ezio-melotti, line 251: Leaving this after updating the example looks a bit out of place though.  Are there any real-world examples that have options?
Otherwise why not just leave the original example unchanged?

documentation/markup.rst
```diff
@@ -246,14 +246,16 @@ directives.)  Looking at this example,
...
 
-      Return a line of text input from the user.
+      Remove (delete) the file *path*.
 
 ``function`` is the directive name.  It is given two arguments here, the
 remainder of the first line and the second line, as well as one option ``bar``
+By providing multiple arguments, the definitions will be grouped instead of
+separated by a empty line.
```
ezio-melotti, line 258: Here you technically provided multiple *names* with the same *arguments*.  The original example had the same *name* (twice) and multiple different *arguments*.
I'm not sure if there is a term that applies to both the *name* and the *argument* -- maybe *lines* is ok:
```suggestion
By providing multiple lines, the definitions will be grouped instead of
separated by a empty line.
```

documentation/markup.rst
nineteendo: Yeah, let's keep the orignal example.

documentation/markup.rst
nineteendo: I explained the second use in text only.

nineteendo:
Thanks Ezio.

nineteendo:
Can this be merged? It should happen before group other definitions in the docs.

documentation/markup.rst
```diff
@@ -429,7 +432,7 @@ following example shows all of the features of this directive type::
                   ham(eggs)
        :noindex:
 
-       Spam or ham the foo.
+       Spam or ham the eggs.
```
erlend-aasland, line 435: Please revert unrelated changes.

```suggestion
       Spam or ham the foo.
```

nineteendo: Alright, let's leave it then, it's too minor for another pull request.

nineteendo:
Do we merge this or not?

nineteendo:
Let's close this, it probably doesn't solve the initial issue.

PR #1327, opened 2024-05-27
Define port for htmllive to avoid collisions

hugovk:
It's not unusual that I might run `make htmllive` for a pair of CPython docs, devguide and PEPs at the same time.

To avoid clashes, use a randomly-chosen ["ephemeral port"](https://en.wikipedia.org/wiki/Ephemeral_port) between 49152–65535 for the local server for each of devguide and PEPs.

```python
python -c "import random; print(random.randint(49152, 65535))"
```

Also use the more descriptive Sphinx long options (added in [7.3.0](https://www.sphinx-doc.org/en/master/changes.html#release-7-3-0-released-apr-16-2024)) for the Makefiles. 

See also https://github.com/python/peps/pull/3792.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ htmlview: html
 
 .PHONY: htmllive
 htmllive: SPHINXBUILD = $(VENVDIR)/bin/sphinx-autobuild
-htmllive: SPHINXOPTS = --re-ignore="/\.idea/|/venv/" --open-browser --delay 0
+htmllive: SPHINXOPTS = --re-ignore="/\.idea/|/venv/" --open-browser --delay 0 --port 55311
 htmllive: html
 
 .PHONY: check
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 PYTHON        = python3
 VENVDIR       = ./venv
 SPHINXBUILD   = $(VENVDIR)/bin/sphinx-build
-SPHINXOPTS    = -W --keep-going
+SPHINXOPTS    = --fail-on-warning --keep-going
 BUILDDIR      = _build
 BUILDER       = html
 JOBS          = auto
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -170,7 +170,7 @@ htmlview: html
 
 .PHONY: htmllive
 htmllive: SPHINXBUILD = $(VENVDIR)/bin/sphinx-autobuild
-htmllive: SPHINXOPTS = --re-ignore="/\.idea/|/venv/" --open-browser --delay 0
+htmllive: SPHINXOPTS = --re-ignore="/\.idea/|/venv/" --open-browser --delay 0 --port 55311
```
ezio-melotti, line 173: I would add a comment to specify that you picked it randomly.

Are you going to pick random ones for the other repos too?
Would it be better to just pick a range and have e.g. 55300 for cpython, 55301 for the devguide, and so on?

hugovk: > I would add a comment to specify that you picked it randomly.

Will do.

> Are you going to pick random ones for the other repos too?

Yes, I added a random one for PEPs here: https://github.com/python/peps/pull/3792

I made sure it started with a 5 here, and a 6 there.

I wasn't going to add one for CPython, at least not yet. I figured that it should be fine with the default. Although it could clash with other default servers (e.g. `python -m http.server`), if it comes up with any regularity, we could use one beginning with 4.

> Would it be better to just pick a range and have e.g. 55300 for cpython, 55301 for the devguide, and so on?

I don't know, can change if you'd prefer.

ezio-melotti: The reason why I'm asking these questions, is that seeing an arbitrary number like 55311 makes me wonder if it's some specific port that was picked for some specific reason.

Seeing other random ports in other projects will make this even more puzzling.  Adding comments will solve this concern, but also having a range of consecutive numbers will help clarify implicitly that the numbers themselves are not special.

I would also change the last digit rather than first, for two reasons:
1. after you used 4, 5, and 6, you will have to start reusing one of those;
2. having consecutive values helps indicate that it's just an incrementing sequence of arbitrary ports;

hugovk: Sounds reasonable to me 👍 

Updated, how's this look?

Makefile
```diff
@@ -170,7 +170,8 @@ htmlview: html
 
 .PHONY: htmllive
 htmllive: SPHINXBUILD = $(VENVDIR)/bin/sphinx-autobuild
-htmllive: SPHINXOPTS = --re-ignore="/\.idea/|/venv/" --open-browser --delay 0
+# Arbitrarily selected ephemeral port between 49152–65535:
```
ezio-melotti, line 173:
```suggestion
# Arbitrarily selected ephemeral port between 49152–65535
# to avoid conflicts with other processes:
```
I would also state the problem being solved.

ezio-melotti (approved):

kiko1842 (approved):

kiko1842 (approved):

PR #1328, opened 2024-05-28
Prefer "feature release"

hugovk:
"Feature release/version" is the terminology used at https://devguide.python.org/developer-workflow/development-cycle/#devcycle and for example [PEP 602](https://peps.python.org/pep-0602/) (Annual Release Cycle for Python).

* Fix 'major release' -> 'feature release'
  These used major release when they really meant feature release.

* Prefer 'feature release' over 'minor release/version'
  Minor release is fine, but feature release is clearer.

----
📚 Documentation preview 📚: https://cpython-devguide--1328.org.readthedocs.build/

### DIFF
--- a/developer-workflow/c-api.rst
+++ b/developer-workflow/c-api.rst
@@ -174,7 +174,7 @@ Unstable C API
 
 The unstable C API tier is meant for extensions that need tight integration
 with the interpreter, like debuggers and JIT compilers.
-Users of this tier may need to change their code with every minor release.
+Users of this tier may need to change their code with every feature release.
 
 In many ways, this tier is like the general C API:
 
@@ -189,7 +189,7 @@ The differences are:
 
 - Names of functions structs, macros, etc. start with the ``PyUnstable_``
   prefix. This defines what's in the unstable tier.
-- The unstable API can change in minor versions, without any deprecation
+- The unstable API can change in feature releases, without any deprecation
   period.
 - A stability note appears in the docs.
   This happens automatically, based on the name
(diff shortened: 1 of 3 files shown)


erlend-aasland (approved):

ezio-melotti (approved):
This got me confused for a bit since both major and minor versions introduce features, and both were being converted to "feature release"  (I now realize that `major` was used incorrectly).

I see two problems that might need a solution here:
1. decide which set(s) of term we want to use;
2. ensure that the users understand what we mean.

The second problem can be solved by linking to the [`devcycle` section](https://devguide.python.org/developer-workflow/development-cycle/#devcycle) (or possibly a glossary entry).  This will ensure that the meaning of the terms we use is clear regardless of the term we decide to pick.
 
As for the terms themselves:
* The [`devcycle` section](https://devguide.python.org/developer-workflow/development-cycle/#devcycle) you linked talks about the "formal" nomenclature `major.minor.micro`.  This is also what is used by [`sys.version_info`](https://docs.python.org/3/library/sys.html#sys.version_info);
* The same section also offers and document another set of commonly-used and alternative terms: "minor versions are feature release" and "micro versions are bugfix releases".  It doesn't offer an alternative term for major versions and just says that "major versions are exceptional".  This might not be an issue though, since major releases are referred to rarely compared to feature/bugfix releases, so keep using `feature` and `bugfix` to colloquially refer to `minor` and `micro` releases might be ok.
* The official terminology used by [semver](https://semver.org/#summary) is `major.minor.patch`.  Using the semver terminology would ideally be better (since most users should be familiar with it). However we have been using `micro` instead of `patch`, and starting to use `patch` while still having `micro` in `sys.versions_info` and elsewhere will create additional confusion.  


In summary:
* I would fix the two misuses of `major`, since they are wrong.  In the `c-api` page I think both `minor` and `feature` are ok.
* I would keep the status quo of having both the "formal" `major.minor.micro` and "colloquial" alternatives `feature` and `bugfix` for `minor`/`micro`, since they are all commonly used terms within our docs.
* I would add links to the `devcycle` section when these terms are used to ensure the users understand what we are talking about (this can be done separately).

erlend-aasland:
We want to avoid the connection with SemVer (since we're not adhering to it). With that in mind, it makes sense to denote a new _minor_ release as a feature release. For a SemVer project, a feature release would mean a new _major_ release, but we all know that there will never be a new _major_ release of Python; Python 4 is not happening.

PR #1337, opened 2024-06-07
Put 'Install dependencies' before 'Compile and build'

erlend-aasland:
----
📚 Documentation preview 📚: https://cpython-devguide--1337.org.readthedocs.build/getting-started/setup-building/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -334,7 +334,6 @@ on Linux, macOS and iOS.
    CPython. If you use the pre-compiled binaries, you should unpack each tarball
    into a separate folder, and use that folder as the configuration target.
 
-
 .. _compiling:
 
 Compile and build
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -143,6 +143,198 @@ To make sure your code is linted correctly, we recommend setting up
 
 Now pre-commit will run automatically on ``git commit``.
 
+.. _build-dependencies:
+.. _deps-on-linux:
+.. _macOS and OS X:
+.. _macOS:
+
+Install dependencies
+====================
+
+This section explains how to install additional extensions (e.g. ``zlib``)
+on Linux, macOS and iOS.
+
+.. tab:: Linux
+
+   For Unix-based systems, we try to use system libraries whenever available.
+   This means optional components will only build if the relevant system headers
+   are available. The best way to obtain the appropriate headers will vary by
+   distribution, but the appropriate commands for some popular distributions
+   are below.
+
+   On **Fedora**, **RHEL**, **CentOS** and other ``dnf``-based systems::
+
+      $ sudo dnf install dnf-plugins-core  # install this to use 'dnf builddep'
+      $ sudo dnf builddep python3
+
+
+   On **Debian**, **Ubuntu**, and other ``apt``-based systems, try to get the
+   dependencies for the Python you're working on by using the ``apt`` command.
+
+   First, make sure you have enabled the source packages in the sources list.
+   You can do this by adding the location of the source packages, including
+   URL, distribution name and component name, to ``/etc/apt/sources.list``.
+   Take Ubuntu 22.04 LTS (Jammy Jellyfish) for example::
+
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -143,6 +143,197 @@ To make sure your code is linted correctly, we recommend setting up
...
+.. _build-dependencies:
+.. _deps-on-linux:
+.. _macOS and OS X:
+.. _macOS:
+
+Install dependencies
+====================
+
+This section explains how to install additional extensions (e.g. ``zlib``)
+on Linux, macOS and iOS.
```
nedbat, line 155: Originally I guess this was later because many people don't need those extra modules.  Perhaps we could add a statement here like, "these dependencies are needed to build the complete standard library, but many people don't need those modules so it might be fine to skip this section." (but worded better).

hugovk:
Also update "below" -> "above" in "Compile and build":
```diff
-If you want to install these optional dependencies, consult the Install dependencies section below.
+If you want to install these optional dependencies, consult the Install dependencies section above.
```

getting-started/setup-building.rst
erlend-aasland: Yeah, I agree, but I would prefer to do that in a follow-up PR.

erlend-aasland:
> Also update "below" -> "above" in "Compile and build":

Thanks; addressed in f23f169c0f22e9a0beef7103a58c0e0c07d57b18.

getting-started/setup-building.rst
ezio-melotti: I think it might even be better to just leave it afterward, but possibly add a note here with a short explanation and link to it.

IME, unless you are working on some specific module that require some of those extensions, you can just ignore this whole section.

Also the section currently explains how to install dependencies, but doesn't cover why/when you might want to do that.

ezio-melotti: (On a slightly unrelated note, that page seems a bit daunting, especially when -- at least on Linux -- you generally just need 3 commands: `git clone`, `./configure`, and `make`.  Maybe we should add a quick start at the top that just lists these commands and links to the various other sections for more details.)

hugovk (approved):

getting-started/setup-building.rst
erlend-aasland: From time to time we get "bug reports" from people who complain about extension modules that fail to build because of missing dependencies. This also happen on the Help forum on Discourse. We _might_ get fewer reports like that with this change.

I do like the suggestion about adding a "tl;dr" where we spell out the clone-configure-make recipe on the top of the page.

ezio-melotti: > From time to time we get "bug reports" from people who complain about extension modules that fail to build because of missing dependencies.

I remember having this discussion already a while ago, and IIRC the wording was changed to clarify that they are not blocking failures (or maybe it was just proposed and never implemented?).

erlend-aasland: Examples of bug reports bco. missing dependencies:

- https://github.com/python/cpython/issues/116224
- https://github.com/python/cpython/issues/108856
- https://github.com/python/cpython/issues/105706

ezio-melotti: Thanks for the examples!  ISTM that:
* Some users don't need the extra deps, and would benefit from a simpler process (i.e. cover `configure`/`make` first, and the installation of extra deps later).  This group should include most contributors, which are the target audience for the devguide;
* Some users need the extra deps, and for them having the section about installing them first would be better.  This group includes people that are manually installing Python to use it, and want all the extra bits.

Regardless of their intentions:
* IMHO adding a note at the beginning of the "Compile and build" section that links to the "Installing dependencies" section (with a short explanation about when/why you might want to do that) instead of moving the "Installing dependencies" at the top is a good enough compromise;
* The warning message about missing dependencies could be clarified further, and could possibly include a link to the devguide;
* The whole "setup and building" page could be streamlined a bit more with a simple list of steps that link to the various sections;

erlend-aasland:
I opened an issue for this change; let's discuss how we want to solve this before continuing with a PR.

PR #1343, opened 2024-07-02
Remove e.g. and i.e. throughout

nedbat:
I'm not sure if this is controversial or not.  It's ready to merge if people like it.

----
📚 Documentation preview 📚: https://cpython-devguide--1343.org.readthedocs.build/


erlend-aasland (approved):
Thanks; I like this very much!

documentation/style-guide.rst
```diff
@@ -88,6 +88,15 @@ Unix
    1970s.
 
 
+Use simple language
+===================
+
+Avoid esoteric phrasing where possible.  Our audience is world-wide and may not
+be native English speakers.
+
+Don't use Latin abbreviations like "e.g." or "i.e." where English words will do.
```
hugovk, line 97: ... such as "for example" and "that is" ?

getting-started/pull-request-lifecycle.rst
```diff
@@ -42,7 +42,7 @@ Here is a quick overview of how you can contribute to CPython:
 #. `Create Pull Request`_ on GitHub to merge a branch from your fork
 
 #. Make sure the :ref:`continuous integration checks on your Pull Request
-   are green <keeping-ci-green>` (i.e. successful)
+   are green <keeping-ci-green>` (that is, successful)
```
hugovk, line 45: This could be omitted (but okay to ignore this):
```suggestion
   are green <keeping-ci-green>` (successful)
```

hugovk (approved):

documentation/style-guide.rst
nedbat: done

getting-started/pull-request-lifecycle.rst
nedbat: done

ezio-melotti:
Is there any background for this decision?
I did a quick search and there are some article about using `e.g.` instead of `eg` or `eg.`, but didn't find style guides that recommend against using them.

hugovk:
Here's some style guides recommending against Latin abbreviations:

* Google: https://developers.google.com/style/abbreviations
* Microsoft: https://learn.microsoft.com/en-us/style-guide/word-choice/use-us-spelling-avoid-non-english-words
* Red Hat: https://stylepedia.net/style/6.2/#contractions-and-abbreviations
* Apple: https://support.apple.com/en-gb/guide/applestyleguide/apsg3acde405/1.0/web/1.0#apde9a26a57d4df4
* Intuit: https://contentdesign.intuit.com/word-list/#i-e
* Salesforce: https://developer.salesforce.com/docs/atlas.en-us.salesforce_pubs_style_guide.meta/salesforce_pubs_style_guide/style_ie.htm
* edX: https://draft-edx-style-guide.readthedocs.io/en/latest/global_English.html#abbreviations
* GOV.UK: https://insidegovuk.blog.gov.uk/2016/07/20/changes-to-the-style-guide-no-more-eg-and-ie-etc/ https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#eg-etc-and-ie
* Australian Government Style Manual: https://www.stylemanual.gov.au/grammar-punctuation-and-conventions/shortened-words-and-phrases/latin-shortened-forms
* "Most style manuals (APA, MLA, and Chicago) suggest limiting the use of Latin abbreviations in the main text": https://paperpal.com/blog/academic-writing-guides/language-grammar/abbreviations-in-research

ezio-melotti (approved):
From those links:
* Google: 
  > Prefer English terms over Latin abbreviations. Don't use i.e. or e.g.; instead, use that is or for example, respectively. One exception: it's [okay to use etc. in some circumstances](https://developers.google.com/style/word-list#etc).

  > Avoid using etc., and so forth, and and so on wherever possible. If you really need to use one, use etc. Always include the period, even if a comma follows immediately after. 
* Microsoft
  > Avoid Latin abbreviations for common English phrases.
  > Exception It’s OK to use etc., in situations where space is limited. Otherwise, see and so on for alternatives.
* Red Hat:
  > Take care also with abbreviations; replace "e.g." with "for example", and replace "i.e." with "that is", and so on. 
* Apple:
  > Latin: Avoid using Latin abbreviations.
  >   Correct: for example, and others, and so on, and that is, or equivalent phrases
  >   Incorrect: e.g. (for example), et al. (and others), etc. (and so on), i.e. (that is)

I personally like to use those abbreviations but the style guides disagree with me, and I agree with the style guides.

PR #1345, opened 2024-07-02
update python organization owners

ewdurbin:
- @JacobCoffee has started as PSF Infrastructure Engineer
- Van is no longer acting as counsel to the PSF and owner role was removed some time ago

----
📚 Documentation preview 📚: https://cpython-devguide--1345.org.readthedocs.build/


JacobCoffee:
The team will also need updating to reflect that change in https://github.com/orgs/python/teams/organization-owners

ewdurbin:
> The team will also need updating to reflect that change in https://github.com/orgs/python/teams/organization-owners

Good call! Done!

PR #1349, opened 2024-07-13
Use latest version of OpenSSL

redorlik:
The dev guide pins the version of OpenSSL to version 3.0, but the current latest version i 3.3.1 (July 2024).

This change pins the OpenSSL version to 3.

----
📚 Documentation preview 📚: https://cpython-devguide--1349.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -691,7 +691,7 @@ on Linux, macOS and iOS.
 
       For **Homebrew**, install dependencies using ``brew``::
 
-         $ brew install pkg-config openssl@3.0 xz gdbm tcl-tk mpdecimal
+         $ brew install pkg-config openssl@3 xz gdbm tcl-tk mpdecimal
 
       .. tab:: Python 3.13+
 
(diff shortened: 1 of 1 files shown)


ezio-melotti (commented):
OpenSSL was updated from 1.1 to 3.0 in:
* #1111
* #1112

The changes LGTM unless there is some reason to prefer pinning 3.0.

ned-deily:
> The changes LGTM unless there is some reason to prefer pinning 3.0.

IIRC, when we finally switched from recommending OpenSSL 1.x to OpenSSL 3.0.x, the thinking was that 3.0.x was the OpenSSL [stable Long Term Support (LTS) branch](https://www.openssl.org/policies/releasestrat.html) with the longest support life and that is still the case. Whether that should be the deciding factor today isn't clear to me. Other opinions? @gpshead ? @tiran ?

gpshead:
Yep, so long as OpenSSL project is providing the concept of "long term support" releases as you linked to above, we should stick with those in our binary release builds shipped by CPython. We want to ship builds made with the OpenSSL version having the latest EOL date to minimize potentially disruptive changes in Python patch releases.

ned-deily:
> Just be aware that when a new OpenSSL release comes out, not all branches of CPython may be ready to build and link against it yet (including `main`). If we accept this PR and that hurts us in the future we could revisit this change.

I agree with the above though the net effect of the change in the PR is that upgrades from one branch of OpenSSL to another will happen somewhat unpredictably as it will depend on when Homebrew updates their `openssl3` recipe to a different OpenSSL 3.x branch and then when the user of the devguide suggestion updates their installation of Homebrew.

ned-deily (approved):

PR #1350, opened 2024-07-13
Explain how to add an extension module

### LINKED ISSUE #317: Add a section on how to add an extension module to the stdlib.
Every once in a while someone adds a new extension module to the stdlib.  How to do so isn't very clear, particularly regarding all the cross-platform bits.  I suppose nobody has added that info to the devguide because it doesn't happen very often; however, it would be particularly helpful in the devguide specifically because we don't do it very often. :)  You can mostly look for prior art (e.g. search for what was done when the last module was added), but that's not ideal and it's easy to miss stuff.

I'd expect a new (short) section, probably in [stdlibchanges.rst](/python/devguide/blob/master/stdlibchanges.rst), with a brief summary (including expectations) and a step-by-step for supported platforms.  Each platform would have it's own subsection with the steps as well as any caveats.  For myself, the Windows section would be where I'd get the most value.  However, I imagine that Windows folks could say the same for non-Windows (trivial as it might seem to *me*).

picnixz:
This is a first proposal for explaining how to add an extension module. Feel free to correct me, tell me that it's too verbose or some places need more information!

Fixes #317 

----
📚 Documentation preview 📚: https://cpython-devguide--1350.org.readthedocs.build/developer-workflow/extension-modules/


developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
 Standard library extension modules
 ==================================
 
-In this section, we could explain how to write a CPython extension with the C language, but the topic can take a complete book.
+In this section, we could explain how to write a CPython extension with the
+C language, but the topic can take a complete book. We will however explain
+how to add a new extension module to the standard library, e.g., a module
```
hugovk, line 10: Please avoid Latin abbreviations: https://devguide.python.org/documentation/style-guide/#use-simple-language

```suggestion
how to add a new extension module to the standard library, for example, a module
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+           return "Hello World!"
+
+Some modules in the standard library are implemented both in C and in Python,
+such as :mod:`decimal` or :mod:`io`, and the C implementation is expected
+to improve performance when available (such modules are usually referred to as
+*accelerator modules*). In our example, we need to:
+
+- determine where the extension module is to be placed;
+- determine which files to modify in order to compile the project;
+- determine which Makefile rules to invoke in the end.
```
hugovk, line 58:
```suggestion
- determine where to place the extension module;
- determine which files to modify in order to compile the project;
- determine which ``Makefile`` rules to invoke at the end.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+   #include "foomodule.h"
+
+   PyObject *_Py_fast_bar() {
+       return PyUnicode_FromString("Hello World!");
+   }
+
+.. tip::
+
+   Do not forget that symbols exported by libpython must start
+   with "Py" or "_Py", which is verified via ``make smelly``.
```
hugovk, line 208:
```suggestion
   Do not forget that symbols exported by ``libpython`` must start
   with ``Py`` or ``_Py``, which is verified via ``make smelly``.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+Compile the CPython project
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Now that everything is in place, it remains to compile the project:
+
+.. code-block:: shell
+
+   $ make regen-configure
+   $ make regen-all
+   $ make regen-stdlib-module-names
```
hugovk, line 303: The copy button omits the dollar signs, but not when highlighting and manually copying.

```suggestion
   make regen-configure
   make regen-all
   make regen-stdlib-module-names
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+- The ``regen-stdlib-module-names`` updates the standard module names,
+  making ``_foo`` discoverable and importable via ``import _foo``.
+
+You can now compile the entire project by running the following commands:
+
+.. code-block:: shell
+
+   $ ./configure --with-pydebug
+   $ make
```
hugovk, line 319:
```suggestion
   ./configure --with-pydebug
   make
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+.. tip:: Use ``make -j12`` to speed-up the compilation if you have enough CPU cores.
```
hugovk, line 321:
```suggestion
.. tip:: Use ``make -j12`` to speed-up compilation if you have enough CPU cores.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+Troubleshooting
+^^^^^^^^^^^^^^^
+
+This section addresses common issues that developers may face when following this tutorial.
```
hugovk, line 326:
```suggestion
This section addresses common issues you may face when following this tutorial.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+``make regen-configure`` does not work!
+.......................................
+
+Since this rule requires Docker to be running and a Docker instance,
+the following can be done on Linux platforms (systemctl-based):
```
hugovk, line 332:
```suggestion
the following can be done on Linux platforms (``systemctl``-based):
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+.. code-block:: shell
+
+   $ systemctl status docker        # is the docker service running?
+   $ sudo systemctl start docker    # start it if not!
+   $ sudo systemctl restart docker  # or restart it!
```
hugovk, line 338:
```suggestion
   systemctl status docker        # is the docker service running?
   sudo systemctl start docker    # start it if not!
   sudo systemctl restart docker  # or restart it!
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+
+If Docker complains about missing permissions, the following StackOverflow post
```
hugovk, line 340:
```suggestion
If Docker complains about missing permissions, this Stack Overflow post
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,338 @@
...
+In this section, we could explain how to write a CPython extension with the
+C language, but the topic can take a complete book. We will however explain
+how to add a new extension module to the standard library, e.g., a module
+responsible for accelerating some parts of the library.
 
-For this reason, we prefer to give you some links where you can read a very good documentation.
+For writing a CPython extension itself, we prefer to give you some links
+where you can read a very good documentation.
 
 Read the following references:
```
hugovk, line 16:
```suggestion
For writing a CPython extension itself, we prefer to give you some links
where you can read good documentation:
```

developer-workflow/extension-modules.rst
picnixz: I plead guilty. I didn't read the guidelines...

picnixz:
Thank you Hugo! do you have comments on the content itself, or parts that were not clear enough?

hugovk:
> Thank you Hugo! do you have comments on the content itself, or parts that were not clear enough?

Content looks okay to me, but let's wait for someone who is more familiar with adding extension modules to CPython to also review.

By the way, we don't need to add `gh-` prefixes here to PR titles, they're only needed on the CPython for Bedevere to link issues to PRs and backports.

picnixz:
Yes, that's what I observed afterwards. I thought that *maybe* there would be Bedevere that would update the issue automatically but in the end I needed to add the "Fix #..." myself.

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
 Standard library extension modules
 ==================================
```
ezio-melotti, line 6: Titles should follow this guidelines: https://devguide.python.org/documentation/markup/#sections

```suggestion
==================================
Standard library extension modules
==================================
```

In addition, depending on the differences between "CPython extensions" and "extensions module" (see my other comment), you might want to have a section for each, with the one for "CPython extensions" only having a short paragraph and a list of links.  If there is no meaningful difference, one of the top-levels of the page could be removed.  For reference, this is the current structure of the page:
```
Standard library extension modules
    Adding an extension module to CPython
        Make the CPython project compile
        Compile the CPython project
        Troubleshooting
            make regen-configure does not work!
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
...
+For writing a CPython extension itself, we prefer to give you some links
+where you can read good documentation:
 
 * https://docs.python.org/dev/c-api/
 * https://docs.python.org/dev/extending/
 * :pep:`399`
 * https://pythonextensionpatterns.readthedocs.io/en/latest/
+
+Adding an extension module to CPython
+-------------------------------------
```
ezio-melotti, line 22:
```suggestion
Adding an extension module to CPython
=====================================
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
...
+
+   Do not forget that symbols exported by ``libpython`` must start
+   with ``Py`` or ``_Py``, which is verified via ``make smelly``.
+
+One could imagine having more ``.h`` files, or no ``helper.c`` file if it is
+not needed. Here, we wanted to illustrate a simple example without making it
+too trivial.
+
+Make the CPython project compile
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ezio-melotti, line 213:
```suggestion
Make the CPython project compile
--------------------------------
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
...
+     </ClCompile>
+     <ClCompile Include="..\Modules\foo\helper.c">
+       <Filter>Modules\foo</Filter>
+     </ClCompile>
+
+Observe that ``.h`` files use ``<ClInclude ...>`` whereas ``.c`` files
+use ``<ClCompile ...>`` tags.
+
+Compile the CPython project
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ezio-melotti, line 293:
```suggestion
Compile the CPython project
---------------------------
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
...
+
+.. code-block:: shell
+
+   ./configure --with-pydebug
+   make
+
+.. tip:: Use ``make -j12`` to speed-up compilation if you have enough CPU cores.
+
+Troubleshooting
+^^^^^^^^^^^^^^^
```
ezio-melotti, line 322:
```suggestion
Troubleshooting
---------------
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
...
+
+This section addresses common issues that you may face when following this tutorial.
+
+``make regen-configure`` does not work!
+.......................................
```
ezio-melotti, line 327:
```suggestion
``make regen-configure`` does not work!
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,336 @@
 Standard library extension modules
 ==================================
 
-In this section, we could explain how to write a CPython extension with the C language, but the topic can take a complete book.
+In this section, we could explain how to write a CPython extension with the
+C language, but the topic can take a complete book. We will however explain
+how to add a new extension module to the standard library, for instance, a
+module responsible for accelerating some parts of the library.
```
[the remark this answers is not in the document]
picnixz: The current sentence is "In this section, we could explain how to write a CPython extension with the C language, but the topic can take a complete book." so I assumed that "CPython extension" designated whatever extension you added to Python (like a plugin) whether it's in C or C++, internally (committed to main) or externally (linked against `libpython` and discoverable if you provide the .so/.dll files).

The name of the page is "Standard library extension modules" so I used the terminology "module" to be consistent and to highlight that the components are in a "module" file/folderr placed in `Modules/`. 

I just made a quick search and didn't find a lot of occurrences of "CPython extensions" but a lot of "extension module". So maybe it's better to remove "CPython extension" and just use the term "extension module" (because there are also C++ extension modules).

developer-workflow/extension-modules.rst
picnixz: I should have read the style guide beforehand... 

>  If there is no meaningful difference, one of the top-levels of the page could be removed.

For me, there is none. Unless you want to make the distinction between "CPython extension" for the extensions part of the CPython project only.

Ideally, I'd like to have the following sections:

```text
Standard library extension modules
    Adding an extension module to CPython
        Make the CPython project compile
        Compile the CPython project
        Troubleshooting
            make regen-configure does not work!

    Adding a frozen module to CPython
        Make the CPython project compile
        Compile the CPython project
        Troubleshooting

	Adding a configuration value to CPython (like Py_HAVE_C_COMPLEX)
``` 

The assumptions on frozen modules are distinct and I also wanted to add something about how to add configuration values. For me "Standard library extension modules" referred to the CPython project itself and how to extend the C implementation itself (whatever is in Modules/ (maybe Objects/ and Include/) but not in Lib/).

The section "Adding to the stdlib" only focuses on the process but not how to make the project work after (like where to propose things, etc but not the technical aspects). There is also a separte page for the Python C API but I think it's also something separate (I mean, you cannot change the C API with new files if you don't know how to...).

picnixz: Btw, the title already has '===' above. It's just that GitHub didn't show it in the diff =/

picnixz:
I was reading what I wrote and I forgot something about the bootstrapping so I'll update the tutorial

developer-workflow/extension-modules.rst
ezio-melotti: > Btw, the title already has '===' above. It's just that GitHub didn't show it in the diff =/

My bad, I didn't notice there were a few more lines above.  I removed the suggestion now.

Also no worries about the style guide -- that's what reviewers are for :)

Regarding the distinction between "CPython extensions" and "extension modules", we should hear from someone that is more familiar with those (@ncoghlan, @matrixise?).  The rest of your plan sounds good to me.

picnixz: Actually, by digging into the issue I mentioned (missing the configure/bootstrap) I am now confused about the terminology. I *think* the accelerator modules are actually referred to as *built-in modules* and some other modules such as blake2 are *extension* modules.

Also, it's still unclear to me which files must be manually edited and which ones need to be autogenerated. AFAIK, I managed to make my PR https://github.com/python/cpython/pull/121446 work but I'm pretty sure that this is perhaps not the way to go :') So I would be very happy if someone could tell me the *exact* steps to do (the solution I propose should work but it's probably not the best one?)

picnixz: Update: frozen modules are actually written in Python, so they wouldn't be part of this tutorial although incoporating them into CPython is done at the C level.

picnixz:
Ok, I think I understood what should be done. However, there are a few questions that I want to ask in general:
	
- what is the policy for choosing between a built-in module (statically linked and in `Setup.bootstrap.in`) and a non-built-in one (in `Setup.stdlib.in`)? and what is the policy on documenting something that should be shared? (I mean, it's fine to say that a module can be a shared library but... if you don't know what to implement to provide your own .so, it's better to keep it statically linked)
- should I really hardcode the modules to export in PC/config.c?

erlend-aasland:
> * what is the policy for choosing between a built-in module (statically linked and in `Setup.bootstrap.in`) and a non-built-in one (in `Setup.stdlib.in`)? and what is the policy on documenting something that should be shared? (I mean, it's fine to say that a module can be a shared library but... if you don't know what to implement to provide your own .so, it's better to keep it statically linked)

I don't think there's a formal rule, but IIRC modules that are needed for bootstrapping Python are built-in modules; the rest are not. So, IMO, I think the default should be that a new extension module should be optional. Perhaps this is worth a short mention in the introduction of this section?

IOW recommend using `Setup.stdlib.in`.

Moreover, I think you need to explain both `PY_STDLIB_MOD` and `PY_STDLIB_MOD_SIMPLE`, and when to use the former.

Also, I find the `cfoo` a bit strange. We've been using `_foo` for C modules (AFAICS, there are 76 of them). OTOH, the only `c` prefix I can find is `cmathmodule.c`.

picnixz:
> but IIRC modules that are needed for bootstrapping Python are built-in modules; the rest are not

Yes, that's what I thought.

> So, IMO, I think the default should be that a new extension module should be optional

That's what I think as well (in general, it's easier to have a statically linked module since you don't have less linker issues, at least from my experience).

> Perhaps this is worth a short mention in the introduction of this section?

More than worth I'd say :')

> Moreover, I think you need to explain both `PY_STDLIB_MOD` and `PY_STDLIB_MOD_SIMPLE`, and when to use the former.

Heheh... I think I myself used it incorrectly in my other PR :') Yes, I definitely should update this one. Correct me if I'm wrong but `PY_STDLIB_MOD` should be used if the module must be disabled or cannot be supported depending on some external conditions and `PY_STDLIB_MOD_SIMPLE` guarantees that the module is always present right (and cannot be absent due to some configuration, except `PY_STDLIB_MOD_SET_NA`)? I couldn't find more information, so this is something I could assume.

> Also, I find the cfoo a bit strange. We've been using `_foo` for C implementation of stdlib modules (AFAICS, there are 76 of them). OTOH, the only c prefix I can find is cmathmodule.c.

Actually, I wanted to use a name that is not prefixed by an underscore, just for visual differences. But maybe, it's indeed better to have `_foo` (note that the module was named "fastfoo" just to distinguish it from a possible directory name, because you might be confused which name to use in the template files).

encukou:
> it's easier to have a statically linked module since you don't have less linker issues

That choice is for the person who configures and builds Python, except for the few modules that *must* be linked statically (for the reasons listed in `Setup.bootstrap.in`) or that must be shared (see the very end of `Setup.stdlib.in`).

Where one can choose, the default is shared modules. I don't know the reasoning behind that, but it's good that they get as much testing as possible :)

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,489 @@
...
+       from fastfoo import greet
+   except ImportError:
+       # fallback to the pure Python implementation
+       def greet():
+           return "Hello World!"
+
+Some modules in the standard library are implemented both in C and in Python,
+such as :mod:`decimal` or :mod:`itertools`, and the C implementation is expected
+to improve performance when available (such modules are commonly referred
+to as *accelerator modules*). In our example, we need to determine:
```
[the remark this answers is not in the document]
picnixz: Oh thanks for the PEP tip, I wasn't aware of it! I think the introduction is the correct place to talk about them in their generality.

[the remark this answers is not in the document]
picnixz: Update: actually, I said that `itertools` was both in C and Python but there is no Python implementation of itertools... Now, I use `datetime` and `pickle` as examples since they really have identical Python implementation.

picnixz:
I want to talk about configuration variables in the configure script (those are usually important when adding accelerator modules, especially when we have different linkage and/or compiler features) but I don't want to make the PR too large, so I've added a small comment.

By the way, thank you everyone for your reviews; it helps a lot since different insights are valuable in this case!

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,571 @@
...
+
+According to :pep:`399`, *new* extension modules MUST provide a working and
+tested pure Python implementation, unless a special dispensation is given.
+Please ask the :github:`Steering Council <python/steering-council>` if such
+dispensation is needed.
+
+While new built-in extension modules could be considered exceptions to that
+rule, shared extension modules must provide a Python implementation, not only
+because of the guidelines, but also as a fallback if the corresponding shared
+library is not be present on the system.
```
encukou, line 63: I'd remove this paragraph, and leave the reasoning to the PEP.
The PEP says:
> Typically the exemption will be granted only when a module wraps a specific C-based library (e.g., [sqlite3](http://docs.python.org/py3k/library/sqlite3.html)).

That still holds. If `sqlite3` was being considered today, it probably wouldn't be added to stdlib at all, but if it would, it would be a shared extension without a Python implementation.

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,571 @@
...
+---------------
+
+This section addresses common issues that you may face when following this tutorial.
+
+No rule to make target ``regen-configure``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This usually happens after running ``make distclean`` (which removes
+the ``Makefile``). The solution is to regenerate the :cpy-file:`configure`
+script as follows:
```
encukou, line 545: Or run `Tools/build/regen-configure.sh` directly.

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,571 @@
...
+
+   ./configure            # for creating a Makefile
+   make regen-configure   # for updating 'configure'
+   ./configure            # for updating the Makefile
+
+``make regen-configure`` and missing permissions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Since this rule requires `Docker <https://docs.docker.com/desktop>`_ to be
+running, the following can be done on Linux platforms (``systemctl``-based):
```
encukou, line 557: Not sure if it's worth mentioning that instead of Docker you can use `podman`, which can be just installed, it doesn't need to be running.

encukou (commented):
Thank *you* for gathering the scattered tribal knowledge in one place!

One other thing that could be added to the examples is starting out with the [limited API](https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface): `#define` `Py_LIMITED_API` to `0x030d0000` (or the current `Py_HEX_VERSION`) before including `Python.h`.

It's fine to remove that whenever you need some API that's not in the limited set.

developer-workflow/extension-modules.rst
picnixz: I thought about that, but I wanted to reduce the number of tools or rules I introduce. Also, if by any chance the Makefile is doing something else, at least, I wouldn't need to update that section. But I can mention it if you think it's better.

developer-workflow/extension-modules.rst
picnixz: Well... I have no idea how to help people with podman :') but yes, I'll mention it as well. I'll have a look at it afterwards

picnixz:
> Thank you for gathering the scattered tribal knowledge in one place!

Well.. since I needed to learn it anyway, it was natural to document what I learned so that other people don't need a trial-and-error to make the CI/CD happy :D 

> One other thing that could be added to the examples is starting out with the [limited API](https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface)

I don't know a lot about the limited API and I think the tutorial should not necessarily have too much things as an introduction. How about some additional tips which talk about those? There was also something about `Py_BUILD_CORE_BUILTIN/Py_BUILD_CORE_MODULE` defines in `pyport.h`. Should I talk about them?

ncoghlan:
> OTOH, the only `c` prefix I can find is `cmathmodule.c`.

And that `c` stands for "complex", not "implemented in C". Best to continue the underscore-prefixed convention (new C modules will typically be accelerators for pure Python modules anyway).

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+
+   make regen-configure
+   ./configure
+   make regen-all
+   make regen-stdlib-module-names
+   make
+
+.. tip::
+
+   Use ``make -j12`` to speed-up compilation if you have enough CPU cores.
```
picnixz, line 515: Quick question but should I say ``make -j12``, ``make -j`` or ``make -j4``? There is an opened PR about this https://github.com/python/devguide/pull/1087 but I put 12 because I don't like to wait (also because I have 32 in total) but should I go for something like `-j8` instead (or even a more conservative `-j4`?)

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+   }
+
+.. tip::
+
+   Functions or data that do not need to be shared across different C source
+   files should be declared ``static`` to avoid exporting their symbols from
+   ``libpython``.
+
+   If symbols need to be exported, their names must start with ``Py`` or
+   ``_Py``. This can be verified by ``make smelly``.
```
ncoghlan, line 256: When discussing exporting symbols, reference https://devguide.python.org/developer-workflow/c-api/

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+     :caption: :cpy-file:`configure.ac`
+     :emphasize-lines: 4
+
+     dnl Modules that are not available on some platforms
+     AS_CASE([$ac_sys_system],
+         ...
+         [PLATFORM_NAME], [PY_STDLIB_MOD_SET_NA([fastfoo])],
+         ...
+     )
+
```
ncoghlan, line 354: Likely worth stating explicitly that folks should look at the comments and existing modules in `configure.ac` for guidance on adding new external build dependencies for modules that need them.

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+If needed, add the following line to the section for module dependencies:
+
+.. code-block:: text
+   :caption: :cpy-file:`Makefile.pre.in`
+   :emphasize-lines: 4
+
+   ##########################################################################
+   # Module dependencies and platform-specific files
+   ...
+   MODULE_FASTFOO_DEPS=$(srcdir)/Modules/_foo/_foomodule.h
```
ncoghlan, line 367: As per above comment `_foomodule.h` -> `helper.h`

ncoghlan (commented):
This is mostly looking good, just some comments inline about header files and naming conventions.

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+Updating the CPython project tree
+---------------------------------
+
+Usually, accelerator modules are added in the :cpy-file:`Modules` directory of
+the CPython project. If more than one file is needed for the extension module,
+it is more convenient to create a sub-directory in :cpy-file:`Modules`.
+
+For our extension module ``fastfoo``, we consider the following working tree:
+
+- :ref:`Modules/_foo/_foomodule.h` --- the extension module shared prototypes.
```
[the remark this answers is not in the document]
picnixz: I see. Because most of the modules that are not just a single .c files have a corresponding .h which has a bit of everything inside (almost all other modules have a single HUGE .c which makes it very hard to read in general). 

I'll change this one then!

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+- :ref:`Modules/_foo/_foomodule.c` --- the extension module implementation.
+- :ref:`Modules/_foo/helper.c` --- the extension helpers implementation.
+
+We deliberately named the working tree directory and files with names distinct
+from the actual Python module to import (whether it is the pure Python module
+or its C implementation) to highlight the differences in configuration files.
```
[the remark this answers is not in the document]
picnixz: Here is what I suggest:

- I use different names for the directory tree and the C module (namely, `_foo` and `_fastfoo`).
- But I say that the convention is to prefix the Python module with `_` if it's an accelerator module and use the same name for the directory tree and for the arguments clinic input. Would it be fine like this?

Or should I just harmonize everything as if it were production code and don't tell people that they could actually change the names ? :D

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,579 @@
...
+
+.. code-block:: python
+   :caption: Lib/foo.py
+
+   def greet():
+       return "Hello World!"
+
+Instead of using the Python implementation of :func:`!foo.greet`, we want to
+use its corresponding C implementation exposed in some :mod:`!fastfoo` module
+written in C. Ideally, we want to modify ``foo.py`` as follows:
```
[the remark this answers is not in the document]
picnixz: Ah! I have a question on this one:

- in the implementation of the io module, we have 

```c
PyMODINIT_FUNC
PyInit__io(void)
{
    return PyModuleDef_Init(&_PyIO_Module);
}
```

However:

```c
struct PyModuleDef _PyIO_Module = {
    .m_base = PyModuleDef_HEAD_INIT,
    .m_name = "io",  // <-- it's "io" but not "_io"
    .m_doc = module_doc,
    .m_size = sizeof(_PyIO_State),
    .m_methods = module_methods,
    .m_traverse = iomodule_traverse,
    .m_clear = iomodule_clear,
    .m_free = iomodule_free,
    .m_slots = iomodule_slots,
};
```

I thought that the PyInit function should be suffixed with the same name as the one in "m_name" but it doesn't seem the case here. What I know is that there is an entry in the Setup.bootstrap.in that has `_io _iomodule/...` so I assume that the `_io` here is paired with `PyInit_*`.

As for those in Setup.stdlib.in, it's something like `
@MODULE_NAME_TRUE@name ...` and I assume that "NAME == name.upper()" and that "name" is the name of the module to be exported, right? or is it something else?

developer-workflow/extension-modules.rst
picnixz: Recommending `-j` and `-jN` instead.

developer-workflow/extension-modules.rst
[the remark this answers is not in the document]
picnixz: Update: IIUC, the marker in Setup.stdlib.in or Setup.bootstrap.in is the one that should be matched against the `PyInit_*` function.

developer-workflow/extension-modules.rst
erlend-aasland: > [...] but if it would, it would be a shared extension without a Python implementation.

... which _is_ the case today; there is no Python implementation.

erlend-aasland:
> And that `c` stands for "complex", not "implemented in C".

Oh, TIL! I never looked in that corner of the code base (yet) :)

erlend-aasland:
> I don't know a lot about the limited API and I think the tutorial should not necessarily have too much things as an introduction.

It may make sense to separate out some concepts, such as the limited API, into how-to guides instead of baking them into the tutorial.

developer-workflow/extension-modules.rst
encukou: Yup.
Compare to something like `binascii`, which also doesn't have a Python implementation, but would absolutely need one if it was added today.

encukou:
> I don't know a lot about the limited API and I think the tutorial should not necessarily have too much things as an introduction.

@vstinner has been converting modules to limited API, maybe he's a better person to argue that it should be in the initial example, as a default to start with :)
For starters, it makes the module friendlier to non-CPython implementations, as the limited API is (nowadays) designed to avoid CPython-specific details.
Removing the define gives you access to more API.

`Py_BUILD_CORE_MODULE` is kind of the opposite: gives you access to internals, which might change frequently. That coupling makes both the module and the internals harder to maintain, but you get more power and speed.

picnixz:
Thank you for the explanation on the Limited API!  IIUC:

- With Limited API + `#include "Python.h"` -> 
- only `#include "Python.h"` -> do not expose internals
- Py_BUILD_CORE_MODULE + `#include "Python.h"` -> expose internals

Ideally, I don't want to duplicate the explanation on the macros themselves (I don't remember where it is) but I can a small paragraph summarizing the above? What remains to add on the page is (in separate PRs):

- configuration variables (without explaining how to use autoconf syntax, I'll just link a reference to the autoconf page). Typically useful when adding stuff related to maths.
- frozen modules, but I don't really know about them so I'll leave it to someone else (I didn't find any definition of them...)

encukou:
The limited API is [documented publicly](https://docs.python.org/3/c-api/stable.html), but Py_BUILD_CORE_MODULE is internal. I think this would be a good place to mention it, since it's only useful for core modules. But, it doesn't need much more than that bullet point.

picnixz:
I don't really want to add it in the main section because it could disrupt the reading flow.

So I think I can add something in the Troubleshooting section when the compiler complains with `Py_BUILD_CORE` being missing when you include some headers for some functions (and you don't know that you actually needed that macro). That way, people wouldn't add the `Py_BUILD_CORE` macro directly (the compiler message is literal "this requires `Py_BUILD_CORE` and I think it's a bit misleading since it does not *need* `Py_BUILD_CORE` itself but rather `Py_BUILD_CORE_MODULE`).

Actually, there is this page: https://docs.python.org/3/using/configure.html#c-extensions which explains somewhat those macros. Should it be moved to the devguide? it also explains the difference with the `Py_BUILD_CORE_BUILTIN` macro.

(By the way, I don't like the naming "module" and "built-in" and I was confused a lot at first! I would have preferred something like "static" and "shared" because it's much more explanatory... but I think we cannot change this =/ (unless there are some other conditions I am not aware of)).

encukou:
Yes, I think mentioning `Py_BUILD_CORE_BUILTIN` in the user-facing documentation is a mistake. Users could read that page and think that they can define `Py_BUILD_CORE_MODULE` for their own dynamic libraries.
@vstinner, would you agree with moving the mentions of `Py_BUILD_CORE*` macros from [`configure` docs](https://docs.python.org/3/using/configure.html#c-extensions) to the devguide?

> I don't like the naming "module" and "built-in" and I was confused a lot at first! I would have preferred something like "static" and "shared" because it's much more explanatory... 

Maybe some of the confusion remains? 
Most core extension modules can be *either* linked statically or be a shared library. The person who *builds Python*, not the one that writes the module code, makes that choice. 

IMO, the distinction isn't all that important, and it could moved all the way down to “Configuring the CPython project”, with `Setup.stdlib.in` treated as the normal and `Setup.bootstrap.in` as a less important special case.

The reasons for putting a module `Modules/Setup.bootstrap.in` are:
- the module defines C APIs that are used in core
- the module is used by importlib, deepfreeze, freeze, runpy, or sysconfig
- it's a “commonly used core module”

If you're adding your first module, these aren't very likely to apply.

A similar special case, then, is modules that *must* be built as shared libraries -- that's mostly modules that test the loading of shared libraries.

vstinner:
> @vstinner, would you agree with moving the mentions of Py_BUILD_CORE* macros from [configure docs](https://docs.python.org/3/using/configure.html#c-extensions) to the devguide?

I would prefer to keep the doc in the configure docs. You can duplicate the doc in the devguide if you want.

vstinner (approved):
Nice work! LGTM.

picnixz:
> Maybe some of the confusion remains?

No.... the confusing is because I'm tired and I forgot about the `*shared*` and `*static*` markers... (I even wrote a section about them!!)

I'll add the paragraph on the macros. I'm waiting for https://github.com/python/cpython/issues/122544 to be decided because the troubleshooting section will need to be updated (and I'm taking holidays next week so it's better to wait for the correct image to be chosen before merging (it saves us 1 PR if we merge it before)).

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
 Standard library extension modules
 ==================================
 
-In this section, we could explain how to write a CPython extension with the C language, but the topic can take a complete book.
+In this section, we are interested in extending the CPython project with
+an :term:`extension module`. We will not explain how to write the module
+in C but rather explain how to configure the project and make it compile.
```
willingc, line 10:
```suggestion
In this section, we explain how to configure and compile the CPython project with
a C :term:`extension module`. We will not explain how to write a C extension module and
offer links to other documentation.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+   while shared extension modules are *optional* in the sense that they
+   might be supplied, overridden or disabled externally.
+
+   Usually, accelerator modules are built as *shared* extension modules,
+   especially if they already have a pure Python implementation.
+
+According to :pep:`399`, *new* extension modules MUST provide a working and
+tested pure Python implementation, unless a special dispensation is given.
+Please ask the :github:`Steering Council <python/steering-council>` if such
+dispensation is needed.
```
willingc, line 65:
```suggestion
According to :pep:`399`, *new* extension modules MUST provide a working and
tested pure Python implementation, unless a special dispensation from
the :github:`Steering Council <python/steering-council>` is given.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+Adding an extension module to CPython
+=====================================
+
+Assume that the standard library contains a pure Python module :mod:`!foo`
+together with the following :func:`!foo.greet` function:
```
willingc, line 71:
```suggestion
with the following :func:`!foo.greet` function:
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+.. code-block:: python
+   :caption: Lib/foo.py
+
+   def greet():
+       return "Hello World!"
+
+Instead of using the Python implementation of :func:`!foo.greet`, we want to
+use its corresponding C implementation exposed in some :mod:`!_foo` module
+written in C. Ideally, we want to modify :cpy-file:`!Lib/foo.py` as follows:
```
willingc, line 81:
```suggestion
Instead of using the Python implementation of :func:`!foo.greet`, we want to
use its corresponding C extension implementation exposed in :mod:`!_foo` module.
Ideally, we want to modify :cpy-file:`!Lib/foo.py` as follows:
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+   except ImportError:
+       # fallback to the pure Python implementation
+       def greet():
+           return "Hello World!"
+
+.. note::
+
+   Accelerator modules should *never* be imported directly, whence the
+   convention is to mark them as private implementation details with the
+   underscore prefix (namely, :mod:`!_foo` in this example).
```
willingc, line 98:
```suggestion
   Accelerator modules should *never* be imported directly. The
   convention is to mark them as private implementation details with the
   underscore prefix (namely, :mod:`!_foo` in this example).
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+In order to incorporate the accelerator module, we need to determine:
+
+- where to place the extension module source code in the CPython project tree,
+- which files to modify in order to compile the CPython project, and
+- which ``Makefile`` rules to invoke at the end.
```
willingc, line 104:
```suggestion
- where to update the CPython project tree with the extension module source code 
- which files to modify to configure and compile the CPython project, and
- which ``Makefile`` rules to invoke at the end.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+By convention, the source file containing the extension module implementation
+is called ``<NAME>module.c``, where ``<NAME>`` is the name of the module that
+will be later imported (in our case :mod:`!_foo`). In addition, the directory
+containing the implementation should also be named similarly.
+
+One could imagine having more files, or no helper files at all. Here,
+we wanted to illustrate a simple example without making it too trivial. If
+the extension module does not require additional files, it may directly be
+placed in :cpy-file:`Modules` as ``Modules/_foomodule.c``.
```
willingc, line 127:
```suggestion
Usually, accelerator modules are added in the :cpy-file:`Modules` directory of
the CPython project. If more than one file is needed for the extension module,
it is more convenient to create a sub-directory in :cpy-file:`Modules`.

In the simplest example where the extension module is one file, it may be
placed in :cpy-file:`Modules` as ``Modules/_foomodule.c``.

For a non-trivial example of the extension module :mod:`!_foo`, we consider
the following working tree:

- :ref:`Modules/_foo/_foomodule.c` --- the extension module implementation.
- :ref:`Modules/_foo/helper.h` --- the extension helpers declarations.
- :ref:`Modules/_foo/helper.c` --- the extension helpers implementations.

By convention, the source file containing the extension module implementation
is called ``<NAME>module.c``, where ``<NAME>`` is the name of the module that
will be later imported (in our case :mod:`!_foo`). In addition, the directory
containing the implementation should also be named similarly.
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+   module name ``<NAME>`` used in import statements (here ``_foo``),
+   and which usually coincides with :c:member:`PyModuleDef.m_name`.
+
+   Other identifiers such as those used in :ref:`Argument Clinic <clinic>`
+   inputs do not have such naming requirements.
+
+Configuring the CPython project
+-------------------------------
+
+Now that we have implemented our extension module, we need to update some
```
willingc, line 279:
```suggestion
Now that we have added our extension module to the CPython source tree, we need to update some
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+  .. code-block:: text
+     :caption: :cpy-file:`configure.ac`
+     :emphasize-lines: 2, 3
+
+     PY_STDLIB_MOD([_foo],
+                   [test "$ac_sys_system" = "Linux"],
+                   [test "$ARCH_RUN_32BIT" = "true"],
+                   [-I\$(srcdir)/Modules/_foo], [])
+
+  More generally, the status of the extension is determined as follows:
```
willingc, line 405:
```suggestion
  More generally, the host's configuration status of the extension is determined as follows:
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+  +-----------+-----------------+----------+
+  | Enabled   | Supported       | Status   |
+  +===========+=================+==========+
+  | true      | true            | yes      |
```
willingc, line 410: Perhaps available instead of yes.

```suggestion
  | true      | true            | available      |
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+.. tip::
+
+   Header files use ``<ClInclude>`` tags, whereas
+   source files use ``<ClCompile>`` tags.
+
+
+Compiling the CPython project
+-----------------------------
+
+Now that everything is in place, it remains to compile the project:
```
willingc, line 542:
```suggestion
Now that configuration is in place, it remains to compile the project:
```

developer-workflow/extension-modules.rst
```diff
@@ -5,13 +5,679 @@
...
+
+* The final ``make`` step is generally not needed since the previous ``make``
+  invokations may completely rebuild the project, but it could be needed in
+  some specific cases.
+
+Troubleshooting
+---------------
+
+This section addresses common issues that you may face when following
+this tutorial.
```
willingc, line 574:
```suggestion
this example of adding an extension module.
```

willingc (commented):
Overall, fantastic job @picnixz. I've made several suggestions to help improve the flow and comprehension of the example. Thank you!

developer-workflow/extension-modules.rst
picnixz: Actually, for this one, I will to disagree as `./configure` says:

```
dnl Check for stdlib extension modules
dnl PY_STDLIB_MOD([NAME], [ENABLED-TEST], [SUPPORTED-TEST], [CFLAGS], [LDFLAGS])
dnl sets MODULE_$NAME_STATE based on PY_STDLIB_MOD_SET_NA(), ENABLED-TEST,
dnl and SUPPORTED_TEST. ENABLED-TEST and SUPPORTED-TEST default to true if
dnl empty.
dnl    n/a: marked unavailable on platform by PY_STDLIB_MOD_SET_NA()
dnl    yes: enabled and supported
dnl    missing: enabled and not supported
dnl    disabled: not enabled
``` 

and 

``` 
  AS_IF([test "$modstate" != "n/a"], [
    AS_IF([m4_ifblank([$2], [true], [$2])],
       [AS_IF([m4_ifblank([$3], [true], [$3])], [modstate=yes], [modstate=missing])],
       [modstate=disabled])
  ])
```

willingc: Sounds good. Thanks for clarifying.

willingc (approved):
Thank you @picnixz for a helpful addition to the devguide. You did a lovely job with this. I'm going to merge this and any further edits can be made in future PRs. 💐 ☀️

PR #1351, opened 2024-07-15
Replace custom roles with styled extlinks

hugovk:
We can replace the custom roles used for `cpy-file` and `gh-label` with [extlinks](https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html), which removes a fair amount of code. One thing, though, the custom roles had code formatting, and extlinks have no special formatting by default.

However, with the newly-released Sphinx 7.4.0, a unique class is added to each extlink, so we can now add our own styling.

* https://www.sphinx-doc.org/en/master/changes.html#release-7-4-0-released-jul-15-2024
* https://github.com/sphinx-doc/sphinx/pull/12319
* https://github.com/sphinx-doc/sphinx/issues/11745

----
📚 Documentation preview 📚: https://cpython-devguide--1351.org.readthedocs.build/


_extensions/custom_roles.py
```diff
@@ -1,43 +0,0 @@
...
-def autolink(pattern):
-    def role(name, rawtext, text, lineno, inliner, _options=None, _content=None):
-        """Combine literal + reference (unless the text is prefixed by a !)."""
-        if " " in text:
-            url_text = urllib.parse.quote(text)
-        else:
-            url_text = text
-        url = pattern.format(url_text)
-        # don't create a reference if the text starts with !
-        if text.startswith('!'):
```
ezio-melotti, line 35: Does this work with extlinks?

AA-Turner: No; see https://cpython-devguide--1351.org.readthedocs.build/internals/parser/#how-to-regenerate-the-parser

However we could just use ``:file:`Makefile` `` or ``` ``Makefile`` ```?

A

hugovk: We have quite a few ``` ``Makefile`` ``` already, so I've changed ``` :cpy-file:`!Makefile` ``` to that.

ezio-melotti (approved):

AA-Turner (approved):

PR #1352, opened 2024-07-15
Clarify 'time zone' in a Specific words section

nedbat:
https://github.com/python/cpython/pull/118449 corrects "timezone" to "time zone".  This records the guidance for posterity.

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -54,10 +54,15 @@ starting it with a lowercase letter should be avoided.
 Many special names are used in the Python documentation, including the names of
 operating systems, programming languages, standards bodies, and the like. Most
 of these entities are not assigned any special markup, but the preferred
-spellings are given here to aid authors in maintaining the consistency of
-presentation in the Python documentation.
+spellings are given in :ref:`glossary` to aid authors in maintaining the
+consistency of presentation in the Python documentation.
 
-Other terms and words deserve special mention as well; these conventions should
+.. _glossary:
+
+Specific words
+==============
+
+Some terms and words deserve special mention. These conventions should
 be used to ensure consistency throughout the documentation:
 
 C API
@@ -78,6 +83,11 @@ reST
    For "reStructuredText," an easy to read, plaintext markup syntax
    used to produce Python documentation.  When spelled out, it is
    always one word and both forms start with a lowercase 'r'.
+
+time zone
+   Spell out as two words when talking about the real-world concept.
+   Only spell as one word when referring to a Python term like a module,
+   class, or argument.
 
 Unicode
    The name of a character coding system. This is always written


AA-Turner (approved):
We could perhaps use a ``.. glossary::`` directive here?

A

hugovk (approved):

documentation/style-guide.rst
```diff
@@ -79,6 +84,11 @@ reST
    used to produce Python documentation.  When spelled out, it is
    always one word and both forms start with a lowercase 'r'.
 
+time zone
+   Spell out as two words when talking about the real-world concept.
+   Only spell as one word when referring to a Python term like a module,
+   class, or argument.
```
ezio-melotti, line 90: Shouldn't they have markup like `:mod:` or similar in this case?
If that's the case, "timezone" should never appear in plain text.
Not sure if it's worth clarifying this.

erlend-aasland: > Not sure if it's worth clarifying this.

Anyone reading the devguide style guide will probably already know how to mark up classes, so I'm not sure it's worth complicating the current wording.

erlend-aasland (approved):

nedbat:
> We could perhaps use a `.. glossary::` directive here?

I'm not sure this rises to that level yet, but will keep it in mind.

documentation/style-guide.rst
```diff
@@ -54,10 +54,15 @@ starting it with a lowercase letter should be avoided.
...
-spellings are given here to aid authors in maintaining the consistency of
-presentation in the Python documentation.
+spellings are given in :ref:`glossary` to aid authors in maintaining the
+consistency of presentation in the Python documentation.
 
-Other terms and words deserve special mention as well; these conventions should
+.. _glossary:
+
+Specific words
+==============
```
hugovk, line 63: I predict this will confuse me, shall we have the same name for both the link and the title?

documentation/style-guide.rst
ezio-melotti: What I meant is:
* talking about the concept? Always use two words and no markup.
* talking about the module/class/arg? Always use one word and the appropriate markup.

This implies that "timezone" will never appear in plain text and "time zone" will never appear within markup.

IOW, the fact that you are using markup or not already determines whether to use one or two words, without having to think if you are referring to a module/class/argument or not.

documentation/style-guide.rst
ezio-melotti: There is also a [`glossary` directive](https://devguide.python.org/documentation/markup/#roles) (and a `:term:` role) that might add to the confusion.

documentation/style-guide.rst
erlend-aasland: Sure, if it does not clutter the wording too much, I'm fine with such a change.

erlend-aasland:
Related:

- https://github.com/python/cpython/pull/121907

documentation/style-guide.rst
ezio-melotti, line 90:
```suggestion
   When referring to a Python term like a module, class, or argument
   it is always spelled as one word and surrounded by the appropriate markup
   (for example, ``:mod:`timezone```).
   When talking about the real-world concept it is spelled as two words.
```

nedbat: I added a modified version of this text.

documentation/style-guide.rst
nedbat: I changed the reference to `specific words`.

PR #1362, opened 2024-08-04
Remove GNU make-specific directive from Makefile

hugovk:
Like https://github.com/python/cpython/pull/122662 and https://github.com/python/peps/pull/3891.

This is mostly to keep the `Makefile`s similar in each repo to ease maintenance.

Also remove the `make serve` stub, the target was removed in April 2022 in https://github.com/python/devguide/pull/826.

----
📚 Documentation preview 📚: https://cpython-devguide--1362.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,11 +198,6 @@ _ensure-pre-commit:
 lint: _ensure-pre-commit
 	$(VENVDIR)/bin/python3 -m pre_commit run --all-files
 
-.PHONY: serve
-serve:
-	@echo "The 'serve' target was removed, use 'htmlview' instead" \
-	      "(see https://github.com/python/cpython/issues/80510)"
-
 include/branches.csv: include/release-cycle.json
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,15 +2,17 @@
...
+VENVDIR      = ./venv
+UV           = uv
+SPHINXBUILD  = $(VENVDIR)/bin/sphinx-build
+SPHINXOPTS   = --fail-on-warning --keep-going
+BUILDDIR     = _build
+BUILDER      = html
+JOBS         = auto
+PAPER        =
+SPHINXLINT   = $(VENVDIR)/bin/sphinx-lint
+REQUIREMENTS = requirements.txt
```
hugovk, line 15: The main changes here are adding the `UV` and `REQUIREMENTS` lines.

The others have a space removed, so it's the same as the CPython/Docs and peps `Makefiles` and easier to diff.

AA-Turner (approved):

ezio-melotti (approved):

PR #1365, opened 2024-08-08
Remove obsolete `serve` target from `make.bat`.

ezio-melotti:
This is a follow-up of:
* #1362 

The PR removed the `make serve` target from the `Makefile` but not from `make.bat`.

---

Slightly unrelated, but I noticed that the `serve` replacement -- `htmllive` -- seems to be missing from `make.bat`.  `htmllive` (formerly known as `autobuild`) was introduced in:
* #1208 
`sphinx-autobuild` seems to be multiplat, so it should be possible to add it for Windows too.

----
📚 Documentation preview 📚: https://cpython-devguide--1365.org.readthedocs.build/

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -218,11 +218,6 @@ rem Ignore the tools and venv dirs and check that the default role is not used.
 cmd /S /C "%SPHINXLINT% -i tools -i venv --enable default-role"
 goto end
 
-:serve
-	echo.The serve target was removed, use htmlview instead ^
-(see https://github.com/python/cpython/issues/80510)
-goto end
-
 if "%1" == "versions" (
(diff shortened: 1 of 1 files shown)


AA-Turner (approved):

hugovk (approved):
> Slightly unrelated, but I noticed that the `serve` replacement -- `htmllive` -- seems to be missing from `make.bat`.

I probably skipped it because I don't have Windows to test it on. But it's so useful, I'd recommend someone with Windows adds it.

AA-Turner:
I don't think I've once used `make.bat`, to be honest. One of those Sphinx people should really look into making their CLI better...!

A

ezio-melotti:
* #1366

PR #1370, opened 2024-08-14
Use "modern" Makefile template

AA-Turner:
Sphinx added this format [in 2014](https://github.com/sphinx-doc/sphinx/commit/91d92be8adae2e44e1e23e1c07b7088c35261abf). Nevertheless, it is the "modern" format, and far less verbose in the Makefile.

A

----
📚 Documentation preview 📚: https://cpython-devguide--1370.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,36 +17,19 @@ REQUIREMENTS = requirements.txt
 # Internal variables.
 PAPEROPT_a4     = --define latex_paper_size=a4
 PAPEROPT_letter = --define latex_paper_size=letter
-ALLSPHINXOPTS   = --builder $(BUILDER) \
-                  --doctree-dir $(BUILDDIR)/doctrees \
-                  --jobs $(JOBS) \
+ALLSPHINXOPTS   = --jobs $(JOBS) \
                   $(PAPEROPT_$(PAPER)) \
-                  $(SPHINXOPTS) \
-                  . $(BUILDDIR)/$(BUILDER)
+                  $(SPHINXOPTS)
 
 .PHONY: help
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  venv       to create a venv with necessary tools"
 	@echo "  html       to make standalone HTML files"
+	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  htmlview   to open the index page built by the html target in your browser"
 	@echo "  htmllive   to rebuild and reload HTML files in your browser"
 	@echo "  clean      to remove the venv and build files"
-	@echo "  dirhtml    to make HTML files named index.html in directories"
-	@echo "  singlehtml to make a single large HTML file"
-	@echo "  pickle     to make pickle files"
-	@echo "  json       to make JSON files"
-	@echo "  htmlhelp   to make HTML files and a HTML help project"
-	@echo "  qthelp     to make HTML files and a qthelp project"
-	@echo "  devhelp    to make HTML files and a Devhelp project"
-	@echo "  epub       to make an epub"
-	@echo "  latex      to make LaTeX files, you can set PAPER=a4 or PAPER=letter"
-	@echo "  latexpdf   to make LaTeX files and run them through pdflatex"
-	@echo "  text       to make text files"
-	@echo "  man        to make manual pages"
-	@echo "  changes    to make an overview of all changed/added/deprecated items"
-	@echo "  linkcheck  to check all external links for integrity"
-	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  check      to run a check for frequent markup errors"
 	@echo "  lint       to lint all the files"
 	@echo "  versions   to update release cycle after changing release-cycle.json"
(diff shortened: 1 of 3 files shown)


hugovk:
Before, in classic `make` style, this would only run the script when the `include/release-cycle.json` dependency had been changed more recently than the `include/branches.csv`, `include/end-of-life.csv` and `include/release-cycle.svg` targets.

Now, it always runs it.

Can we maintain the original behaviour?

AA-Turner:
In my local testing, it always re-generated, because `versions` was marked as PHONY. I might have missed something or not got the setup quite right through WSL, though.

I'm out at the moment but will have a look when I get back.

A

AA-Turner:
Current behaviour (adding a "date" command into the `versions` target for double-checking):

```console
.../devguide$ make html
venv already exists.
To recreate it, remove it first with `make clean-venv'.
date "+%Y-%m-%d %H:%M:%S"
2024-08-14 17:54:08
Release cycle data generated.
./venv/bin/sphinx-build --builder html --doctree-dir _build/doctrees --jobs auto --fail-on-warning --keep-going . _build/html
[snip]
.../devguide$ make html
venv already exists.
To recreate it, remove it first with `make clean-venv'.
date "+%Y-%m-%d %H:%M:%S"
2024-08-14 17:54:39
Release cycle data generated.
./venv/bin/sphinx-build --builder html --doctree-dir _build/doctrees --jobs auto --fail-on-warning --keep-going . _build/html
[snip]
```

A

AA-Turner:
I haven't worked out how to construct a [pattern rule](https://www.gnu.org/software/make/manual/make.html#Pattern-Rules) that looks for an naked name (no suffix, no directory), so this hopefully serves as a work-around. Feel free to test on a proper UNIX system!

A

hugovk:
## On `main`

Yeah, so `make html` is calling `make versions` each time. I'm not too bothered either way about that.

However, `versions` depends on the CSVs and SVG. 

The targets for those depend on the JSON.

The flow is: JSON -> CSVs + SVG

If the JSON is older than the CSVs and SVG, then those ones won't run.

We can test this by putting a "hello" in the Python script:

<details>
<summary>example diff</summary>

```diff
diff --git a/Makefile b/Makefile
index 1f8e73e..6aef618 100644
--- a/Makefile
+++ b/Makefile
@@ -210,3 +210,4 @@ include/release-cycle.svg: include/release-cycle.json
 .PHONY: versions
 versions: venv include/branches.csv include/end-of-life.csv include/release-cycle.svg
 	@echo Release cycle data generated.
+	date "+%Y-%m-%d %H:%M:%S"
diff --git a/_tools/generate_release_cycle.py b/_tools/generate_release_cycle.py
index 27b5cc3..1027154 100644
--- a/_tools/generate_release_cycle.py
+++ b/_tools/generate_release_cycle.py
@@ -153,4 +153,5 @@ def main() -> None:
 
 
 if __name__ == "__main__":
+    print("hello")
     main()
```

</details>

Let's make sure the JSON is newer than the CSVs and SVG:

```console
❯ touch include/branches.csv include/end-of-life.csv include/release-cycle.svg

❯ touch include/release-cycle.json
```
Then run it and we see "hello":
```console
❯ make versions
venv already exists.
To recreate it, remove it first with `make clean-venv'.
date "+%Y-%m-%d %H:%M:%S"
2024-08-14 21:24:40
./venv/bin/python3 _tools/generate_release_cycle.py
hello
Release cycle data generated.
date "+%Y-%m-%d %H:%M:%S"
2024-08-14 21:24:40
```
As expected, the script ran and updated the CSVs and SVG.

Let's run again. This time the JSON is older than the newly updated CSVs and SVG. There's no "hello"; the script isn't run:

```
❯ make versions
venv already exists.
To recreate it, remove it first with `make clean-venv'.
Release cycle data generated.
date "+%Y-%m-%d %H:%M:%S"
2024-08-14 21:27:03
```

hugovk:
## On the PR

Adding the `print("hello")` and running similar commands gives us just a single "hello" so this looks like it's working now :)

```console
❯ touch include/branches.csv include/end-of-life.csv include/release-cycle.svg

❯ touch include/release-cycle.json

❯ make html
./venv/bin/python3 _tools/generate_release_cycle.py
hello
Release cycle data generated.
./venv/bin/sphinx-build -M html "." "_build" --jobs auto --fail-on-warning --keep-going -q
matplotlib is not installed, social cards will not be generated

❯ make html
./venv/bin/sphinx-build -M html "." "_build" --jobs auto --fail-on-warning --keep-going -q
matplotlib is not installed, social cards will not be generated
```

Will review a bit closer :)

Makefile
```diff
@@ -83,89 +65,6 @@ ensure-venv:
...
-linkcheck: html
-	@echo
-	@echo "Link check complete; look for any errors in the above output " \
-	      "or in $(BUILDDIR)/$(BUILDER)/output.txt."
-
-.PHONY: doctest
-doctest: BUILDER = doctest
-doctest: html
-	@echo "Testing of doctests in the sources finished, look at the " \
-	      "results in $(BUILDDIR)/$(BUILDER)/output.txt."
```
hugovk, line 167: So all of the above are handled by Sphinx?

(I don't know if anyone uses any of the more exotic ones like pickle, epub, latex and so on in this repo, I bet some of them don't even work.)

Makefile
```diff
@@ -10,50 +10,32 @@ SPHINXOPTS   = --fail-on-warning --keep-going
...
-	@echo "  latex      to make LaTeX files, you can set PAPER=a4 or PAPER=letter"
-	@echo "  latexpdf   to make LaTeX files and run them through pdflatex"
-	@echo "  text       to make text files"
-	@echo "  man        to make manual pages"
-	@echo "  changes    to make an overview of all changed/added/deprecated items"
-	@echo "  linkcheck  to check all external links for integrity"
-	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  check      to run a check for frequent markup errors"
 	@echo "  lint       to lint all the files"
 	@echo "  versions   to update release cycle after changing release-cycle.json"
```
hugovk, line 33: Shall we remove this `versions` help, or reinstate `make versions`?

hugovk (commented):
On `main`, we get:

```console
❯ make html
venv already exists.
To recreate it, remove it first with `make clean-venv'.
Release cycle data generated.
./venv/bin/sphinx-build --builder html --doctree-dir _build/doctrees --jobs auto  --fail-on-warning --keep-going . _build/html
Running Sphinx v8.0.2
loading translations [en]... done
matplotlib is not installed, social cards will not be generated
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
writing output...
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
reading sources...
looking for now-outdated files... none found
no targets are out of date.
Writing redirects...
build succeeded.

The HTML pages are in _build/html.
```


On the PR we get:
```console
❯ make html
./venv/bin/sphinx-build -M html "." "_build" --jobs auto --fail-on-warning --keep-going -q
matplotlib is not installed, social cards will not be generated
```

Comparing the called commands, rearranging and adding spaces to aid comparison:

```diff
-./venv/bin/sphinx-build --builder html --doctree-dir _build/doctrees --jobs auto --fail-on-warning --keep-going  .   _build/html
+./venv/bin/sphinx-build        -M html                               --jobs auto --fail-on-warning --keep-going "." "_build"    -q
```

So that removed `--doctree-dir _build/doctrees` is fine because it's the default.

The output dir change from `_build/html` to `_build` doesn't seem to matter, the output files are in `_build/html` anyway.

What's `-M`? It seems to be `--builder` but the help says the short option is `-b`.

It also adds `-q` aka `--quiet`: "no output on stdout, just warnings on stderr".

Do we want to suppress the output? Is there a way to permanently suppress that `-q`, or temporarily do so for a single run?

AA-Turner:
> So that removed `--doctree-dir _build/doctrees` is fine because it's the default.
> 
> The output dir change from `_build/html` to `_build` doesn't seem to matter, the output files are in `_build/html` anyway.
> 
> What's `-M`? It seems to be `--builder` but the help says the short option is `-b`.

Sphinx's [*make mode*](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-M) -- fully replicating the old-style (e.g. here) `Makefile` logic within Sphinx. In effect it just sets `doctreedir` to `<build>/doctrees` and `builddir` to `<build>/<builder_name>`.

> It also adds `-q` aka `--quiet`: "no output on stdout, just warnings on stderr".
> 
> Do we want to suppress the output? Is there a way to permanently suppress that `-q`, or temporarily do so for a single run?

This was unintended, sorry (accidental commit of local tests).

Makefile
AA-Turner: Yep, the `%` catch-all target passes whatever name through to Sphinx. But as you say, many are unlikely to be used.

hugovk (approved):
Nice cleanup, thanks!

I've not reviewed `make.bat` or `make.ps1`.

PR #1372, opened 2024-08-14
Adopt Ruff

AA-Turner:
I've set `target-version = "py312"` as that is what we have in CPython, but we could lower this requirement if wanted.

A

----
📚 Documentation preview 📚: https://cpython-devguide--1372.org.readthedocs.build/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,33 +1,13 @@
 repos:
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.5.7
     hooks:
-      - id: pyupgrade
-        args: [--py38-plus]
-
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 23.12.1
-    hooks:
-      - id: black
-        args: [--skip-string-normalization]
-
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.13.2
-    hooks:
-      - id: isort
-        args: [--profile=black]
-
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.1.0
-    hooks:
-      - id: flake8
-        additional_dependencies:
-          [flake8-2020, flake8-implicit-str-concat]
-
-  - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.10.0
-    hooks:
-      - id: python-check-blanket-noqa
+      - id: ruff-format
+        name: Run Ruff (format)
+        args: [--check]
+      - id: ruff
+        name: Run Ruff (lint)
+        args: [--exit-non-zero-on-fix]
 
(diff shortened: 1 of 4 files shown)


.ruff.toml
```diff
@@ -0,0 +1,34 @@
+target-version = "py312"
+fix = true
```
ezio-melotti, line 2: Does this fix the issues automatically on commit?

AA-Turner: Not on commit, but when you run `ruff check`.

.ruff.toml
```diff
@@ -0,0 +1,34 @@
...
+    "G",     # flake8-logging-format
+    "I",     # isort
+    "ISC",   # flake8-implicit-str-concat
+    "LOG",   # flake8-logging
+    "PERF",  # perflint
+    "PGH",   # pygrep-hooks
+    "PT",    # flake8-pytest-style
+    "TCH",   # flake8-type-checking
+    "UP",    # pyupgrade
+    "W",     # pycodestyle
```
hugovk, line 25: We had this for Flake8, let's keep it:
```suggestion
    "W",     # pycodestyle
    "YTT",   # flake8-2020
```

.pre-commit-config.yaml
```diff
@@ -1,33 +1,13 @@
...
-      - id: python-check-blanket-noqa
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.5.7
+    hooks:
+      - id: ruff-format
+        name: Run Ruff (format)
+        args: [--check]
+      - id: ruff
+        name: Run Ruff (lint)
+        args: [--exit-non-zero-on-fix]
```
hugovk, line 10: Ruff's pre-commit docs say:

> When running with `--fix`, Ruff's lint hook should be placed _before_ Ruff's formatter hook, and
_before_ Black, isort, and other formatting tools, as Ruff's fix behavior can output code changes
that require reformatting.

https://github.com/astral-sh/ruff-pre-commit#using-ruff-with-pre-commit

```suggestion
      - id: ruff
        name: Run Ruff (lint)
        args: [--exit-non-zero-on-fix]
      - id: ruff-format
        name: Run Ruff (format)
        args: [--check]
```

hugovk (commented):
> I've set `target-version = "py312"` as that is what we have in CPython, but we could lower this requirement if wanted.

I don't mind much, but some others have asked for lower versions before. We don't gain a lot by requiring a higher version, just the `datetime` thing at the moment, which isn't too much bother.

.ruff.toml
```diff
@@ -0,0 +1,34 @@
+target-version = "py312"
```
AA-Turner, line 1:
```suggestion
target-version = "py311"
```

.ruff.toml
AA-Turner: I won't in this PR to keep it more focussed (I'm also unfamiliar with the service), but perhaps something to consider.

A

Issue #1375, opened 2024-08-16
Forking & Branching instructions on Git bootcamp and cheat sheet page


kevinbowen777:
Hello,
I am in the process of working through the initial steps on the [Git bootcamp and cheat sheet](https://devguide.python.org/getting-started/git-boot-camp/) for the first time.
I am wondering if I am doing something incorrectly or, if there are some steps
missing that could use clarification.

In the "Forking CPython GitHub repository" section, should there be a note to
uncheck the "Copy the main branch only" check box in the GitHub forking dialogue?

If one leaves the box checked, then in the "Creating and switching branches"
section, when one issues the command `git branch -a`, no remote branches are
visible.

If the box is unchecked in the initial forking section, then the remote branches are
visible. For example:
```
* main
  remotes/origin/3.10
  remotes/origin/3.11
  remotes/origin/3.12
  <snip>
  ...
```

Would this also require changing the command in the same branch section from
`git switch -c 3.12 upstream/3.12`
to
`git switch -c 3.12 remotes/origin/3.12`?

Or, is there a step missing that would allow one to use the original `git switch`
command with `upstream/3.12` as the start point?

Please let me know if you need additional information. If I am not wrong in my assumptions, I would be glad to help update the documentation.

Thanks in advance.

hugovk:
> In the "Forking CPython GitHub repository" section, should there be a note to uncheck the "Copy the main branch only" check box in the GitHub forking dialogue?

Yes, let's add this to the instructions.

I think the checkbox is relatively new, at least newer than the instructions, and before it used to copy all branches.

Would you like to create a PR?

At the same time, let's change that URL in the last step to be in code formatting, we don't need a partial link there: 

![image](https://github.com/user-attachments/assets/c48eae68-99cd-4d0a-8f1a-5e0437a114bc)

hugovk:
> If the box is unchecked in the initial forking section, then the remote branches are visible. For example:
> 
> ```
> * main
>   remotes/origin/3.10
>   remotes/origin/3.11
>   remotes/origin/3.12
>   <snip>
>   ...
> ```
> 
> Would this also require changing the command in the same branch section from `git switch -c 3.12 upstream/3.12` to `git switch -c 3.12 remotes/origin/3.12`?
> 
> Or, is there a step missing that would allow one to use the original `git switch` command with `upstream/3.12` as the start point?


With a fresh clone, I got:

```console
❯ git switch -c 3.12 upstream/3.12
[...]

PR #1376, opened 2024-08-18
Improve forking instructions (#1375)

kevinbowen777:
- Add item to uncheck "Copy main branch only" button
  - Add `git fetch upstream` command to creating branches section

----
📚 Documentation preview 📚: https://cpython-devguide--1376.org.readthedocs.build/

### DIFF
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -46,11 +46,11 @@ You will only need to do this once.
 
 2. Press ``Fork`` located near the top right of the page.
 
-4. Uncheck the "Copy the ``main`` branch only" button.
+3. Uncheck the "Copy the ``main`` branch only" button.
 
-3. Press the ``Create fork`` button.
+4. Press the ``Create fork`` button.
 
-4. Your forked CPython repository will be created at ``https://github.com/<username>/cpython``.
+5. Your forked CPython repository will be created at ``https://github.com/<username>/cpython``.
 
 .. _clone-your-fork:
 
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -46,7 +46,6 @@ You will only need to do this once.
 
 2. Press ``Fork`` located near the top right of the page.
 
-   * The required fields ``Owner`` and ``Repository name`` should already be pre-populated with your username and the original repository name "cpython".
    * Uncheck the "Copy the ``main`` branch only" button.
 
 3. Press the ``Create fork`` button.
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -108,7 +108,7 @@ To verify the upstream for ``main``::
 It should emit ``upstream``, indicating to track/pull changes for ``main`` from the
 ``upstream`` remote.
 
-Once this is verified, update your local repository with the upstream branches::
+Once this is verified, update your local clone with the upstream branches::
 
    $ git fetch upstream
 
--- a/getting-started/git-boot-camp.rst
+++ b/getting-started/git-boot-camp.rst
@@ -44,11 +44,14 @@ You will only need to do this once.
 
 1. Go to https://github.com/python/cpython.
 
-2. Press ``Fork`` on the top right.
+2. Press ``Fork`` located near the top right of the page.
 
-3. When asked where to fork the repository, choose to fork it to your username.
+   * The required fields ``Owner`` and ``Repository name`` should already be pre-populated with your username and the original repository name "cpython".
+   * Uncheck the "Copy the ``main`` branch only" button.
 
-4. Your forked CPython repository will be created at https://github.com/<username>/cpython.
+3. Press the ``Create fork`` button.
+
+4. Your forked CPython repository will be created at ``https://github.com/<username>/cpython``.
 
 .. _clone-your-fork:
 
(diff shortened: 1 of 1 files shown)


getting-started/git-boot-camp.rst
```diff
@@ -44,11 +44,14 @@ You will only need to do this once.
 
 1. Go to https://github.com/python/cpython.
 
-2. Press ``Fork`` on the top right.
+2. Press ``Fork`` located near the top right of the page.
 
-3. When asked where to fork the repository, choose to fork it to your username.
+   * The required fields ``Owner`` and ``Repository name`` should already be pre-populated with your username and the original repository name "cpython".
```
hugovk, line 49: Maybe we can skip this, it's telling about the defaults.
```suggestion
```

getting-started/git-boot-camp.rst
```diff
@@ -44,11 +44,14 @@ You will only need to do this once.
...
+   * Uncheck the "Copy the ``main`` branch only" button.
```
hugovk, line 50: And we could make this its own step:
```suggestion
3. Uncheck the "Copy the ``main`` branch only" button.
```

getting-started/git-boot-camp.rst
```diff
@@ -44,11 +44,14 @@ You will only need to do this once.
...
 
-4. Your forked CPython repository will be created at https://github.com/<username>/cpython.
+3. Press the ``Create fork`` button.
```
hugovk, line 52:
```suggestion
4. Press the ``Create fork`` button.
```

getting-started/git-boot-camp.rst
```diff
@@ -44,11 +44,14 @@ You will only need to do this once.
...
+
+4. Your forked CPython repository will be created at ``https://github.com/<username>/cpython``.
```
hugovk, line 54:
```suggestion
5. Your forked CPython repository will be created at ``https://github.com/<username>/cpython``.
```

getting-started/git-boot-camp.rst
```diff
@@ -105,6 +108,10 @@ To verify the upstream for ``main``::
 It should emit ``upstream``, indicating to track/pull changes for ``main`` from the
 ``upstream`` remote.
 
+Once this is verified, update your local repository with the upstream branches::
```
hugovk, line 111: We called it a "local clone" elsewhere on this page:
```suggestion
Once this is verified, update your local clone with the upstream branches::
```

hugovk (commented):
Thanks! A few suggestions.

getting-started/git-boot-camp.rst
ezio-melotti: Should we recommend this at all?
Other branches can be always fetched later, if/when needed.
Unchecking this will also make the initial cloning slower, and most contributors won't even need other branches.

kevinbowen777: If this isn't unchecked, then subsequent instructions in the guide will be in error or incorrect.

getting-started/git-boot-camp.rst
kevinbowen777: I'm not sure I agree. While the fields are required, this is explicity stating _how_ the user-specified fields _should_ be to work with the rest of the devguide. 
We don't know, per user, what state will be in (perhaps due to multiple GitHub accounts or previous forks).

ezio-melotti: By default it seems to use the logged (personal?) account, which is what people generally want.

I tried forking a repo that I already forked and I got:
image

I'm not convinced this sentence is needed -- the repo name is always the same (as stated by the fork page), and if the user has multiple accounts and/or belongs to multiple orgs, GitHub will still use the right default or ask the user to pick one.
Keep in mind that we aim to keep the instruction brief and simple whenever possible.

kevinbowen777: Thanks for checking this and verifying that the fields are unlikely to change. I guess it can be dropped without losing clarity.

I'm not attached to this wording as I am to the unchecking the "Copy the main branch only" button. 

I was approaching this with the mindset that if a field/button _could_ be changed by the user/app, it is good to be clear on what is expected in the fields to minimize possible confusion for a new, less experienced user.

I appreciate the feedback!

kevinbowen777:
@hugovk  @ezio-melotti 

Is there anything more needed from me to get this PR merged? I'm not very familiar with how PR reviews work on GitHub. Thanks.

willingc (approved):
Looks good to me @kevinbowen777. I will let @hugovk and @ezio-melotti review that their comments have been incorporated. It looks like they have to me. Thanks!

getting-started/git-boot-camp.rst
```diff
@@ -44,11 +44,13 @@ You will only need to do this once.
 
 1. Go to https://github.com/python/cpython.
 
-2. Press ``Fork`` on the top right.
+2. Press ``Fork`` located near the top right of the page.
 
-3. When asked where to fork the repository, choose to fork it to your username.
+   * Uncheck the "Copy the ``main`` branch only" button.
```
hugovk, line 49: This is an action to be followed, let's promote it to a numbered item, and remember to renumber the following ones:
```suggestion
4. Uncheck the "Copy the ``main`` branch only" button.
```

kevinbowen777: @hugovk I've updated as you've suggested.

hugovk (approved):
Thank you for your contribution! 🚀

PR #1377, opened 2024-08-21
Add further recommendations to the style guide

nedbat:
These are the remaining items from the (private) Editorial Board style guide suggestions:
https://docs.google.com/document/d/1rSkUIGFuI5zDMYturyJCo9agYPQ-d53Z3MntSV0rB3E/edit

----
📚 Documentation preview 📚: https://cpython-devguide--1377.org.readthedocs.build/

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -200,11 +200,12 @@ language):
     descriptor resources are released in a timely manner.
 
 
-Bylines
-=======
+Author attribution
+==================
 
-For new documentation, do not use bylines. They tend to discourage other users
-from updating community documentation.
+For new documentation, do not use a byline (naming the author of the document).
+Explicit attribution tends to discourage other users from updating community
+documentation.
 
 Existing documentation with bylines will not be changed unless the author
 decides to do so. This is subject to change in the future.
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -218,12 +218,13 @@ Avoid first-person language.  Our documentation is owned and maintained by a
 community.  First-person language mistakenly implies a single author.
 
 
-Pronunciation of dunders
-========================
+Pronunciation of dunder names
+=============================
 
-"Dunder methods" like ``__init__`` can be awkward in running prose: is it "an
-__init__" or "a __init__"?  A `quick poll`__ indicates we should use
-"an __init__."
+"Dunder names" like ``__init__`` can be awkward in running prose: is it "an
+init" or "a dunder init"?  Our recommendation is to ignore the underscores and
+use the article that is appropriate for the word in the name.  A `quick poll`__
+backs this up: "an __init__."
 
 __ https://hachyderm.io/@nedbat/112129685322594689
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -211,13 +211,6 @@ Existing documentation with bylines will not be changed unless the author
 decides to do so. This is subject to change in the future.
 
 
-First-person language
-=====================
-
-Avoid first-person language.  Our documentation is owned and maintained by a
-community.  First-person language mistakenly implies a single author.
-
-
 Pronunciation of dunder names
 =============================
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -26,6 +26,7 @@ the footnote reference.
 
 Footnotes may appear in the middle of sentences where appropriate.
 
+
 Capitalization
 ==============
 
@@ -143,6 +144,7 @@ explanation.
 Please consult the `Diátaxis <https://diataxis.fr/>`_ guide for more
 detail.
 
+
 Links
 =====
 
@@ -171,6 +173,7 @@ documentation for ``map``.  You can suppress the link while keeping the
 semantic presentation of the function name by adding an exclamation point
 prefix: ``:func:`!map```.  See :ref:`roles` for more details.
 
+
 Affirmative tone
 ================
 
@@ -196,6 +199,34 @@ language):
     achieve the same effect.  This assures that files are flushed and file
     descriptor resources are released in a timely manner.
 
+
+Bylines
+=======
+
+For new documentation, do not use bylines. They tend to discourage other users
+from updating community documentation.
+
+Existing documentation with bylines will not be changed unless the author
+decides to do so. This is subject to change in the future.
+
+
+First-person language
+=====================
+
+Avoid first-person language.  Our documentation is owned and maintained by a
+community.  First-person language mistakenly implies a single author.
+
+
+Pronunciation of dunders
+========================
+
+"Dunder methods" like ``__init__`` can be awkward in running prose: is it "an
+__init__" or "a __init__"?  A `quick poll`__ indicates we should use
+"an __init__."
+
+__ https://hachyderm.io/@nedbat/112129685322594689
+
+
 Economy of expression
 =====================
 
@@ -206,6 +237,7 @@ It is an unfortunate fact that making documentation longer can be an impediment
 to understanding and can result in even more ways to misread or misinterpret the
 text.  Long descriptions full of corner cases and caveats can create the
 impression that a function is more complex or harder to use than it actually is.
+
 
 Security considerations (and other concerns)
 ============================================
@@ -228,6 +260,7 @@ module (for example, OS level pipe buffers filling up and stalling child process
 these can be documented in a "Common Errors" section and cross-referenced
 rather than repeated for every affected interface.
 
+
 .. _code-examples:
 
 Code examples
@@ -247,6 +280,7 @@ be used sparingly, where it is necessary to clearly differentiate between input
 lines and output lines.  Besides contributing visual clutter, it makes it
 difficult for readers to cut-and-paste examples so they can experiment with
 variations.
+
 
 Code equivalents
 ================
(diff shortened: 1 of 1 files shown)


nedbat:
The changelog entry: https://github.com/python/editorial-board/pull/14

documentation/style-guide.rst
```diff
@@ -196,6 +199,34 @@ language):
     achieve the same effect.  This assures that files are flushed and file
     descriptor resources are released in a timely manner.
 
+
+Bylines
+=======
+
+For new documentation, do not use bylines. They tend to discourage other users
```
pitrou, line 206: It's not clear to me what a "byline" is. Do you mean author attribution?

nedbat: Good point, I've updated it.

pitrou: It's much clearer like this, thank you!

documentation/style-guide.rst
```diff
@@ -196,6 +199,35 @@ language):
     achieve the same effect.  This assures that files are flushed and file
     descriptor resources are released in a timely manner.
 
+
+Author attribution
+==================
+
+For new documentation, do not use a byline (naming the author of the document).
+Explicit attribution tends to discourage other users from updating community
+documentation.
```
AA-Turner, line 208: I'd be interested to see background on this, if possible. I've personally never felt deterred by a byline, and it is a nice way to credit someone who's written a major guide or document. However, perhaps there is a more quantitative study that we could reference if introducing this guidance?

documentation/style-guide.rst
```diff
@@ -196,6 +199,35 @@ language):
...
+
+Existing documentation with bylines will not be changed unless the author
+decides to do so. This is subject to change in the future.
+
+
+First-person language
+=====================
+
+Avoid first-person language.  Our documentation is owned and maintained by a
+community.  First-person language mistakenly implies a single author.
```
AA-Turner, line 218: Would it be nicer to phrase this positively, saying something like:

```suggestion
Grammatical person 
==================

We try to maintain a consistent tone of voice across our documentation.
As part of this, prefer using the third-person when writing reference text and either the second or third person when writing tutorials or guides, which are naturally more 'interactive'.
```

This removes the negative language from the text, and additionally provides better guide rails for authors of different types of text.

documentation/style-guide.rst
```diff
@@ -196,6 +199,35 @@ language):
...
+
+
+Pronunciation of dunders
+========================
```
AA-Turner, line 222:
```suggestion
Pronunciation of dunder methods
===============================
```

documentation/style-guide.rst
```diff
@@ -196,6 +199,35 @@ language):
...
+
+"Dunder methods" like ``__init__`` can be awkward in running prose: is it "an
+__init__" or "a __init__"?  A `quick poll`__ indicates we should use
+"an __init__."
```
AA-Turner, line 226: I'd perhaps phrase this according to the underlying rule that this draws out: that the pronunciation-guided spelling is effectively 'an init method' -- you use the first vowel sound of the word to decide a/an and ignore the underscores.

(Obviously we can also recommend the definite article!)

documentation/style-guide.rst
JelleZijlstra: I do feel uneasy about changing docs that are attributed to a specific author. It feels like I'm putting words in their mouth that they may not agree with.

ezio-melotti: I think it depends on the type of document and the type of change.  For example, updating module docs is easier than updating a howto, since in the howto the original author might have had a more specific vision and goals.

On the other hand, having a byline is also useful to get in touch with the original author and ask for advice and reviews before making major changes (this could technically be replaced by `git blame` and `CODEOWNERS`).

documentation/style-guide.rst
ezio-melotti: Should a note be added about existing documents that use the first-person?
Adding a few examples would be useful too.
Is the first-person plural allowed?  I've seen it used (and used it myself) in tutorials and howtos.

documentation/style-guide.rst
ezio-melotti: Instead of using a question/answer, I would use a statement that says to ignore the underscore and use the article that matches the first sound (like @AA-Turner suggested).  The quick poll could be a footnote.

documentation/style-guide.rst
JelleZijlstra: I like the original language here better than @AA-Turner's suggestion; it is more concise and clear.

AA-Turner: My concrete suggestion was a very quick draft -- the point was more about using constructive language and adding more guidance. Just to clarify that the suggestion is more an opening offer than final wording!

hugovk: Style guides often tell you what to avoid. A quick example:

https://developers.google.com/style/inclusive-documentation

documentation/style-guide.rst
hugovk: I generally agree with being guided by pronunciation, however `__init__` is often pronounced as "dunder init", leading to "a dunder init method".

This makes the pronunciation rule difficult to apply in this case, hence this specific dunder rule.

documentation/style-guide.rst
Mariatta: > it is a nice way to credit someone who's written a major guide or document

How do we credit someone who has written a major code changes? I don't think we add their name to the top of module name, so why would this be any different? This is an open source documentation not a blog.

> having a byline is also useful to get in touch with the original author and ask for advice and reviews before making major changes (this could technically be replaced by git blame and CODEOWNERS)

As you said, this can be done by `git blame` and `codeowners`. However I would like for us to rethink the notion that the original author is the only one capable of reviewing/giving advice. Their advice may be sought, but should not be the only advice that matters, and should not be the only advice we rely on. Other contributors' review and advice should be taken into consideration equally.

documentation/style-guide.rst
nedbat: I left out "methods" because there are a few that are not methods.  How about "dunder names"?

documentation/style-guide.rst
nedbat: Thanks, I've expanded on this paragraph.

AA-Turner: I might be confused here: is the intend to use 'an' for *every* dunder method, or just those that start with a vowel sound? ``init`` is a member of both sets.

Taking an example of a method starting with a consonant sound, would we say "a `__complex__` method" or "an `__complex__` method"? To me, the first sounds right, following the don't-pronounce-underscores rule, but I would argue it also follows the replace-underscores-with-dunder rule.

I missed the original poll as it didn't come up on Discourse, sorry!

A

documentation/style-guide.rst
nedbat: I like the idea of providing examples of good and bad usage.  Also, first-person plural can be useful: "If we change the method like this, ..."

I'll work on a re-wording tomorrow.

documentation/style-guide.rst
AA-Turner: Ignore me, the most recent commit clears this up. Thanks @nedbat!

A

documentation/style-guide.rst
pitrou: My experience in CPython is that personal ownership of areas has sometimes given rise to conflicts when other contributors wanted to meaningfully participate (without always bowing to the original author's preferences). So I'm in favor of this guidline.

documentation/style-guide.rst
nedbat: I've done some more reading and thinking, and I'd like to remove this from this pull request and have more discussion before making a recommendation.

nedbat: BTW, here is a pull request that removes some first-person language: https://github.com/python/cpython/pull/107449/files, with discussion.

documentation/style-guide.rst
willingc: I agree with @pitrou.

willingc: > I do feel uneasy about changing docs that are attributed to a specific author

@JelleZijlstra I am supportive of this for existing documentation. However, this PR is concerning new documentation :smile:

willingc (approved):

documentation/style-guide.rst
willingc: > I'd be interested to see background on this, if possible. I've personally never felt deterred by a byline, and it is a nice way to credit someone who's written a major guide or document. However, perhaps there is a more quantitative study that we could reference if introducing this guidance?

@AA-Turner The byline style recommendation is related to new documentation. We have seen that bylines in certain areas of the documentation has caused conflicts where an individual acts as owner of the community resource. We see similar behavior when we look at academic publishing: https://pubmed.ncbi.nlm.nih.gov/31161378/ 

My general feeling is that reference documentation should not have bylines. As others have mentioned `git blame` or `git log` can offer recognition. If someone creates a new "How to", I think it would be reasonable to recognize in release notes/what's new as we do code contributions.

JelleZijlstra: > I am supportive of this for existing documentation. However, this PR is concerning new documentation

True. We could also consider removing bylines from existing documentation, or replacing them with something more indefinite like "Originally written by Jelle Zijlstra; maintained by the community". However, that's out of scope for this change.

willingc: I like the suggestion @JelleZijlstra for future wording: 
>  "Originally written by Jelle Zijlstra; maintained by the community"

PR #1383, opened 2024-08-31
Python 3.14 now is the main branch

Wulian233:
Also add turtledemo expert (Today new codeowner) https://github.com/python/cpython/commit/1ce9e5880347346105693aba211f4c378f5a9b6a

----
📚 Documentation preview 📚: https://cpython-devguide--1383.org.readthedocs.build/

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -225,6 +225,7 @@ traceback             iritkatriel
 tracemalloc           vstinner
 tty                   Yhg1s*
 turtle                gregorlingl^, willingc
+turtledemo            terryjreedy*
 types                 1st1
 typing                gvanrossum, JelleZijlstra*, AlexWaygood*, carljm, sobolevn*
 unicodedata           malemburg, ezio-melotti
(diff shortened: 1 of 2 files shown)


core-developers/experts.rst
```diff
@@ -225,6 +225,7 @@ traceback             iritkatriel
 tracemalloc           vstinner
 tty                   Yhg1s*
 turtle                gregorlingl^, willingc
+turtledemo            terryjreedy*
```
AA-Turner, line 228: I'm not sure there's much point having two distinct entries?

We should also check that @terryjreedy is content to be listed as a *turtle* expert, as CODEOWNERS and the experts list are not bijective.

Wulian233: > We should also check that @terryjreedy is content to be listed as a *turtle* expert, as CODEOWNERS and the experts list are not bijective.

He said https://github.com/python/cpython/issues/123370#issuecomment-2311954589

terryjreedy: Turtle is to turtledemo somewhat as tkinter is to idlelib.  Yes, I approve of this addition.

PR #1392, opened 2024-09-11
Add guidance about the 'Update branch' button

hugovk:
Sometimes contributors click the "Update branch" button to merge in `main` into their PR, which is fine if there's a good reason, but if not, pings anyone subscribed to the PR, and uses up CI limited resources, which can cause delays on other PRs.

Good reasons to click it include wanting to update an old PR, to make sure it still passes CI. Or if the PR is failing, and the some tests/config has been updated in `main`.

I occasionally need to tell contributors this in PRs. Let's add a section to the devguide so we have something handy to link to. Here's some proposed text, happy to change it.

Preview: https://cpython-devguide--1392.org.readthedocs.build/getting-started/pull-request-lifecycle/#update-branch-button

### DIFF
--- a/getting-started/pull-request-lifecycle.rst
+++ b/getting-started/pull-request-lifecycle.rst
@@ -542,6 +542,18 @@ If you identify such flaky behavior, look for an issue in the `issue tracker`_
 that describes this particular flakiness. Create a new issue if you can't
 find one.
 
+:guilabel:`Update branch` button
+================================
+
+You can click on the :guilabel:`Update branch` button to merge the latest
+changes from the base branch (usually ``main``) into the PR.
+This is useful to :ref:`keep the CI green <keeping-ci-green>` for old PRs,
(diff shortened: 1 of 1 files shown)


getting-started/pull-request-lifecycle.rst
```diff
@@ -542,6 +542,18 @@ If you identify such flaky behavior, look for an issue in the `issue tracker`_
 that describes this particular flakiness. Create a new issue if you can't
 find one.
 
+:guilabel:`Update branch` button
+================================
+
+You can click on the :guilabel:`Update branch` button to merge the latest
```
JelleZijlstra, line 548: We might want to add something like: If the PR is very old, it may be useful to update the branch before merging to ensure that the PR does not fail any CI checks that were added or changed since CI last ran.

We occasionally get issues if we have old PRs that fail a newly added check.

JelleZijlstra (approved):

Issue #1315, opened 2024-05-01
Update description of CPython docs publish frequency


hugovk:
**Describe the bug**

https://devguide.python.org/documentation/devguide/ says:

> Our devguide workflow uses continuous integration and deployment so changes to the devguide are normally published when the pull request is merged. Changes to CPython documentation follow the workflow of a CPython release and are published in the release.

**Expected behavior**

The last sentence isn't the whole story; the docs are also updated about once a day. Let's update it to just mention the daily update.

hugovk:
Fixed by https://github.com/python/devguide/pull/1395.

PR #1395, opened 2024-09-11
Clarify the documentation update process

miguelcsx:
- update the documentation to specify that cpython docs are updated daily

- removed the previous implication that updates occur only with cpython releases

- clarified the connection between daily updates and the cpython release workflow

### Description

This PR updates the documentation to clarify the update workflow for the CPython documentation. The previous text implied that updates only occurred with CPython releases, but the documentation is actually updated on a daily basis. This change ensures accurate and up-to-date information about the documentation workflow.

### Changes Made

- Revised the description of the CPython documentation update process.

### Reason for Change

The original text did not fully reflect the frequency of CPython documentation updates. By specifying that updates occur daily, this pull request provides a more accurate representation of the documentation workflow.

### Impact

This change will ensure that users are correctly informed about the update schedule for CPython documentation, enhancing transparency and understanding of the documentation process.

----
📚 Documentation preview 📚: https://cpython-devguide--1395.org.readthedocs.build/

### DIFF
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -19,10 +19,12 @@ main Python documentation, except for some small differences.  The source
 lives in a `separate repository`_ and bug reports should be submitted to the
 `devguide GitHub tracker`_.
 
-Our devguide follows a continuous integration and deployment workflow, with changes published when pull requests are merged.
-CPython documentation is updated regularly, typically within a day or two, depending on the build process.
-The documentation is versioned for each CPython release, and updates are continuosly incorporated, though with some delay due the build time.
-Aditionally, each CPython release includes a corresponding version of the documentation, which may also be used by redistributors.
+Changes to the Developer's Guide are published when pull requests are merged.
+
+Changes to the Python documentation are published regularly,
+ususally within 48 hours of the change being committed.
+The documentation is also `published for each release <https://docs.python.org/release/>`_,
+which may also be used by redistributors.
 
 
 Developer's Guide workflow
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -19,10 +19,8 @@ main Python documentation, except for some small differences.  The source
 lives in a `separate repository`_ and bug reports should be submitted to the
 `devguide GitHub tracker`_.
 
-Our devguide workflow uses continuous integration and deployment so changes to
-the devguide are normally published when the pull request is merged. Changes
-to CPython documentation follow the workflow of a CPython release and are
-published in the release.
+Our devguide follows a continuous integration and deployment workflow, with changes published when pull requests are merged.
+CPython documentation, on the other hand, is updated daily and aligns with the CPython release workflow, with changes being included in each release.
 
 
 Developer's Guide workflow
--- a/documentation/devguide.rst
+++ b/documentation/devguide.rst
@@ -20,7 +20,9 @@ lives in a `separate repository`_ and bug reports should be submitted to the
 `devguide GitHub tracker`_.
 
 Our devguide follows a continuous integration and deployment workflow, with changes published when pull requests are merged.
-CPython documentation, on the other hand, is updated daily and aligns with the CPython release workflow, with changes being included in each release.
+CPython documentation is updated regularly, typically within a day or two, depending on the build process.
+The documentation is versioned for each CPython release, and updates are continuosly incorporated, though with some delay due the build time.
+Aditionally, each CPython release includes a corresponding version of the documentation, which may also be used by redistributors.
 
 
 Developer's Guide workflow


hugovk:
Thanks for the PR.

However, the CPython docs aren't updated daily.

Short version: it's currently taking over 40 hours.

Longer version:

What happens is: there's an hourly cron on the docs build server. If there's a build currently running, that build continues, and the hourly check stops. If there's no build running, a new one starts.

It will then check each active branch (currently 3.12-3.14) and all the translations. 

If there are docs changes since last run, or changes to translations, that version/language will be rebuilt.

It takes a long time to build all these docs, especially the PDFs. Recently it's been taking over 40 hours.

By the time a full build has finished, because it took so long, inevitably there are new changes on the next loop, requiring a full build yet again.

See https://github.com/python/docsbuild-scripts/issues/190 for efforts to improve this.

Okay, so that's one part of it. The other is releases. Docs are built during releases, and can be found here: https://www.python.org/doc/versions/

hugovk:
And I realise I'm contradicting what I wrote in https://github.com/python/devguide/issues/1315, but I've learned a lot the docs build process in recent months :)

So perhaps if we hedge and say something like "about once a day"? 

(I'm hoping further work at https://github.com/python/docsbuild-scripts/issues/190 and especially https://github.com/python/docs-community/issues/5 will mean we can get the _HTML docs_ updated much sooner after update, perhaps even within an hour.)

miguelcsx:
> And I realise I'm contradicting what I wrote in #1315, but I've learned a lot the docs build process in recent months :)
> 
> So perhaps if we hedge and say something like "about once a day"?
> 
> (I'm hoping further work at [python/docsbuild-scripts#190](https://github.com/python/docsbuild-scripts/issues/190) and especially [python/docs-community#5](https://github.com/python/docs-community/issues/5) will mean we can get the _HTML docs_ updated much sooner after update, perhaps even within an hour.)

@hugovk Ok, I'll update to make sure it's clear to understand, thanks for the clarification.

documentation/devguide.rst
```diff
@@ -19,10 +19,8 @@ main Python documentation, except for some small differences.  The source
 lives in a `separate repository`_ and bug reports should be submitted to the
 `devguide GitHub tracker`_.
 
-Our devguide workflow uses continuous integration and deployment so changes to
-the devguide are normally published when the pull request is merged. Changes
-to CPython documentation follow the workflow of a CPython release and are
-published in the release.
+Our devguide follows a continuous integration and deployment workflow, with changes published when pull requests are merged.
+CPython documentation, on the other hand, is updated daily and aligns with the CPython release workflow, with changes being included in each release.
```
JelleZijlstra, line 23: The "aligns with the CPython release workflow" wording is pretty vague and if I didn't already know how it works, it wouldn't tell me much. The docs are versioned, in that there is a separate documentation for each CPython version, and the online documentation is updated continually (but with a delay as Hugo discussed). Each CPython release also comes with documentation, which may be used by redistributors.

miguelcsx: @hugovk @JelleZijlstra I made some changes aiming to clarify the delay and versioning at this [commit](https://github.com/python/devguide/pull/1395/commits/34dc1ce25fd13e3e0d15d8f5dd47f919e3f2bacf).
I look forward to suggestions for improvement, thank you for your support.

documentation/devguide.rst
```diff
@@ -19,10 +19,10 @@ main Python documentation, except for some small differences.  The source
...
 `devguide GitHub tracker`_.
 
-Our devguide workflow uses continuous integration and deployment so changes to
-the devguide are normally published when the pull request is merged. Changes
-to CPython documentation follow the workflow of a CPython release and are
-published in the release.
+Our devguide follows a continuous integration and deployment workflow, with changes published when pull requests are merged.
+CPython documentation is updated regularly, typically within a day or two, depending on the build process.
+The documentation is versioned for each CPython release, and updates are continuosly incorporated, though with some delay due the build time.
+Aditionally, each CPython release includes a corresponding version of the documentation, which may also be used by redistributors.
```
AA-Turner, line 25: Perhaps as a shorter text?:
```suggestion
Changes to the Developer's Guide are published when pull requests are merged.

Changes to the Python documentation are published regularly,
ususally within 48 hours of the change being committed.
The documentation is also `published for each release <https://docs.python.org/release/>`_,
which may also be used by redistributors.
```

JelleZijlstra: I like that; I like the clear separation between the devguide and CPython in your text.

miguelcsx: Ok, I'm going to modify it, to make it short and have a clear separation, thanks for the suggestions.

AA-Turner: @miguelcsx you should be able to click the 'commit' button on the suggestion to apply it -- you can of course make further amendments should you wish!

A

miguelcsx: @AA-Turner I was thinking of something like this, to clarify the delay, do you think it's necessary?
```md
Changes to the Developer's Guide are published when pull requests are merged.
Changes to the Python documentation are published regularly, usually within 48 hours of the change being committed, though some delay may occur due to the build process.
The documentation is also published for each release <https://docs.python.org/release/>_, which may also be used by redistributors."
```

AA-Turner: Personally I wouldn't say it's needed given we're using a fairly conservative figure (48 hours) for the update cycle -- the longest we've seen in practice is ~40 hours, and we've made significant progress on reducing it.

A

miguelcsx: @AA-Turner Ok, thanks for following the PR, and reviewing it, I committed your changes, again thanks a lot

AA-Turner:
Thanks @miguelcsx 

A

PR #1407, opened 2024-09-23
Cleanup "Reporting an issue" section

### LINKED ISSUE #1403: "Reporting an issue" section is out of sync with actual github options
Current text:
https://github.com/python/devguide/blob/8fcb7a5ccc96751299cb2e8bf5c8ad4fa8eaeb59/triage/issue-tracker.rst?plain=1#L57-L69

Instead we have available:
* Bug report
* Crash report
* Documentation
* Feature or enhancement
* Report a security vulnerability
* Getting help
* Proposing new features

Last two point to d.p.o.

I'm not sure if devguide should list all available options, rather ones that always be present and/or relevant for following text.  Say:
```
 First you need to select what kind of problem you want to report. The 
 available choices include, for example: 
  
 * **Bug report**: an existing feature isn't working as expected; 
 * **Documentation**: there is missing, invalid, or misleading documentation; 
 * **Feature or enhancement**: suggest a new feature for Python; 
```

skirpichev:
Now this lists only few examples from actually available options.

Closes #1403

----
📚 Documentation preview 📚: https://cpython-devguide--1407.org.readthedocs.build/


AA-Turner (approved):

ezio-melotti (approved):

Mariatta (commented):
I think we should still list all the available choices instead of only a sampling.
Though I'm ok with removing the last two (getting help/proposing new features)

JelleZijlstra:
I feel the full list is likely to go out of date as we change the issue template, and duplicating it here isn't very useful as users can see the same information by going to https://github.com/python/cpython/issues/new/choose.

skirpichev:
> I think we should still list all the available choices instead of only a sampling.
> Though I'm ok with removing the last two (getting help/proposing new features)

I think that 1st sentence contradicts to the 2nd.

> I feel the full list is likely to go out of date as we change the issue template

This increase maintenance burden for no real benefits.   Why do we want here list *all* (well, not all actually even with @Mariatta suggestion;)) currently available options?  They seems to be just examples and not referenced somehow in following text.

willingc (approved):

triage/issue-tracker.rst
```diff
@@ -55,18 +55,11 @@ box above the list of bugs. If you're not already signed in to GitHub, it
 will ask you to do so now.
 
 First you need to select what kind of problem you want to report. The
-available choices are:
-
-* **Bug report**: an existing feature isn't working as expected;
-* **Documentation**: there is missing, invalid, or misleading documentation;
-* **Enhancement**: suggest a new feature for Python;
-* **Performance**: something should work faster;
-* **Security**: there is a specific kind of weakness open to exploitation
```
willingc, line 64: Add back security item.

skirpichev: Done.

willingc:
Thanks @skirpichev.

PR #1413, opened 2024-09-25
Document ``next`` in versionadded & similar directives

encukou:
See https://github.com/python/cpython/issues/121277

----
📚 Documentation preview 📚: https://cpython-devguide--1413.org.readthedocs.build/

### DIFF
--- a/documentation/markup.rst
+++ b/documentation/markup.rst
@@ -1083,8 +1083,11 @@ units as well as normal text:
    (``class``, ``attribute``, ``function``, ``method``, ``c:type``, etc),
    a ``versionadded`` should be included at the end of its description block.
 
-   The first argument must be given and is the version in question.  The second
-   argument is optional and can be used to describe the details of the feature.
+   The first argument must be given and is the version in question.
+   Instead of a specific version number, you can---and should---use
+   the word ``next``, indicating that the API will first appear in the
+   upcoming release.
+   The second argument is optional and can be used to describe the details of the feature.
 
    Example::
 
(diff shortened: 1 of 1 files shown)


documentation/markup.rst
```diff
@@ -1083,16 +1083,42 @@ units as well as normal text:
...
+         .. versionadded:: 3.14
+
+   The tool to do this replacement is `update_version_next.py`_
+   in the release-tools repository.
+
+   .. _update_version_next.py: https://github.com/python/release-tools/blob/master/update_version_next.py
+
+   When backporting to versions before 3.14, check if ``Doc/tools/extensions/pyspecific.py``
+   contains the function ``expand_version_arg``. If it's not there,
+   use a specific version instead of ``next``.
```
ezio-melotti, line 1117: It would be better to just decide which versions `next` should be backported to, backport it, and then list those versions here.

Can you also clarify when would backporting be necessary for a `versionadded`?  New features should be documented on the `next` release, and not backported to previous versions.

The only exception I can think of is if we are documenting an addition after the fact, but this is covered in the next paragraph.

`versionchanged` are more likely to be backported, so maybe this paragraph should be moved there.

documentation/markup.rst
```diff
@@ -1106,7 +1132,7 @@ units as well as normal text:
 
          Return foo and bar, optionally with *spam* applied.
 
-         .. versionchanged:: 3.6
+         .. versionchanged:: next
             Added the *spam* parameter.
 
    Note that there should be no blank line between the directive head and the
```
ezio-melotti, line 1138: Should this have a line like "See `versionadded` for more info on `next`."?

documentation/markup.rst
```diff
@@ -1118,22 +1144,25 @@ units as well as normal text:
...
 
 .. describe:: deprecated-removed
 
    Like ``deprecated``, but it also indicates in which version the feature is
    removed.
 
    There are two required arguments: the version from which the feature is
-   deprecated, and the version in which the feature is removed.
+   deprecated (usually ``next``), and the version in which the feature
+   is removed, which must be a specific version number (*not* ``next``).
```
ezio-melotti, line 1161: Does the tool that convert `next` check this?

encukou: Not explicitly, but it will fail.
For a check with a nice error message, could you review https://github.com/python/cpython/pull/124623?

It's not a big priority, since `next` doesn't really make sense here.

documentation/markup.rst
ezio-melotti: To clarify, with the first sentence I meant the support for `next` itself.  IOW, if `next` is supported from e.g. 3.12+, then the docs should simply say something like "When backporting to versions before 3.12, use a specific version instead of `next`.".  If you are planning to use the current wording until the support for `next` has been backported and then update this paragraph accordingly, that's ok (assuming that `next` does the right thing when used on older versions).

For security fixes and similar I was thinking that are more likely to be `versionchanged`, but sometimes new APIs are indeed added.  Since the docs for `versionchanged` start with "Similar to `versionadded`", I now think it's ok to leave the paragraph here.

encukou: Oh! Yes, I expect to change the wording after backporting the feature, and I plan to talk to RMs after this is merged and announced.

Issue #1420, opened 2024-09-27
Add a note about Windows long paths support


itamaro:
this is a gotcha I stumbled upon when setting up a Windows buildbot worker.

let's document that long paths support should be enabled on a Windows worker.

ezio-melotti:
This was fixed in #1421.  Feel free to reopen if there is anything else left to do.

PR #1421, opened 2024-09-27
gh-1420: Add a note about long paths support to Windows buildbot setup

itamaro:
----
📚 Documentation preview 📚: https://cpython-devguide--1421.org.readthedocs.build/


zware:
I'm torn on this one :).  It's good information to add, but ideally our tests should handle the 'disabled' case, since it's the default.

itamaro:
> It's good information to add, but ideally our tests should handle the 'disabled' case, since it's the default.

why not both? :)

if the test suite handles the disabled case (likely by skipping some tests when long paths are disabled), then we probably want to nudge buildbot worker maintainers to enable lang paths to get more comprehensive coverage.

zware:
Fair enough.

PR #1424, opened 2024-09-29
Change reference in GC design doc

Privat33r-dev:
Prior to the change, the link led to a fixed version 3.8. Currently, it's not the latest version.

As the time passes by, it's going to be outdated further and further.

It might be the best solution to refer to the latest dev version of the doc, this way we can ensure that reader will access the latest data.

----
📚 Documentation preview 📚: https://cpython-devguide--1424.org.readthedocs.build/

### DIFF
--- a/internals/garbage-collector.rst
+++ b/internals/garbage-collector.rst
@@ -167,7 +167,7 @@ C APIs
 
 Specific APIs are offered to allocate, deallocate, initialize, track, and untrack
 objects with GC support. These APIs can be found in the `Garbage Collector C API
-documentation <https://docs.python.org/3.8/c-api/gcsupport.html>`_.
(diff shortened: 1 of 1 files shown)


ezio-melotti (approved):
LGTM.
`3` could also be used instead of `dev`, but `dev` is also ok.

Privat33r-dev:
> LGTM. `3` could also be used instead of `dev`, but `dev` is also ok.

I was thinking that the last dev version might be better, since it's a guide for CPython developers, thus it's best if they see docs in the latest state.

Also I am thinking to make a regex search to detect and change Python Doc links from "3.X" to "dev" or at least to "3" version.

willingc:
@Privat33r-dev The garbage collection doc has been moved to the CPython repo under Internals. If you wish to make a change, please do so against that repo. Thanks!

PR #1428, opened 2024-10-04
gh-124989: remove exception related details which have moved to cpython/InternalsDoc

iritkatriel:
See https://github.com/python/cpython/pull/124989


AA-Turner (approved):
Do you intend to move the location table text over in a future PR?

A

iritkatriel:
> Do you intend to move the location table text over in a future PR?
> 
> A

Yes, but I think there already is something about it so it might be a matter of merging old and new.

Ultimately all the internals stuff should move from dev guide to the cpython docs.

PR #1431, opened 2024-10-08
gh-119786: Move parser doc from devguide to InternalDocs

iritkatriel:
First merge https://github.com/python/cpython/pull/125119.

----
📚 Documentation preview 📚: https://cpython-devguide--1431.org.readthedocs.build/


JacobCoffee (approved):

ezio-melotti (approved):
I wonder if we can use rediraffe or something similar to (eventually) get rid of this page altogether and redirect to the CPython repo directly.

JacobCoffee:
@ezio-melotti what if we just updated the link altogether in https://github.com/iritkatriel/devguide/blob/remove_parser_doc/internals/index.rst then we could remove the file.

```rst
===================
CPython's internals
===================

This guide describes the basics of CPython's internals.
It explains the layout of CPython's source code.
It also explains how the parser, compiler, and interpreter
work together to run your Python code.
Finally, it covers the garbage collector and how it manages memory.

.. toctree::
   :maxdepth: 3

   exploring
   parser <https://github.com/python/cpython/blob/main/InternalDocs/parser.md>
   compiler
   interpreter
   garbage-collector
```

ezio-melotti:
There might be external websites linking to this page, and removing it altogether will result in a 404.  This is also why we adopted Rediraffe after we reorganized the devguide to group similar pages in different sections.

OTOH it feels a bit weird to have a redirect that goes outside the devguide, so leaving an almost-empty page is probably ok unless someone has a better idea.

PR #1442, opened 2024-10-13
Recomend `podman` instead of `docker` and remove mentions to Docker images

### LINKED ISSUE #1441: Replace container images to packages from cpython-devcontainers
### Describe the bug

When I wrote the devguide for adding extensions modules, I assumed that we'll continue using Ubuntu public images: https://devguide.python.org/developer-workflow/extension-modules/#make-regen-configure-and-missing-permissions. 

However, this is no more the case and we are using a GHCR package (see https://github.com/python/cpython/issues/124612). I don't think it was actually necessary to show how to do a manual configuration. I think using `podman` is better than using `docker` so I'll rather recommend the former. At the same time, I'll also be able to remove the small Sphinx extension that I wrote to simplify upgrading the image.

### Screenshots

_No response_

### Additional context

_No response_

picnixz:
Fixes #1441.

This also simplifies the instructions, removing the systemd part (because honestly, I think it could be more confusing than helping if someone is *not* using systemd or does not know how to use it...). Using podman is much simpler and if there's any issue we can help them directly in their PR.

----
📚 Documentation preview 📚: https://cpython-devguide--1442.org.readthedocs.build/


picnixz:
Concerning the creation of files owned by the root user by the docker service, this is something **I** encountered but not always. AFAIK, podman does not any super-user permissions and would not create files owned by root (I honestly don't know what happened).

developer-workflow/extension-modules.rst
```diff
@@ -593,34 +593,14 @@ by executing :cpy-file:`Tools/build/regen-configure.sh`:
 ``make regen-configure`` and missing permissions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Since this rule requires `Docker <https://docs.docker.com/desktop>`_ to be
-running, the following can be done on Linux platforms (``systemctl``-based):
```
ezio-melotti, line 597: I think this section should:
1. clarify if `regen-configure` works with both `podman` and `docker`;
2. clarify which one takes the precedence;
3. recommend to (install and) use `podman`, since it should be simpler thank `docker`
4. link to the SO question in case the user is running into permission problems with `docker`

You might also want to update the section title to something more generic.

picnixz: This section was more for troubleshooting since those are the issues I encountered actually :') I can however say that `podman` takes precedence. This section was really for the specific issue of missing permissions (which was solved via the SO post). I think it's better if each issue has its own section together with its resolution (it helps when you encounter some issue). 

I'll put another section which recommends podman in general and explains that podman and docker are interchangeable modulo those weird docker issues.

developer-workflow/extension-modules.rst
```diff
@@ -593,34 +593,14 @@ by executing :cpy-file:`Tools/build/regen-configure.sh`:
...
-   # pull the Docker image if needed
-   docker image pull ubuntu:$CONFIGURE_UBUNTU_VERSION$
-
-.. tip::
-
-   If the issue persists, you may try `podman <https://podman.io/>`_.
-   The commands for listing or pulling an image are the same as ``docker``.
+We also recommend using `podman <https://podman.io/docs/installation>`_
+instead of ``docker`` since the former does not require a background
+service, or does not create files owned by the ``root`` user in some
```
hugovk, line 602:
```suggestion
We also recommend using `Podman <https://podman.io/docs/installation>`_
instead of Docker since the former does not require a background
service, and does not create files owned by the ``root`` user in some
```

developer-workflow/extension-modules.rst
picnixz: I've decided to take your approach, thereby simplifying the troubleshooting section (and make it only relevant for Docker-based configurations, not Podman ones).

developer-workflow/extension-modules.rst
picnixz: Thank you, I've tweaked a bit your formulation!

developer-workflow/extension-modules.rst
```diff
@@ -590,38 +601,13 @@ by executing :cpy-file:`Tools/build/regen-configure.sh`:
...
-
-   systemctl status docker          # is the Docker service running?
-   sudo systemctl start docker      # start it if it is not
-   sudo systemctl restart docker    # or restart it if the issue persists
+``make regen-configure`` and missing permissions with Docker
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If Docker complains about missing permissions, this Stack Overflow post
 could be useful in solving the issue: `How to fix docker: permission denied
 <https://stackoverflow.com/q/48957195/9579194>`_.
```
ezio-melotti, line 609:
```suggestion
<https://stackoverflow.com/q/48957195/9579194>`_.  Alternatively, you can
try using `Podman <https://podman.io/docs/installation>`_.
```
It might still be worth to mention `podman` here, since technically it's also a solution to the problem.

picnixz: It's a solution in the sense "X doesn't work but Y works! try Y!" (and I like those kind of solutions if they are simpler :D)

ezio-melotti (approved):
FTR you can commit review suggestions directly from GitHub, either individually or in a single batch from the "files changed" tab.

picnixz:
> FTR you can commit review suggestions directly from GitHub, either individually or in a single batch from the "files changed" tab.

Yes I know, but I didn't commit your suggestion because I wanted to split the line somewhere else (just to make it more or less aligned with the rest of the paragraph). I also avoid to use the GH UI if I'm on my dev env since otherwise I need to remember to pull from my remote branch :')

PR #1444, opened 2024-10-15
NIS optional module dependency

Abgael:
NIS optional module build seems to fail if libnsl is missing. Installing libnsl-dev on Ubuntu/Debian or libnsl2-devel on Fedora based distros fixes the dependency issue.

----
📚 Documentation preview 📚: https://cpython-devguide--1444.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -638,7 +638,7 @@ some of CPython's modules (for example, ``zlib``).
             gcc gcc-c++ gdb lzma glibc-devel libstdc++-devel openssl-devel \
             readline-devel zlib-devel libffi-devel bzip2-devel xz-devel \
             sqlite sqlite-devel sqlite-libs libuuid-devel gdbm-libs perf \
-            expat expat-devel mpdecimal python3-pip
+            expat expat-devel mpdecimal python3-pip libnsl2-devel
 
 
    On **Debian**, **Ubuntu**, and other ``apt``-based systems, try to get the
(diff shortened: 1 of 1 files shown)


skirpichev:
You can see above apt install & friends:  "To install *some* additional dependencies for optional build and test components".

I.e. this is not an exhaustive list.  Is there some reason why libnsl should be listed too?

Abgael:
I don't think it necessarily should. It is more like a suggestion. The reason I am suggesting this is that with the guide given dependencies, the building process was able to build every single optional module, being NIS the only one to fail. Doing some search on the error message, (“The necessary bits to build these optional modules were not found:
nis”) I find out that other people bumped into the same problem.

itamaro:
generally I'm in favor of having the optional dependencies as comprehensive as possible. the main reason for that, from my point of view at least, is that this is what I follow when setting up a buildbot worker, so having it as comprehensive as possible means that the buildbot will provide better coverage.

I'm unsure about this specific case, considering [the nis module was removed](https://peps.python.org/pep-0594/#nis) in 3.13, so I wouldn't want stale optional dependencies listed.
it's a bit of a grey area since nis is still available up to 3.12, and we still run buildbots for 3.9+ branches.
my instinct would be to leave it out since it's deprecated and removed, but I wouldn't object to including it if others think it's valuable.

encukou:
Thank you for the report, @Abgael!
As Itamar points out, the `nis` module was removed in 3.13. It was for [good reasons](https://peps.python.org/pep-0594/#nis). If you're building Python (any version) *today*, especially to help develop it, you're most likely better off without the `nis` module. If not, you probably know `nis`/`libnsl2` and don't need the devguide to tell you what to do.
(For Buildbots: IMO, the existing fleet should keep `nis` covered until 3.12 goes EOL. As a back-up, Fedora builds and tests `nis` on 3.12 & below.)

With that, I'll close the issue. I hope it doesn't discourage you from contributing in the future!

PR #1458, opened 2024-10-27
Fix link in interpreter.rst

ndunnett:
The link to internal docs currently points to [compiler.md](https://github.com/python/cpython/blob/main/InternalDocs/compiler.md), I've changed it to point to [interpreter.md](https://github.com/python/cpython/blob/main/InternalDocs/interpreter.md) as I assume was intended.

----
📚 Documentation preview 📚: https://cpython-devguide--1458.org.readthedocs.build/

### DIFF
--- a/internals/interpreter.rst
+++ b/internals/interpreter.rst
@@ -5,4 +5,4 @@ The bytecode interpreter
 ========================
 
 This document is now part of the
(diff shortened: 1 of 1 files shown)


ezio-melotti (approved):

iritkatriel (commented):
Thanks

PR #1461, opened 2024-10-31
Contrib guide: more restructuring

nedbat:
I'm continuing to look at the overall structure and move things around.  Comments welcome.  I wanted to flesh out the Code and Docs landing pages before making a real pull request, but maybe it's ok to merge this?

----
📚 Documentation preview 📚: https://cpython-devguide--1461.org.readthedocs.build/


willingc (approved):
Thank you @nedbat!

Looks great. Two minor comments but don't hold merge on them:
- Perhaps center or move the links into the titles of the landing page table
- On the code contributions communications page, flip flop Discourse to come before mailing lists.

:tada:

contrib/code/index.rst
```diff
@@ -1,3 +1,5 @@
+.. _c_code:
```
hugovk, line 1: This made me think of C the language.

Are these `c_` prefixes short for "contrib" and temporary whilst still a draft in parallel with the old devguide?

contrib/index.rst
```diff
@@ -1,28 +1,116 @@
...
 
+.. raw:: html
+
+    <script>
+    document.addEventListener('DOMContentLoaded', function() {
+      activateTab(getOS());
+    });
+    </script>
+    <style>
+      table.docutils td { vertical-align: top; }
```
hugovk, line 15: Perhaps we should move this to `_static/devguide_overrides.css`?

contrib/index.rst
```diff
@@ -1,28 +1,116 @@
...
+
+*[The original table on the devguide home had a fourth column for Core
+Developers. That made the table wider and more confusing. I don't think core
+team members need a quick intro path since they will have been through the
+devguide before.]*
+
+*[I haven't adjusted the links in the table yet other than to add a link to the
+major section at the top of each column.]*
+
+.. list-table::
```
hugovk, line 70: I'm trying to think which is better, lists inside a table, or smaller cells, from an accessibility point of view, and I'm not sure :)

The new table of lists might be better, because there's no relation between the old cells that happen to be next to each other; that is, the old way was just for presentation and not structure. So I think the list might be better?

contrib/index.rst
```diff
@@ -1,28 +1,116 @@
...
+       * :ref:`help`
+       * :ref:`pullrequest`
+       * :ref:`runtests`
+       * :ref:`fixingissues`
+       * :ref:`communication`
+       * :ref:`gitbootcamp`
+       * :ref:`devcycle`
+
+[The contents had been listed below here. It was very long, and I don't think
+it was helpful.  The left sidebar has the contents.]
```
hugovk, line 109: Hmm, I found it helpful: I often cmd+F on the homepage to find something, which won't work with a collapsed left sidebar.

It is a little unusual to have a large table-of-contents on a page, although it is out of the way at the bottom of the page, and not so big as to be a performance problem (like, we definitely wouldn't want the whole CPython TOC).

But it's not the end of the world to remove it if no one else wants it :)

contrib/intro/index.rst
```diff
@@ -10,9 +12,26 @@ Introduction
...
+a series of excellent guides at `Open Source Guides
+<https://opensource.guide/>`_.
+
+Anyone will find the following guides useful:
+
+* `How to Contribute to Open Source <https://opensource.guide/how-to-contribute/>`_
+* `Building Welcoming Communities <https://opensource.guide/building-community/>`_
+
+
 Healthy Collaboration
```
hugovk, line 35: Sentence case titles:
```suggestion
Healthy collaboration
```

And the others further down the page.

contrib/index.rst
willingc: I think it is fine to leave at the very bottom.

contrib/code/index.rst
nedbat: Yes, exactly.  I maybe should have explained that.

contrib/index.rst
nedbat: I didn't know about that file.  I think tables should default to top-aligned, but idk if this will affect other things badly as a global setting.

contrib/index.rst
nedbat: Other people I've talked with definitely felt like the one-link-per-cell style of the current home page table is misleading because the rows have no meaning.

contrib/intro/index.rst
nedbat: Done.

contrib/index.rst
nedbat: I can leave it.

nedbat: I put it back, but not as deep as before, so it's not so long.

nedbat:
@willingc the list has Repos, Discourse, Discord, Mailing lists, so I'm not sure what you are looking for.

nedbat:
@willingc I also don't understand this:
> Perhaps center or move the links into the titles of the landing page table

nedbat:
Oh, "move the links into the titles," I see what you mean.  I've changed it to this:
image

nedbat:
And another option:

image

hugovk (approved):

willingc:
> @willingc the list has Repos, Discourse, Discord, Mailing lists, so I'm not sure what you are looking for.

I wanted to order the list to roughly reflect where conversations happen today. I wanted to de-emphasize the mailing lists relative to Discourse and Repos.

PR #1464, opened 2024-11-03
Fix current admins table

savannahostrowski:
Before:
Screenshot 2024-11-03 at 2 19 16 PM

After:
Screenshot 2024-11-03 at 2 19 07 PM

----
📚 Documentation preview 📚: https://cpython-devguide--1464.org.readthedocs.build/


hugovk:
Thanks!

PR #1466, opened 2024-11-04
contrib: restructure Getting Started - Setup and Building into a Workflows section

nedbat:
The Getting Started section has a lot of technical detail.  I thought it
would be better moved to the end, and referenced from the contribution
paths earlier in the guide.  I haven't updated the paths yet.

Also:

- I added a central links.rst that can be included anywhere.
- I omitted the "Editors and tools" section. It doesn't provide any useful information.
- Perhaps the Workflows section could be further categorized.
- The `c_blah_start` and `c_blah_end` comments are only for the temporary inclusions being used to demonstrate the new organization.

----
📚 Documentation preview 📚: https://cpython-devguide--1466.org.readthedocs.build/


nedbat:
Hmm, not sure how `ned@geometer.lan` was the author of a few commits, but I've fixed them.

Issue #1463, opened 2024-11-02
Feature: Add refleak checker option under testing


willingc:
### Describe the enhancement or feature you would like

`python -m test XXX -R` enables refleak checking during testing.

See https://github.com/python/cpython/pull/126103#discussion_r1826521226

Lincoln-developer:
Hey @willingc I would love to work on this , Thanks

PR #1468, opened 2024-11-05
gh-1463: Added Refleak Checker Option under testing.

Lincoln-developer:
This pull request adds documentation for the -R flag in the python -m test command, enabling reference leak checking. This feature helps developers identify memory leaks when modifying C code.

----
📚 Documentation preview 📚: https://cpython-devguide--1468.org.readthedocs.build/

### DIFF
--- a/testing/run-write-tests.rst
+++ b/testing/run-write-tests.rst
@@ -86,8 +86,8 @@ if the failures are transient or consistent.
 The ``-uall`` flag allows the use of all available
 resources so as to not skip tests requiring, e.g., Internet access.
 
-To check for reference leaks (only needed if you modified C code), 
-you can enable reference leak checking during testing using the ``-R`` flag.  
+To check for reference leaks (only needed if you modified C code),
+you can enable reference leak checking during testing using the ``-R`` flag.
 For example, using the command::
 
     python -m test <test_name> -R <warmups>:<repeats>
--- a/testing/run-write-tests.rst
+++ b/testing/run-write-tests.rst
@@ -90,6 +90,14 @@ To check for reference leaks (only needed if you modified C code),
 you can enable reference leak checking during testing using the ``-R`` flag.
 For example, using the command::
 
+    python -m test <test_name> -R :
+
+This default setting performs a few initial warm-up runs to stabilize the reference count,
+followed by additional runs to check for leaks.
+
+If you want more control over the number of runs, you can specify ``warmups`` and ``repeats`` explicitly
+For example::
+
     python -m test <test_name> -R <warmups>:<repeats>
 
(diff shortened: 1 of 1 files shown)


encukou:
Thank you for the addition!
When first learning about `-R`, you aren't likely to know what values to use for `<warmups>` and `<repeats>`. So, I think it would be good if the first paragraph, and the example, used the defaults, which are specified (somewhat unusually) with a lone colon after the `-R`: `python -m test <test_name> -R :`

A follow-up paragraph can then mention specifying `<warmups>` and `<repeats>` explicitly (which would help explain what the `:` is for).

Lincoln-developer:
> Thank you for the addition! When first learning about `-R`, you aren't likely to know what values to use for `<warmups>` and `<repeats>`. So, I think it would be good if the first paragraph, and the example, used the defaults, which are specified (somewhat unusually) with a lone colon after the `-R`: `python -m test <test_name> -R :`
> 
> A follow-up paragraph can then mention specifying `<warmups>` and `<repeats>` explicitly (which would help explain what the `:` is for).

Thank @encukou for the insight sighted out, let me work on these changes to have a more clear view of this context.

testing/run-write-tests.rst
```diff
@@ -86,10 +86,24 @@ if the failures are transient or consistent.
...
+This default setting performs a few initial warm-up runs to stabilize the reference count,
+followed by additional runs to check for leaks.
+
+If you want more control over the number of runs, you can specify ``warmups`` and ``repeats`` explicitly
+For example::
+
+    python -m test <test_name> -R <warmups>:<repeats>
+
+This enables the refleak checker option, allowing you to perform warm-up runs
+to stabilize reference counts followed by additional runs to verify any leaks.
```
encukou, line 104: This info is duplicated.
```suggestion
```

testing/run-write-tests.rst
```diff
@@ -86,10 +86,24 @@ if the failures are transient or consistent.
...
+you can enable reference leak checking during testing using the ``-R`` flag.
+For example, using the command::
+
+    python -m test <test_name> -R :
+
+This default setting performs a few initial warm-up runs to stabilize the reference count,
+followed by additional runs to check for leaks.
+
+If you want more control over the number of runs, you can specify ``warmups`` and ``repeats`` explicitly
+For example::
```
encukou, line 99: Rather than an example, this is a generic command with placeholders.
```suggestion
If you want more control over the number of runs, you can specify ``warmups`` and ``repeats`` explicitly::
```

Lincoln-developer:
> Sorry for dropping the ball here! I'll merge with a few small adjustments. Thank you for the contribution, @Lincoln-developer!

Thank you for reviewing and merging this pull request.

PR #1470, opened 2024-11-12
Add advice on copyright notices and PRs

hugovk:
Closes https://github.com/python/cpython/issues/126133.

Re: https://github.com/python/cpython/issues/126133#issuecomment-2451271955, this is so we can link to this section if we get PRs attempting to update copyright years each year.

Feedback welcome on wording and where to put this in the devguide.

----
📚 Documentation preview 📚: https://cpython-devguide--1470.org.readthedocs.build/

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -1,5 +1,3 @@
-import time
-
 extensions = [
     'notfound.extension',
     'sphinx.ext.extlinks',
@@ -16,7 +14,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = "Python Developer's Guide"
-copyright = f'2011-{time.strftime("%Y")}, Python Software Foundation'
+copyright = '2011 Python Software Foundation'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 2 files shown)


malemburg:
You should also add a note that for most code in the PSF CPython distribution, the PSF does not own the copyright, but only had redistribution rights to that code, so randomly applying the PSF copyright notice is often plain wrong.

There were a few copyright assignments in the past, but I'm not sure whether we really have a complete list of those. Only code which falls under one of these assignment is legally copyrighted by the PSF.

Note that the PSF does own the copyright on the CPython distributions, i.e. the compilation of the files which make up CPython, which is why there is a copyright statement in the README and the LICENSE.

hugovk:
Thanks for the info.

I'd prefer to keep this particular change scoped only about the existing notices, not about adding any new notices, or how copyright applies. It's primarily so we have something to point at if we get PRs adding new years to the list.

Please could you open a new issue or PR if you'd like to extend it? Thank you!

willingc (approved):

conf.py
```diff
@@ -16,7 +14,7 @@
 
 # General information about the project.
 project = "Python Developer's Guide"
-copyright = f'2011-{time.strftime("%Y")}, Python Software Foundation'
+copyright = '2011 Python Software Foundation'
```
willingc, line 17: Clarifying that this date applies to the devguide only.

willingc (approved):

PR #1472, opened 2024-11-13
Add Savannah to Experts Index

savannahostrowski:

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -56,7 +56,7 @@ __future__
 __main__              gvanrossum, ncoghlan
 _thread
 abc
-argparse
+argparse              savannahostrowski*
 array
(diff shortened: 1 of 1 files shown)


core-developers/experts.rst
```diff
@@ -56,7 +56,7 @@ __future__
 __main__              gvanrossum, ncoghlan
 _thread
 abc
-argparse
+argparse              savannahostrowski*
```
AA-Turner, line 59: Should Serhiy be here too? (@serhiy-storchaka)
```suggestion
argparse              savannahostrowski*, serhiy-storchaka
```

savannahostrowski: I'll leave that up to him! :)

zware (approved):

PR #1492, opened 2025-01-15
Shorten names of country-specific languages

m-aciek:
The “Brasilian Portuguese” is commonly used as a language name, also it looks like the Hindi and Bengali doesn't need to be country-specific: Hindi is India-only language, Bengali is [the same](https://www.reddit.com/r/kolkata/comments/1e5exb1/guys_whats_the_difference_between_indian_bengali/) in both India and Bangladesh.

----
📚 Documentation preview 📚: https://cpython-devguide--1492.org.readthedocs.build/

### DIFF
--- a/documentation/translating.rst
+++ b/documentation/translating.rst
@@ -72,7 +72,7 @@ details.
    * - Portuguese (pt)
      - Gustavo Toffo
      -
-   * - `Brasilian Portuguese (pt-br) <https://docs.python.org/pt-br/>`__
+   * - `Brazilian Portuguese (pt-br) <https://docs.python.org/pt-br/>`__
      - Marco Rougeth
      - :github:`GitHub <python/python-docs-pt-br>`,
        `wiki <https://python.org.br/traducao/>`__,
--- a/documentation/translating.rst
+++ b/documentation/translating.rst
@@ -23,7 +23,7 @@ details.
    * - Arabic (ar)
      - Abdur-Rahmaan Janhangeer (:github-user:`Abdur-rahmaanJ`)
      - :github:`GitHub <Abdur-rahmaanJ/python-docs-ar>`
-   * - Indian Bengali (bn_IN)
+   * - Bengali (bn_IN)
      - Kushal Das (:github-user:`Kushal997-das`)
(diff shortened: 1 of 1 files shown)


StanFromIreland:
Reference: https://en.wikipedia.org/wiki/Brazilian_Portuguese

StanFromIreland (approved):

documentation/translating.rst
```diff
@@ -23,7 +23,7 @@ details.
    * - Arabic (ar)
      - Abdur-Rahmaan Janhangeer (:github-user:`Abdur-rahmaanJ`)
      - :github:`GitHub <Abdur-rahmaanJ/python-docs-ar>`
-   * - Bengali as spoken in India (bn_IN)
+   * - Indian Bengali (bn_IN)
```
StanFromIreland, line 26:
```suggestion
   * - Bengali — India (bn_IN)
```

Maybe simply this for Bengali?

documentation/translating.rst
```diff
@@ -34,7 +34,7 @@ details.
        Fanis Petkos (:github-user:`thepetk`),
        Panagiotis Skias (:github-user:`skpanagiotis`)
      - :github:`GitHub <pygreece/python-docs-gr>`
-   * - Hindi as spoken in India (hi_IN)
+   * - Indian Hindi (hi_IN)
```
StanFromIreland, line 37:
```suggestion
   * - Hindi (hi_IN)
```

India seems to be the only country to have it as an official language. I see no need to specify in this case/

https://www.worldatlas.com/articles/hindi-speaking-countries.html

m-aciek:
It looks that Bengali language is almost the same in India and Bangladesh. https://www.reddit.com/r/kolkata/comments/1e5exb1/guys_whats_the_difference_between_indian_bengali/

I think we should talk with Kushal about removing the country code from it to let it cover both countries.

StanFromIreland (approved):
This looks good to me

rffontenelle (changes requested):

rffontenelle (approved):

AA-Turner (approved):

PR #1497, opened 2025-01-18
conf.py: Update copyright date

StanFromIreland:
Current:
image

`2011 Python Software Foundation`

Proposed:
image

`2011-2025 Python Software Foundation`

The content of most of the pages has been updated since 2011. This is consistent with the format of the docs.python.org footer:
image

----
📚 Documentation preview 📚: https://cpython-devguide--1497.org.readthedocs.build/

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -14,7 +14,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = "Python Developer's Guide"
-copyright = '2011 Python Software Foundation'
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks, but we don't need to update the copyrights, see:

* https://devguide.python.org/getting-started/pull-request-lifecycle/#copyrights
* https://github.com/python/devguide/pull/1470
* https://github.com/python/cpython/issues/126133

PR #1501, opened 2025-01-23
Update Antoine's affiliation

pitrou:
Also remove mention of mentioning multiprocessing, because that's not true anymore.

----
📚 Documentation preview 📚: https://cpython-devguide--1501.org.readthedocs.build/


hugovk (approved):
Thanks!

PR #1516, opened 2025-02-18
Describe the memorialization procedure

ambv:
----
📚 Documentation preview 📚: https://cpython-devguide--1516.org.readthedocs.build/


core-developers/memorialization.rst
```diff
@@ -0,0 +1,132 @@
...
+* The user's "custom status" is set to 🕊 ``in memoriam``;
+* The user's "about me" is amended with ``$firstname passed away on $date. [In memoriam.]($in_memoriam_post_url)``;
+* In the user's security "recently used devices" the staff member chooses "Log out all";
+* In the user's permissions the staff member chooses "Deactivate account";
+* The user's trust level is reset to ``1: basic user`` (trust level 0 doesn't allow links in "About Me");
+* The user's "associated accounts" (like GitHub) that provide an alternative
+  login method, are all disconnected;
+* The user's API keys are revoked;
+* The user's admin or moderator right is revoked;
+* The user's primary email address is reset to ``username@in-memoriam.invalid`` and
```
hugovk, line 67: To follow the pattern below.
```suggestion
* The user's primary email address is reset to ``USERNAME@in-memoriam.invalid`` and
```

core-developers/memorialization.rst
```diff
@@ -0,0 +1,134 @@
...
+SSH server access
+-----------------
+
+* The user is removed from Salt configuration for the PSF infrastructure
+  in `/pillar/base/users <https://github.com/python/psf-salt/tree/main/pillar/base/users>`_.
+
+PyPI
+----
+
+* The PSF staff member notifies PyPI admins to mark the user as inactive,
```
sethmlarson, line 132: Can mention `admin@pypi.org` email address for this use-case.

core-developers/memorialization.rst
```diff
@@ -0,0 +1,134 @@
...
+
+To respect the choices that someone made while alive, we aim to preserve
+content of their accounts without changes after they've passed away.
+To support the bereaved, in some instances, we may remove or change
+certain content when the legacy contact or family members request it.
+
+GitHub
+------
+
+* The user is removed from the `python/ <https://github.com/orgs/python/>`_
```
sethmlarson, line 33: I'm not sure how it works, but I wouldn't be surprised if removing someone from an organization doesn't remove them from a GitHub Environment or GitHub Security Advisory where they are referenced by username (as opposed to being a part of a team).

sethmlarson (approved):
Thank you for creating this Łukasz, I'm not sure if buildbots should have their own section here? Then I had two other comments, otherwise LGTM

core-developers/memorialization.rst
ambv: So far I checked and the person that this process was documented for was in fact removed from all places where they were referred to by name as well, since the only way to be referenced was to either be a team member or an "outside collaborator" on the organization. I'll leave it at that for now, if we find the next time that we need refinement, we'll look into that.

core-developers/memorialization.rst
ambv: Good point. Made the change. Thanks for the review!

ambv:
Added Discord and Buildbot. Landing.

core-developers/memorialization.rst
```diff
@@ -0,0 +1,146 @@
...
+
+python.org email account
+------------------------
+
+The PSF staff member emails ``postmaster@python.org`` to ask the email
+administrator to:
+
+* remove SMTP access from ``USERNAME@python.org``;
+* reset the password to POP3/IMAP for ``USERNAME@python.org``;
+* disable email forwarding, if set up, for ``USERNAME@python.org``;
```
cjw296, line 87: @ambv - apologies to being late in the game on this, but only noticed this PR this morning.
Removing email forwarding like this may cause problems for the deceased's family or estate executors:
- if folks are only communicating with the deceased through their `python.org` address, their family or estate executors may have a harder time if this suddenly stops working
- confusion may crop up if the address ends up being re-used; I've hard problems in the past where various email addresses and social media handles have ended up being reused after a family member passed away.

core-developers/memorialization.rst
```diff
@@ -0,0 +1,146 @@
...
+access to the systems listed below is revoked and some changes are
+made to how the user displays to others.
+
+To respect the choices that someone made while alive, we aim to preserve
+content of their accounts without changes after they've passed away.
+To support the bereaved, in some instances, we may remove or change
+certain content when the legacy contact or family members request it.
+
+GitHub
+------
```
ezio-melotti, line 31: There might also be tokens created by the user for specific services (GH actions, bots, hooks).  Maybe a paragraph should be added about checking for (and possibly remove/replace them) those too?

core-developers/memorialization.rst
```diff
@@ -0,0 +1,146 @@
...
+* In the user's security "recently used devices" the staff member chooses "Log out all";
+* In the user's permissions the staff member chooses "Deactivate account";
+* The user's trust level is reset to ``1: basic user`` (trust level 0 doesn't allow links in "About Me");
+* The user's "associated accounts" (like GitHub) that provide an alternative
+  login method, are all disconnected;
+* The user's API keys are revoked;
+* The user's admin or moderator right is revoked;
+* The user's primary email address is reset to ``USERNAME@in-memoriam.invalid`` and
+  secondary email addresses are removed (this step requires the administrator
+  to contact Discourse.org staff via ``team@discourse.org``)
```
ezio-melotti, line 71: Minor nit, this list item is missing the `.`.

ezio-melotti (commented):
Thanks for taking care of this.

core-developers/memorialization.rst
ambv: I'll add a note to avoid name reuse forever.

As for keeping the email address alive forever, I'm afraid this is out of the question due to security concerns. Removing the ability to send email (SMTP) and receive email (POP3/IMAP/forwarding) via the PSF-provided address after the user passed away is done to disallow password resets and other forms of impersonation.

Per the [python.org email policy](https://www.python.org/psf/records/board/policies/email/), that email address is provided on a best-effort basis, which suggests it should not be used by its users as a primary communication channel.

I will add a note that in case the email shutdown causes issues for the estate executors, the PSF will reasonably try to help if contacted directly.

core-developers/memorialization.rst
ambv: I agree in principle, but I don't know how to actually enforce it. For OAuth app tokens, it's impossible to remove a token only for a single person. We rather clean this via disconnecting the user's GitHub from their Discourse account and logging them out, etc.

For GitHub app tokens and private keys requested by a concrete user, I don't know how to vet this other than clicking through every GH app on every repository. What do we expect to do in this case? Invalidate every token? I expect that the user who created such a private key or token only used it to put it in the GH app's secrets and was not otherwise stored by them locally, but of course that might be wrong.

core-developers/memorialization.rst
ambv: Fixed in #1517.

core-developers/memorialization.rst
ambv: Addressed in #1517.

core-developers/memorialization.rst
ezio-melotti: I just checked the secrets on a repo and there doesn't seem to be any indication of who generated/added the token, and indeed it shouldn't matter too much who does it.

Maybe this is an issue about documenting who manages these services (and is therefore in charge of managing the tokens), so that someone else can take over if needed.  (Some of these are already documented [in the devguide](https://devguide.python.org/developer-workflow/development-cycle/#current-owners).)

sethmlarson: Yeah that was my worry too, that tokens don't tell you who generated them without consulting the audit log. Agreed on documenting who manages the services, I think that's sufficient along with tokens auto-expiring and refreshing.

core-developers/memorialization.rst
cjw296: Totally understand the SMTP and POP3/IMAP side of things, although for the latter, if it's easy enough to do, might be helpful to send an archive to family/executors.

If an address is just forwarding, what's the harm? I appreciate it's best efforts basis :-)

Issue #1286, opened 2024-03-06
Document how to add a bytecode specialization


Fidget-Spinner:
How to add a new bytecode specialization is knowledge only a few of us remember right now. We should document this somewhere (devguide or CPython, I don't mind). Hopefully a new contributor can pick this up!

There reference is this PR: https://github.com/python/cpython/pull/116385/files

Assuming you found an instruction that serves as a good specialization candidate. Let's use the example of `CONTAINS_OP`.
1. Set the original `CONTAINS_OP` to a uop by changing it from `CONTAINS_OP` to `_CONTAINS_OP` in `Python/bytecodes.c`. The instruction definition should also change from `inst` to `op`.
2. Add the uop that calls a specializing function. In this case `_SPECIALIZE_CONTAINS_OP`.
3. The original `CONTAINS_OP` is now a new macro consisting of `_SPECIALIZE_CONTAINS_OP + _CONTAINS_OP`.
4. Define the cache structure in `Include/internal/pycore_code.h`. At the very least, a 16-bit counter is needed.
5. Write the specializing function itself in `Python/specialize.c`. Refer to any other function in that file for the format.

willingc:
Hi @AbudiMutamba. Welcome :wave: Before you get too far along on this. Let's check with @Fidget-Spinner if this is still something that is needed?

Fidget-Spinner:
Yes we still need this.

willingc:
Thanks @Fidget-Spinner. If you want to give this a try @AbudiMutamba, please do so.

faizanoor3001:
Hello, @willingc is it something still open ? Would like to contribute. :)

Fidget-Spinner:
@faizanoor3001 this is still open

faizanoor3001:
Hello, @Fidget-Spinner,  I’ve opened a PR adding “How to Add a New Bytecode Specialization” to the How-To section. We can verify the text changes or any updates are required, but I’m unsure about the best place for this new guide. Would it fit better in:


1.  [Interpreter.md section](https://github.com/python/cpython/blob/main/InternalDocs/interpreter.md ) - appended under the existing InternalDocs?
2. CPython Internals — as its own subsection?

      **Current:**
      ![Image](https://github.com/user-attachments/assets/5b1083ff-3709-48a2-8c3b-8e809f48263f)

      **Potential Alternative:**
      
[...]

PR #1522, opened 2025-02-24
gh-1286 : Document how to add a bytecode specialization

faizanoor3001:
## Solves: #1286 
# Description
- Add a "How to Add a New Bytecode Specialization” section in the dev guide.
    - This guide explains how to add a new bytecode specialization, using 
    ``CONTAINS_OP`` as an example.

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -2213,3 +2213,90 @@ and update your unit tests to reflect the new behaviour.
    If you forget to update your input block during the alpha and beta phases,
    the compiler warning will turn into a compiler error when the
    release candidate phase begins.
+
+How to add a new bytecode specialization
+----------------------------------------
+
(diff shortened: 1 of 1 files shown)


Fidget-Spinner:
Sorry, could you please open a PR at CPython and document it here instead? https://github.com/python/cpython/blob/main/InternalDocs/interpreter.md

I'll close this PR, as it seems the trend is to document internal docs in CPython itself not the devguide.

faizanoor3001:
> Sorry, could you please open a PR at CPython and document it here instead? https://github.com/python/cpython/blob/main/InternalDocs/interpreter.md
> 
> I'll close this PR, as it seems the new trend is to document internal docs in CPython itself not the devguide.

Yes, that's what I was wondering, as it looks like a recent update, will create a new one. No worries.

PR #1529, opened 2025-03-12
Python version status: show when bugfix releases become security releases.

nedbat:
This adds a yellow security tail to the green bugfix bars:
image

The transition date is a fixed two years after the first release.  The release schedule PEPs don't indicate a specific date.

----
📚 Documentation preview 📚: https://cpython-devguide--1529.org.readthedocs.build/versions/

### DIFF
--- a/_tools/release_cycle_template.svg.jinja
+++ b/_tools/release_cycle_template.svg.jinja
@@ -58,25 +58,87 @@
         <!-- Colourful blob with a label -->
         {% set start_x = date_to_x(version.first_release_date) %}
         {% set end_x = date_to_x(version.end_of_life_date) %}
-        {% set mid_x = (start_x + end_x) / 2 %}
-        <rect
-            class="release-cycle-blob release-cycle-blob-{{ version.status }}"
-            x="{{ start_x * SCALE }}"
-            y="{{ (y - 1) * SCALE }}"
-            width="{{ (end_x - start_x) * SCALE }}"
-            height="{{ 1.25 * SCALE }}"
-            rx="0.25em"
-            ry="0.25em"
-        />
-        <text
-            class="release-cycle-blob-label release-cycle-blob-{{ version.status }}"
-            x="{{ mid_x * SCALE }}"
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
I feel this makes it more confusing. The color of each bar is supposed to represent the branch's current status, not its status at some particular time. For example, your version makes it seem like 3.11 was always in security-only mode, instead of marking it as bugfix for the first few years.

We could make an alternative design where we switch fully to this PR's approach, but that would require for example that we color everything after the EOL as red for end-of-life on every branch.

_tools/release_cycle_template.svg.jinja
```diff
@@ -58,25 +58,87 @@
...
+            <path
+                class="release-cycle-blob release-cycle-blob-bugfix"
+                d="
+                    M{{ left_x + radius_value }},{{ recty }}
+                    Q{{ left_x }},{{ recty }} {{ left_x }},{{ recty + radius_value }}
+                    V{{ recty + height - radius_value }}
+                    Q{{ left_x }},{{ recty + height }} {{ left_x + radius_value }},{{ recty + height }}
+                    H{{ middle_x }}
+                    V{{ recty }}
+                    Z
```
encukou, line 86: I find this rather confusing.
Once the final style is decided, would you mind me simplifying & adding comments? Unless you want to do it of course.

merwok:
Jelle has a good point. When you think about it, the original design is a little strange in combining a time chart and current status in label and colour.

Python’s current timeline shows status relative to time of viewing.
Another way is Django’s timeline, with time-relative status and a current date line: django release timeline

On the other hand, the change proposed is useful. Maybe it would be less jarring if the other 3 labels with «security» also showed a «bugfix» segment? Then it would be uniform for the releases that are not EOL or feature.

merwok:
Can you show the results of your version Petr?

encukou:
I'm updating my branch at times; the commit `3a4fe83` renders as:

image

<details>

<summary>with the full chart below the tables (click to expand)</summary>

image


</details>

PR #1531, opened 2025-03-17
Updated "Status of Python versions" page

nedbat:
This includes changes by me and @encukou.  Suggestions welcome.

How would you describe the difference between "feature fixes" and "bug fixes"?

----
📚 Documentation preview 📚: https://cpython-devguide--1531.org.readthedocs.build/versions/


_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
+    def __init__(self, limit_to_active=False, special_py27=False) -> None:
         with open("include/release-cycle.json", encoding="UTF-8") as in_file:
             self.versions = json.load(in_file)
 
         # Generate a few additional fields
         for key, version in self.versions.items():
             version["key"] = key
-            version["first_release_date"] = parse_date(version["first_release"])
+            version["first_release_date"] = r1 = parse_date(version["first_release"])
+            version["start_security_date"] = r1 + dt.timedelta(days=2 * 365)
```
hugovk, line 39: It's only two years for 3.13+:

> Note: 2 years of full support start with Python 3.13. Python versions 3.9 - 3.12 operate on a calendar with 1½ year of full support, followed by 3½ more years of security fixes.

https://peps.python.org/pep-0602/#years-of-full-support-3-more-years-of-security-fixes

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
 class Versions:
     """For converting JSON to CSV and SVG."""
 
-    def __init__(self) -> None:
+    def __init__(self, limit_to_active=False, special_py27=False) -> None:
```
hugovk, line 31: Let's make them keyword-only, to require the name and prevent calling `Versions(True, False)`

(We already do call with keywords 👍)

```suggestion
    def __init__(self, *, limit_to_active=False, special_py27=False) -> None:
```

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
+            version["start_security_date"] = r1 + dt.timedelta(days=2 * 365)
             version["end_of_life_date"] = parse_date(version["end_of_life"])
+
+        self.cutoff = min(ver["first_release_date"] for ver in self.versions.values())
+
+        if limit_to_active:
+            self.cutoff = min(
+                version["first_release_date"]
+                for version in self.versions.values()
+                if version["status"] != 'end-of-life'
```
hugovk, line 48: consistency nit
```suggestion
                if version["status"] != "end-of-life"
```

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
+            )
+            self.versions = {
+                key: version
+                for key, version in self.versions.items()
+                if version["end_of_life_date"] >= self.cutoff
+                or (special_py27 and key == '2.7')
```
hugovk, line 54:
```suggestion
                or (special_py27 and key == "2.7")
```

ned-deily (commented):
The `prerelease` phase description LGTM, thanks.

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
+            }
+            if special_py27:
+                self.cutoff = min(self.cutoff, dt.date(2019, 8, 1))
+            self.id_key = 'active'
+        else:
+            self.id_key = 'all'
```
hugovk, line 60:
```suggestion
            self.id_key = "active"
        else:
            self.id_key = "all"
```

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
             self.versions.values(),
             key=lambda v: [int(i) for i in v["key"].split(".")],
             reverse=True,
         )
 
+        # Set the row (y-coordinate) for the chart, to allow a gap between 2.7
+        # and the rest
+        y = len(self.sorted_versions) + (1 if special_py27 else 0)
+        for version in self.sorted_versions:
+            if special_py27 and version["key"] == '2.7':
```
hugovk, line 72:
```suggestion
            if special_py27 and version["key"] == "2.7":
```

_tools/generate_release_cycle.py
```diff
@@ -28,21 +28,52 @@ def parse_date(date_str: str) -> dt.date:
...
+        else:
+            self.id_key = 'all'
+
         self.sorted_versions = sorted(
             self.versions.values(),
             key=lambda v: [int(i) for i in v["key"].split(".")],
             reverse=True,
         )
 
+        # Set the row (y-coordinate) for the chart, to allow a gap between 2.7
```
hugovk, line 68: We have "X coordinate" elsewhere
```suggestion
        # Set the row (Y coordinate) for the chart, to allow a gap between 2.7
```

_tools/generate_release_cycle.py
```diff
@@ -145,8 +177,13 @@ def main() -> None:
     args = parser.parse_args()
 
     versions = Versions()
+    print(versions.versions.keys())
```
hugovk, line 180: Debug or want to keep this?
```suggestion
```

_tools/release_cycle_template.svg.jinja
```diff
@@ -42,40 +48,150 @@
...
-        </text>
+    {% for version in versions %}
+        {% set y = version.y * LINE_HEIGHT %}
 
         <!-- Colourful blob with a label -->
         {% set start_x = date_to_x(version.first_release_date) %}
         {% set end_x = date_to_x(version.end_of_life_date) %}
-        {% set mid_x = (start_x + end_x) / 2 %}
+
+        <!-- bugfix status needs a security tail.
```
hugovk, line 83:
```suggestion
        <!-- Bugfix status needs a security tail.
```

_tools/release_cycle_template.svg.jinja
```diff
@@ -42,40 +48,150 @@
...
+            Draw the rectangle with two path elements instead.
+            Thanks Claude.ai for the conversion.
```
hugovk, line 85: Needed?
```suggestion
```

_tools/release_cycle_template.svg.jinja
```diff
@@ -42,40 +48,150 @@
...
+        -->
+        {% set half_x = [end_x, date_to_x(version.start_security_date)]|min %}
+        {% set height = 1.25 * SCALE %}
+        {% set left_width = (half_x - start_x) * SCALE %}
+        {% set right_width = (end_x - half_x) * SCALE %}
+        {% set left_x = start_x * SCALE %}
+        {% set middle_x = half_x * SCALE %}
+        {% set right_x = half_x * SCALE %}
+        {% set recty = (y - 1) * SCALE %}
```
hugovk, line 94: Little odd var name, maybe `rect_y` instead? Or what is this?
```suggestion
        {% set recty = (y - 1) * SCALE %}
```

make.ps1
```diff
@@ -64,7 +64,8 @@ if ($target -Eq "clean") {
     $ToClean = @(
         $BUILDDIR,
         $_VENV_DIR,
-        "include/branches.csv", "include/end-of-life.csv", "include/release-cycle.svg"
+        "include/branches.csv", "include/end-of-life.csv",
+        "include/release-cycle.svg",  "include/release-cycle-all.svg"
```
hugovk, line 68:
```suggestion
        "include/release-cycle.svg", "include/release-cycle-all.svg"
```

versions.rst
```diff
@@ -10,13 +10,12 @@ branch that accepts new features.  The latest release for each Python
...
 
 
-Python release cycle
-====================
-
 .. raw:: html
    :file: include/release-cycle.svg
 
-Another useful visualization is `endoflife.date/python <https://endoflife.date/python>`_.
+(See :ref:`below <versions-chart-all>` for a chart with older versions.
```
hugovk, line 16: Perhaps link to the existing `#full-chart` already created by the heading?

It'll be nicer to share URLs like https://devguide.python.org/versions/#full-chart instead of https://devguide.python.org/versions/##versions-chart-all

versions.rst
```diff
@@ -40,17 +39,36 @@ Unsupported versions
...
+Python releases go through five phases:
+
+:feature: Before the first beta, the next full release can accept new features,
+   bug fixes, and security fixes.
+
+:prerelease: After the first beta, no new features can go in, but feature fixes
+   (including significant changes to new features), bug fixes, and security fixes
+   are accepted for the upcoming feature release.
+
+:bugfix: Once a version has been released, bug fixes and security fixes are
```
hugovk, line 63: Perhaps, to distinguish from alpha releases etc?
```suggestion
:bugfix: Once a version has been fully released, bug fixes and security fixes are
```

versions.rst
```diff
@@ -40,17 +39,36 @@ Unsupported versions
...
+   accepted. New binaries are built and released. (Also called **maintenance**
+   mode or **stable** release)
```
hugovk, line 65:
```suggestion
   mode or **stable** release.)
```

versions.rst
```diff
@@ -40,17 +39,36 @@ Unsupported versions
...
+
+:security: After two years, only security fixes are accepted and no more
```
hugovk, line 67: "two years" is only for 3.13+

hugovk (commented):
Mostly nits.

_tools/generate_release_cycle.py
nedbat: Should we add an actual security date to the JSON?  The PEPs don't have dates for this transition.

nedbat: Another option: hardcode the 1.5-year vs 2-year delta based on the version number.

nedbat: I chose the hard-coded option.

versions.rst
nedbat: Change to "a few"

_tools/generate_release_cycle.py
nedbat: removed.

_tools/release_cycle_template.svg.jinja
nedbat: Changed.

versions.rst
nedbat: I don't see a real difference between the two URLs (your second one accidentally has two hashes though).

encukou:
Here's the promised cleanup: https://github.com/nedbat/devguide/pull/1

_tools/release_cycle_template.svg.jinja
```diff
@@ -42,40 +48,149 @@
...
+        />
+        <rect
+            x="{{ (LEGEND_WIDTH - RIGHT_MARGIN) * SCALE }}"
+            y="0"
+            width="{{ RIGHT_MARGIN * SCALE }}"
+            height="{{ diagram_height * SCALE }}"
+            fill="url(#release-cycle-mask-gradient-{{ id_key }})"
+        />
+        <rect
+            x="{{ (LEGEND_WIDTH ) * SCALE }}"
```
ezio-melotti, line 68:
```suggestion
            x="{{ LEGEND_WIDTH * SCALE }}"
```

versions.rst
```diff
@@ -40,17 +39,36 @@ Unsupported versions
    :file: include/end-of-life.csv
 
 
+.. _versions-chart-all:
+
+Full chart
+==========
+
+.. raw:: html
+   :file: include/release-cycle-all.svg
```
ezio-melotti, line 48: I find the page structure a bit confusing, since it has two charts and two tables, with some level of overlapping between them but otherwise no further symmetry:
* first there's the chart with 2.7 + 3.6-3.14
* then there's a table with 3.9-3.14 (supported versions)
* then there's another table with 2.6-2.7 + 3.0-3.8 (unsupported versions)
* then there's the full chart with 2.6-2.7 + 3.0-3.14 (all versions)

In particular:
* Is there a reason why the first chart includes 2.7/3.6-3.8 even though they are unsupported?  Perhaps this could become the chart of supported versions only?
* It might be less confusing if both tables were under the same header ("Version support"?), with either a `<caption>` or a colspanning `<th>` on top of each table for "supported"/"unsupported" (or even joined together, with the colspanning `<th>`s)

This can also be addressed in a separate PR if you prefer.

versions.rst
ezio-melotti: To me `a few` seems both vague and higher than two.
Maybe just say `After 2 years (1.5 for versions <3.13), ...`?

Another option is keeping `After two years` and adding a footnote that explains that before 3.13 it was 1.5 and than it became 2, possibly linking the relevant PEPs/discussion.

_tools/generate_release_cycle.py
hugovk: Hard-coding is good for now.

I'm thinking about having all the release dates in a reusable JSON file, but that's for the future:

https://github.com/python/pythondotorg/issues/2701#issuecomment-2730267006

versions.rst
hugovk: There's a UX inconsistency: clicking "See below" gives one, clicking the header gives the other, for no obvious benefit.

I suggest we remove the `versions-chart-all` reference and just use `full-chart`.

(Thanks, removed the second hash.)

nedbat: Done.

versions.rst
nedbat: Updated, with some other tweaks.

_tools/release_cycle_template.svg.jinja
nedbat: This is obsoleted by Petr's clean-up.

nedbat:
I think everything has been addressed.

_tools/release_cycle_template.svg.jinja
```diff
@@ -33,59 +39,159 @@
...
+                x="{{ (start_x + middle_x) / 2 }}"
+                text-anchor="middle"
+            {% elif version.status == "security" %}
+                x="{{ (middle_x + end_x) / 2 }}"
+                text-anchor="middle"
+            {% elif version.status == "end-of-life" %}
+                x="{{ end_x + (0.25 * SCALE) }}"
+                text-anchor="start"
+            {% else %}
+                x="{{ start_x - (0.5 * SCALE) }}"
```
encukou, line 170: Missed this one:
```suggestion
                x="{{ start_x - (0.25 * SCALE) }}"
```

_tools/release_cycle_template.svg.jinja
```diff
@@ -33,59 +39,159 @@
...
+        {% set height = 1.25 * SCALE %}
+        {% set start_x = date_to_x(version.first_release_date) %}
+        {% set end_x = date_to_x(version.end_of_life_date) %}
+        {% set radius = 0.25 * SCALE %}
+
+        {% set small_text_y = version.y * line_height - 0.1 * SCALE %}
+
+        <!-- bugfix/security blobs need to be split between the two phases.
+            Draw the rectangle with two path elements instead.
+            Thanks Claude.ai for the initial conversion.
```
hugovk, line 89: Unrevert
```suggestion
        <!-- Bugfix/security blobs need to be split between the two phases.
            Draw the rectangle with two path elements instead.
```

hugovk (approved):
Thanks all!

versions.rst
encukou: This structure is deliberate. It keeps the tables together, and the least-useful info (the full chart) at the end, and a heavy distinction between active and EOL branches.

> Perhaps this could become the chart of supported versions only?

When we get tired of telling people that 2.7 is EOL, we set the `special_py27` argument to False :)
That'll still show versions that went EOL during an active branch's lifetime. That should (for 3.8+) match the recommended deprecation period per PEP 387.

_tools/release_cycle_template.svg.jinja
encukou: Claude served its role -- it allowed Ned to get started -- even if its code was messy and basically none of it remains.

encukou:
Thank you for pushing this through, Ned!

versions.rst
ezio-melotti: > It keeps the tables together

This is good, since they complement each other -- that's why I suggested putting them in the same subsection too.

> the least-useful info (the full chart) at the end, and a heavy distinction between active and EOL branches.

Removing EOL branches from the first chart would make the distinction even clearer IMHO.

> When we get tired of telling people that 2.7 is EOL

I don't think we need it in the first chart for that -- the fact that it's not even included should give them a hint, and if that's not enough they can find 2.7 in the table and in the full chart.

PR #1532, opened 2025-03-18
use page titles as link text to avoid warnings

nedbat:
Previously we had scary red messages:
```
devguide/core-developers/memorialization.rst:5: WARNING: Duplicate explicit target name: "here". [docutils]
devguide/core-developers/memorialization.rst:5: WARNING: Duplicate explicit target name: "here". [docutils]
```

----
📚 Documentation preview 📚: https://cpython-devguide--1532.org.readthedocs.build/

### DIFF
--- a/core-developers/memorialization.rst
+++ b/core-developers/memorialization.rst
@@ -42,7 +42,7 @@ cancellation as this action is reserved for next-of-kin or designated by
 the deceased GitHub user to act as an account successor.
 
 The general policy regarding deceased users on GitHub is described on their
-`GitHub Deceased User Policy <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_
+`Deceased User Policy <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_
 page.
 
 Repositories in the organization
--- a/core-developers/memorialization.rst
+++ b/core-developers/memorialization.rst
@@ -41,8 +41,9 @@ The PSF staff does not follow up with GitHub with regards to GitHub account
 cancellation as this action is reserved for next-of-kin or designated by
 the deceased GitHub user to act as an account successor.
 
-The general policy regarding deceased users on GitHub is described
-`here <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_.
+The general policy regarding deceased users on GitHub is described on their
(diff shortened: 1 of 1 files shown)


nedbat:
Though TBH, that warning seems kind of dumb to me... Maybe we can disable it? /cc @hugovk

hugovk:
A double trailing underscore make the link an "anonymous" one and avoids warnings:

* https://stackoverflow.com/a/74258281/724176

```diff
-`here <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_.
+`here <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`__.
```

ezio-melotti (commented):
In general the text of the link shouldn't be a generic text like "here" or "link", but should reflect the content of the page it links to, so rephrasing those links would be preferred.

If possible, the link should be the actual title of the page and/or the name of the service (e.g. "See also the [Discourse best practices for deceased community members](https://meta.discourse.org/t/best-practices-for-deceased-community-members/146210)").  Doing this will also avoid warnings.

hugovk (commented):
Good point: https://usability.yale.edu/web-accessibility/articles/links#link-text

core-developers/memorialization.rst
```diff
@@ -41,8 +41,9 @@ The PSF staff does not follow up with GitHub with regards to GitHub account
 cancellation as this action is reserved for next-of-kin or designated by
 the deceased GitHub user to act as an account successor.
 
-The general policy regarding deceased users on GitHub is described
-`here <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_.
+The general policy regarding deceased users on GitHub is described on their
+`GitHub Deceased User Policy <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_
```
ezio-melotti, line 45: Based on what we discussed, it would be better to include "GitHub" in the link, but the current sentence repeats it twice.  I would either remove the other one:
```suggestion
The general policy regarding deceased users is described on their
`GitHub Deceased User Policy <https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_ page.
```
or simply:
```suggestion
See also the `GitHub Deceased User Policy
<https://docs.github.com/en/site-policy/other-site-policies/github-deceased-user-policy>`_ page.
```

nedbat:
I've tweaked the GitHub link title to remove the duplication.

hugovk (approved):

PR #1534, opened 2025-03-23
Updated the homepage design

jenniferlinet:
----
📚 Documentation preview 📚: https://cpython-devguide--1534.org.readthedocs.build/


Mariatta:
Thanks, but I think the readme file is good without this change.

PR #1541, opened 2025-04-18
Ensure ``make -j`` uses a reasonable argument

AA-Turner:
----
📚 Documentation preview 📚: https://cpython-devguide--1541.org.readthedocs.build/

### DIFF
--- a/developer-workflow/extension-modules.rst
+++ b/developer-workflow/extension-modules.rst
@@ -547,7 +547,7 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j`` to speed-up compilation by utilizing as many CPU cores
+   Use ``make -j $(nproc)`` to speed-up compilation by utilizing as many CPU cores
    as possible or ``make -jN`` to allow at most *N* concurrent jobs.
 
 * ``make regen-configure`` updates the :cpy-file:`configure` script.
--- a/developer-workflow/extension-modules.rst
+++ b/developer-workflow/extension-modules.rst
@@ -547,8 +547,11 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j $(nproc)`` to speed-up compilation by using as many CPU cores
-   as possible or ``make -jN`` to allow at most *N* concurrent jobs.
+   Use ``make -jN`` to speed-up compilation by utilizing as many CPU cores
+   as possible, where *N* is as many CPU cores you want to spare (and have
+   memory for). Be careful using ``make -j`` with no argument, as this puts
+   no limit on the number of jobs, and compilation can sometimes use up a
+   lot of memory (like when building with LTO).
 
 * ``make regen-configure`` updates the :cpy-file:`configure` script.
 
--- a/developer-workflow/extension-modules.rst
+++ b/developer-workflow/extension-modules.rst
@@ -547,7 +547,7 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j $(nproc)`` to speed-up compilation by utilizing as many CPU cores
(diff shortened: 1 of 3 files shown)


developer-workflow/extension-modules.rst
```diff
@@ -547,7 +547,7 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j`` to speed-up compilation by utilizing as many CPU cores
+   Use ``make -j $(nproc)`` to speed-up compilation by utilizing as many CPU cores
```
hugovk, line 550:
```suggestion
   Use ``make -j $(nproc)`` to speed-up compilation by using as many CPU cores
```

getting-started/setup-building.rst
```diff
@@ -203,14 +203,13 @@ do to get a pydebug build of CPython.
...
 This will build CPython with only warnings and errors being printed to
-stderr and utilize up to 2 CPU cores. If you are using a multi-core machine
-with more than 2 cores (or a single-core machine), you can adjust the number
-passed into the ``-j`` flag to match the number of cores you have (or if your
-version of Make supports it, you can use ``-j`` without a number and Make
-will not limit the number of steps that can run simultaneously.).
+stderr. The ``-j`` argument means that :program:`make` will concurrently run
+tasks, limiting the number of parallel jobs to the number of CPU cores in your
+computer. You can adjust the number passed to the ``-j`` flag to change
+the limit on parallel jobs, which can trade RAM usage vs compilation time.
```
hugovk, line 212:
```suggestion
the limit on parallel jobs, which can trade RAM usage versus compilation time.
```

hugovk:
Let's use `$(nproc)` so we can copy and paste and get something more useful. Many modern machines have much more than 4 cores; I have an 8 and a 10.

getting-started/setup-building.rst
```diff
@@ -203,14 +203,13 @@ do to get a pydebug build of CPython.
 
 Once ``configure`` is done, you can then compile CPython with::
 
-   $ make -s -j2
```
picnixz, line 206: Note that `nproc` is part of GNU coreutils and may not necessarily be available on all Unix-like systems (namely, macOS). So Mac users should first do `brew install coreutils` or use `sysctl` instead to get the number of physical or logical cores (I don't remember which one nproc returns)

AA-Turner: I've merged, as Hugo uses macOS and didn't complain, but we can open a follow-up if need be.

hugovk: Thanks for pointing that out. I do have coreutils installed via Homebrew, which is why it worked for me. Let's not require or assume others do.

Shall we use a hardcoded value under the macOS tabs? Something like 8 feels like a good default?

gpshead: TL;DR - this is good enough.

I feel like people who understand how to run `make` already understand what `$(nproc)` can be substituted with manually if they don't have the command.  if we wanted to be ironic we'd suggest `$(python3 -c 'import os; print(os.cpu_count())')` to the person who is building python and may not yet have a built python3 (but realistically do anyways because all Linux distros offer it and macOS xcode comes with one). 😛

zware: > I feel like people who understand how to run `make` already understand what `$(nproc)` can be substituted with

That seems like a rash assumption to me; I would expect anyone with that level of comfort with `make` to not need to be walked through "run `./configure && make`".  There's also the point that `make -j $(nproc)` in the absense of a working `nproc` will leave the user with `make -j `, which is what we were trying to avoid with this change.

picnixz: The assumption would be correct if we were in the `extension-modules.rst` file, but this file is for the "getting started". I don't think we should assume that the reader is extra familiar with `make` and even knows about `nproc` and co (this section would be read by newcomers I think)

Issue #1542, opened 2025-04-18
Bug: devguide gives incorrect advice for `make -j`


Yhg1s:
### Describe the bug

The devguide suggests using `make -j` to build CPython in a few places, which is not actually a good idea. `make -j` means _no_ limit on the number of parallel jobs, not "a sensible limit". It already uses `make -j4` in a few places, and that's a very reasonable default that will work well for machines with 2-10 cores. Let's do that everywhere.

Yhg1s:
Fixed by #1541.

PR #1543, opened 2025-04-18
gh-1542: Do not suggest `make -j` 

Yhg1s:
Do not suggest `make -j` for building CPython. `make -j` means _no_ limit on

the number of concurrent jobs, not "a sensible limit considering the available memory and CPU cores". With LTO it's easy to run out of memory when too many jobs run at the same time.

### DIFF
--- a/developer-workflow/extension-modules.rst
+++ b/developer-workflow/extension-modules.rst
@@ -547,8 +547,11 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j`` to speed-up compilation by utilizing as many CPU cores
-   as possible or ``make -jN`` to allow at most *N* concurrent jobs.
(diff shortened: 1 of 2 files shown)


developer-workflow/extension-modules.rst
```diff
@@ -547,8 +547,11 @@ Now that the configuration is in place, it remains to compile the project:
 
 .. tip::
 
-   Use ``make -j`` to speed-up compilation by utilizing as many CPU cores
-   as possible or ``make -jN`` to allow at most *N* concurrent jobs.
+   Use ``make -jN`` to speed-up compilation by utilizing as many CPU cores
```
hugovk, line 550:
```suggestion
   Use ``make -jN`` to speed-up compilation by using as many CPU cores
```

Issue #1548, opened 2025-05-05
Bug: release-blocker/deferred-blocker label description needs work


hugovk:
### Describe the bug

This bullet point on GitHub labels needs a bit of work:

> * [release-blocker](https://github.com/python/cpython/labels/release-blocker)/[deferred-blocker](https://github.com/python/cpython/labels/deferred-blocker): for issues/PRs and the [branch’s release manager](https://devguide.python.org/versions/#branchstatus) removing or retaining the label as appropriate. that, unless fixed, will hold the current or next release respectively. Triagers may set these labels for issues that must be fixed before a release, will review them and determine if they indeed qualify,

https://devguide.python.org/triage/labels/#other-labels

This doesn't quite scan. There's a sentence beginning with a lowercase letter, and another ending in a comma. Perhaps part of a sentence was accidentally moved or deleted during a re-organisation.

Let's check the history, and fix it.

StanFromIreland:
This was accidentally done in https://github.com/python/devguide/pull/1432/commits/a9cd738cde85809235c3e2e0de5197de4816cf26 which attempted to reorder them.

PR #1549, opened 2025-05-05
Fix blocker label description

StanFromIreland:
Closes https://github.com/python/devguide/issues/1548


hugovk (approved):
That makes more sense, thanks!

PR #1551, opened 2025-05-06
Change Savannah Ostrowski to Savannah Bailey

savannahostrowski:
I got married! My GitHub handle will remain the same, though, because updating it is too much of a pain.

----
📚 Documentation preview 📚: https://cpython-devguide--1551.org.readthedocs.build/


JelleZijlstra (approved):
Congratulations!

You may also want to change python/voters

PR #1552, opened 2025-05-07
Add Python 3.15

hugovk:
2031, yikes!

----
📚 Documentation preview 📚: https://cpython-devguide--1552.org.readthedocs.build/

### DIFF
--- a/include/release-cycle.json
+++ b/include/release-cycle.json
@@ -1,8 +1,16 @@
 {
-  "3.14": {
+  "3.15": {
     "branch": "main",
-    "pep": 745,
+    "pep": 790,
     "status": "feature",
+    "first_release": "2026-10-01",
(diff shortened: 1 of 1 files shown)


hugovk:
linkcheck failed because GitHub rate limits us and then the workflow times out.

```
-rate limited-   https://github.com/python/cpython/blob/main/Doc/README.rst | sleeping...
-rate limited-   https://github.com/python/cpython/blob/main/Doc/Makefile | sleeping...
-rate limited-   https://github.com/python/cpython/blob/main/Grammar | sleeping...
-rate limited-   https://github.com/python/cpython/blob/main/Doc/tools/extensions/c_annotations.py | sleeping...
Error: The operation was canceled.
```

Shall we skip checking GitHub in linkcheck?

zware (approved):

Issue #1557, opened 2025-05-09
Feature: Add zstd dev lib to dependencies


devdanzin:
### Describe the enhancement or feature you would like

With PEP 784, `libzstd` becomes an optional dependency so it should be listed in the developer's guide.

PR #1558, opened 2025-05-09
Add zstd

StanFromIreland:
Closes https://github.com/python/devguide/issues/1557

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -77,6 +77,7 @@ collections           rhettinger*
 collections.abc       rhettinger*, stutzbach^
 colorsys
 compileall            carljm
+compression
 concurrent.futures    pitrou, brianquinlan, gpshead*
 configparser          ambv*
 contextlib            ncoghlan, 1st1
@@ -258,6 +259,7 @@ zipapp                pfmoore
 zipfile               alanmcintyre^, serhiy-storchaka, Yhg1s, gpshead
 zipimport             Yhg1s*
 zlib                  Yhg1s, gpshead*
+compression.zstd
 ====================  =============================================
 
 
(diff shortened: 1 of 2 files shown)


core-developers/experts.rst
```diff
@@ -258,6 +259,7 @@ zipapp                pfmoore
 zipfile               alanmcintyre^, serhiy-storchaka, Yhg1s, gpshead
 zipimport             Yhg1s*
 zlib                  Yhg1s, gpshead*
+compression.zstd
```
StanFromIreland, line 262: Do we use this ordering or should it be under compression?

ned-deily (commented):
The macOS dependencies section also needs to be updated:

```
-         $ brew install pkg-config openssl@3 xz gdbm tcl-tk mpdecimal
+         $ brew install pkg-config openssl@3 xz gdbm tcl-tk mpdecimal zstd

```
and:

```
+         $ sudo port install pkgconfig openssl xz gdbm tcl tk +quartz mpdecimal
-         $ sudo port install pkgconfig openssl xz gdbm tk +quartz mpdecimal zstd

```
(The removal of `tcl` is deliberate; it's a dependency of the `tk` port and thus does not need to be specified.)

core-developers/experts.rst
```diff
@@ -77,6 +77,8 @@ collections           rhettinger*
 collections.abc       rhettinger*, stutzbach^
 colorsys
 compileall            carljm
+compression
+compression.zstd
```
StanFromIreland, line 81: cc @AA-Turner You have been active here, do you want to be added? I don't think Emma can be added as I believe it is limited to core.

core-developers/experts.rst
```diff
@@ -77,6 +77,8 @@ collections           rhettinger*
 collections.abc       rhettinger*, stutzbach^
 colorsys
 compileall            carljm
+compression
```
hugovk, line 80: I don't think we need `compression` on its own?

Let's move the others (`lzma`, `bz2`, `gzip` and `zlib`) under `compression.`.

https://docs.python.org/3.14/whatsnew/3.14.html#pep-784-adding-zstandard-to-the-standard-library

ned-deily (commented):
Thanks for making the changes. They LGTM.

AA-Turner (approved):

Issue #1309, opened 2024-04-18
Document how to rebuild only one or a few docs files


hugovk:
Let's document the method to build docs but only rebuild one or a few files like `make -C Doc html SOURCES=<path>`, it can be much quicker than rebuilding the whole set.

For example:

* `make -C Doc html SOURCES=whatsnew/3.13.rst`
* `make -C Doc html SOURCES="whatsnew/3.13.rst whatsnew/3.12.rst"` (double check this syntax)

And I think it might be Unix only?

AmboThom:
Hi @hugovk,

I'm willing to take this issue and it might take me about 2 weeks to work on this task. Before I start have a few questions:

- Where would the method to build the docs go? Would it be on a new page or added to an existing page? If it's an existing page, maybe [this page](https://devguide.python.org/documentation/help-documenting/) or [here](https://devguide.python.org/documentation/devguide/).
- You mentioned a few files, do you know the maximum amount of docs that can be built this way?

It might take me about 2 weeks to work on this task.

AA-Turner:
> And I think it might be Unix only?

More precisely, platforms that support `make` -- I believe it is possible to use on Windows via cygwin.

On Windows (or any platform), one can use ``sphinx-build -M Doc Doc/build [SOURCES]``, though you need to run ``blurb merge`` yourself.

> It might take me about 2 weeks to work on this task.

There's no rush in terms of a PR, but I wouldn't expect this to take a fortnight to write up -- to set expectations, this would only be a paragraph or two of extra text

> Where would the method to build the docs go?

I'd suggest in the ["Building the documentation" section](https://devguide.python.org/documentation/start-documenting/#building-the-documentation)

> the maximum amount of docs that can be built this way?

No limit is imposed. But we should suggest only using this in sensible cases, such as those that Hugo outlined.

A

MichaByte:
Heya, this seems stale. Does anyone mind if I take a crack at it?

hugovk:
@MichaByte Please do!

MichaByte:
I just made #1562!

PR #1562, opened 2025-05-19
gh-1309: Document how to selectively build RST pages

MichaByte:
This PR adds a small section to `documentation/start-documenting` that informs the reader of how to build only certain documentation pages. I tested my instructions on Linux and Windows. Please let me know if any changes are needed!

Thanks,
Micha

----
📚 Documentation preview 📚: https://cpython-devguide--1562.org.readthedocs.build/

### DIFF
--- a/documentation/start-documenting.rst
+++ b/documentation/start-documenting.rst
@@ -153,6 +153,25 @@ To build the docs as HTML, run:
            start a local server, and automatically reload the page in your
            browser when you make changes to reST files (Unix only).
 
+It is also possible to build only certain pages of the documentation in order
+to save time during the build process. Following is an example for building two
+pages:
+
+.. tab:: Unix/macOS
+
+   .. code-block:: shell
+
+      make html SOURCES="tutorial/classes.rst tutorial/inputoutput.rst"
+
+.. tab:: Windows
+
+   See :ref:`using-sphinx-build`. When invoking ``sphinx-build``, pass the
+   desired pages as the final parameter, like so:
+
+   .. code-block:: dosbatch
+
+      python -m sphinx -b html . build/html tutorial/classes.rst tutorial/inputoutput.rst
+
 To check the docs for common errors with `Sphinx Lint`_
 (which is run on all :ref:`pull requests <pullrequest>`), use:
 


documentation/start-documenting.rst
```diff
@@ -153,6 +153,25 @@ To build the docs as HTML, run:
            start a local server, and automatically reload the page in your
            browser when you make changes to reST files (Unix only).
 
+It is also possible to build only certain pages of the documentation in order
+to save time during the build process. Following is an example for building two
+pages:
```
StanFromIreland, line 158:
```suggestion
To build only certain pages of the docs as HTML, run:
```
It can be simpler, like the previous introduction.

documentation/start-documenting.rst
```diff
@@ -153,6 +153,25 @@ To build the docs as HTML, run:
...
+.. tab:: Unix/macOS
+
+   .. code-block:: shell
+
+      make html SOURCES="tutorial/classes.rst tutorial/inputoutput.rst"
+
+.. tab:: Windows
+
+   See :ref:`using-sphinx-build`. When invoking ``sphinx-build``, pass the
+   desired pages as the final parameter, like so:
```
StanFromIreland, line 169:
```suggestion
   When invoking ``sphinx-build``, pass the
   desired pages as the final parameter, like so:
```
Do we need the reference here, if we really do IMO it is more natural to have it after i.e.

`When ... parameter. See XXX.`

documentation/start-documenting.rst
```diff
@@ -153,6 +153,25 @@ To build the docs as HTML, run:
            start a local server, and automatically reload the page in your
```
StanFromIreland, line 153: Why not just add it as a note here?

StanFromIreland (commented):
IMO this does not need a whole section, I've never built specific pages, by the time I've written the command `make html` would have finished anyway. I don't see much need for this here, especially such a large section.

A better place would be above in the `.. tip::` explaining you can pass `SOURCES=""` with desired paths.

ZeroIntensity:
Stan, be aware that a lot of these changes are being discussed at the sprints in-person.

StanFromIreland:
I assumed so, yet I am not there :-(, so all I can do is comment.

ZeroIntensity:
Right, but some changes have already been informally approved elsewhere, so it's not particularly helpful to say things like "IMO this does not need a whole section."

I get that you're aspiring to be a triager or on the documentation team or whatever, but this isn't the way to go about it! Documentation reviews are the most helpful when they're about actually fixing typos or incorrect information, but wording and formality nitpicks (like you tend to suggest) are just noise that actually *slow down* documentation PRs.

I think you would really benefit from having a core dev mentor that walked you through this. Have you asked around at all?

willingc (approved):
Thanks @MichaByte for your contribution during the sprint. Also congrats on your first contribution to the devguide. 🎉 :sunny: I hope that there will be more in the future. :smile:

willingc:
Closes #1309

PR #1567, opened 2025-06-03
Add Irvan Putra as coordinator of Indonesian translation

irvan-putra:
After discussion with @jwjacobson, there will be at least two coordinating people in charge of Indonesian translation.

----
📚 Documentation preview 📚: https://cpython-devguide--1567.org.readthedocs.build/

### DIFF
--- a/documentation/translating.rst
+++ b/documentation/translating.rst
@@ -42,7 +42,7 @@ details.
      - :github:`GitHub <python/python-docs-hu>`,
        `mailing list <https://mail.python.org/pipermail/python-hu>`__
    * - `Indonesian (id) <https://docs.python.org/id/>`__
-     - Oon Arfiandwi (:github-user:`oonid`),
+     - Irvan Putra (:github-user:`irvan-putra`),
        Jeff Jacobson (:github-user:`jwjacobson`)
      - :github:`GitHub <python/python-docs-id>`
    * - Italian (it)


documentation/translating.rst
```diff
@@ -42,7 +42,7 @@ details.
      - :github:`GitHub <python/python-docs-hu>`,
        `mailing list <https://mail.python.org/pipermail/python-hu>`__
    * - `Indonesian (id) <https://docs.python.org/id/>`__
-     - Oon Arfiandwi (:github-user:`oonid`),
```
AA-Turner, line 45: For now please don't remove any names, we'd prefer that inactive co-ordinators remove themselves at present.

Mariatta: Since the person has been unreachable despite multiple attempts, in my mind it is unlikely they will come back here and remove themselves. It would be more accurate to display the current active coordinators only.

oonid: hello,
I have coordinated with @irvan-putra,
confirmed that he will lead the next translation effort.

irvan-putra: So is it enough that @oonid commented above or @oonid need to create separate pull request?

Mariatta: I would say this is good enough, we don't need a separate PR. Thank you!

hugovk (approved):
Thank you!

PR #1570, opened 2025-06-06
Switch to non-deprecated Tools/wasm/wasi

encukou:
`Tools/wasm/wasi.py` gave me a deprecation warning.

----
📚 Documentation preview 📚: https://cpython-devguide--1570.org.readthedocs.build/

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -87,6 +87,7 @@ linkcheck_anchors_ignore = [
 linkcheck_ignore = [
     # Checks fail due to rate limits
     r'https://github.com/.*',
+    r'https://www.gnu.org/software/autoconf/',
     # The Discourse groups are private unless you are logged in
     'https://discuss.python.org/groups/staff',
     'https://discuss.python.org/groups/moderators',
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -297,13 +297,13 @@ Emscripten            hoodmane, pmp-p, rdb, rth, ryanking13
 FreeBSD
 HP-UX
 iOS                   freakboy3742, ned-deily
+JVM/Java              frank.wierzbicki^
 Linux
 macOS                 ronaldoussoren, ned-deily, freakboy3742
 NetBSD1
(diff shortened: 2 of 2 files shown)


brettcannon (approved):

brettcannon:
Unfortunately it's a Python 3.15 thing, so the old instructions are necessary for 3.13 and 3.14. But since the location is practically the same I suspect people will notice and figure it out.

hugovk:
Or shall we serve our users better and add version tabs like the macOS instructions here? 

https://devguide.python.org/getting-started/setup-building/#macos

encukou:
Here, I'd prefer a description of what changed. The alternatives are very similar and I might need both in a single session.

brettcannon:
> Here, I'd prefer a description of what changed. The alternatives are very similar and I might need both in a single session.

The only change is `wasi/__main__.py` vs. `wasi.py`. There's no functional difference right now and it was all done to make the files more manageable long-term (I would have backported to 3.14 but this was all part of my snafu when I accidentally committed to `main` right after the `3.14` branch was created).

encukou:
So, right now, `wasi.py` works best for users?

Maybe 3.20 should be the version that *starts* showing the scary `⚠️ WARNING`?

encukou:
> If "best" is "works for most people", then yes. But things will diverge more as time goes on anyway, e.g. adding externals support is hopefully coming in 3.15.

So, as soon as those are in, the devguide should recommend `wasi/`. And have a note about 3.14 and below. Right?


> I would have backported to 3.14 but this was all part of my snafu when I accidentally committed to main right after the 3.14 branch was created

There's a chance @hugovk would allow backporting a backwards-compatible infra change -- especially if the backport would omit the warning.

willingc (approved):

brettcannon:
> @brettcannon Did you want this PR to wait until [python/cpython#133600](https://github.com/python/cpython/issues/133600) landed?

It would have been convenient, but now I have a bit of motivation to get that issue closed this week. 😅

PR #1580, opened 2025-06-17
New core dev instructions: Use a script instead of hand-editing.

encukou:
The `developers.csv` got a bit out of sync recently. Update instructions to help avoid that happening again.

----
📚 Documentation preview 📚: https://cpython-devguide--1580.org.readthedocs.build/


core-developers/become-core-developer.rst
```diff
@@ -55,8 +55,9 @@ are granted through these steps:
    - Enable the various new privileges.
    - Remove the new committer from the triage team, if applicable.
    - Add their details to `🔒 python/voters <https://github.com/python/voters>`_.
-   - Update the devguide to publicly list their team membership
-     at :ref:`developers`.
+   - Once the python/voters update is merged, regenerate the public team membership
+     list at :ref:`developers`.
+     See "Public list of members" in the ``voters`` README.
```
hugovk:
```suggestion
     See `🔒 Public list of members
     <https://github.com/python/voters#public-list-of-members>`__.
```

encukou: Almost the same link is right above this; I don't think we need another :lock: link here.

PR #1584, opened 2025-06-19
Add tomasr8 to gettext & i18n

tomasr8:
I don't consider myself an expert but I've worked in this area both in CPython and my $job so I thought it'd be useful to add myself here :)

----
📚 Documentation preview 📚: https://cpython-devguide--1584.org.readthedocs.build/


hugovk:
Thank you!

PR #1598, opened 2025-07-02
Update libmpdec-related instructions

skirpichev:
* drop --with-system-libmpdec on macOS (default to yes since 3.13)
* recommend deb.sury.org for Debian 12 and Ubuntu 24.04

----
📚 Documentation preview 📚: https://cpython-devguide--1598.org.readthedocs.build/

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -781,9 +781,12 @@ some of CPython's modules (for example, ``zlib``).
             libncurses5-dev libreadline6-dev libsqlite3-dev libssl-dev \
             lzma lzma-dev tk-dev uuid-dev zlib1g-dev libmpdec-dev libzstd-dev
 
-   Note that Debian 12 and Ubuntu 24.04 do not have the ``libmpdec-dev`` package.  You can safely
-   remove it from the install list above and the Python build will use a bundled version.
-   Alternatively, you can install this package from https://deb.sury.org.
+   Note that Debian 12 and Ubuntu 24.04 do not have the ``libmpdec-dev``
+   package.  You can safely remove it from the install list above and the
+   Python build will use a bundled version.  But we recommend using the system
+   `libmpdec <https://www.bytereef.org/mpdecimal/doc/libmpdec/>`_ library.
+   Either built it from sources or install this package from
+   https://deb.sury.org.
 
 .. tab:: macOS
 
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -823,7 +824,6 @@ some of CPython's modules (for example, ``zlib``).
             $ GDBM_CFLAGS="-I$(brew --prefix gdbm)/include" \
                GDBM_LIBS="-L$(brew --prefix gdbm)/lib -lgdbm" \
                ./configure --with-pydebug \
-                           --with-system-libmpdec \
```
hugovk, line 826: Confirmed builds and tests pass on macOS without this for `3.13`/`3.14`/`main`.

getting-started/setup-building.rst
```diff
@@ -783,6 +783,7 @@ some of CPython's modules (for example, ``zlib``).
 
    Note that Debian 12 and Ubuntu 24.04 do not have the ``libmpdec-dev`` package.  You can safely
    remove it from the install list above and the Python build will use a bundled version.
+   Alternatively, you can install this package from https://deb.sury.org/.
```
hugovk:
```suggestion
   Alternatively, you can install this package from https://deb.sury.org.
```

I'm not familiar with this source so will leave for someone else to review if we should recommend it.

But generally, do I get this right: For Debian 12 and Ubuntu 24.04, if users don't use a third-party package source, they'll use the fallback to the bundled copy. But we'll remove that bundled copy in Python 3.16 (branch opens May 2026, final release October 2027).

However, Ubuntu 24.04 is EOL in 2029 (and Debian 12 in June 2026). What is our recommendation to users of Ubuntu 24.04? (And later versions of Ubuntu and Debian)

And also: if the bundled fallback is already deprecated and going away in 3.16, should we be encouraging everyone to start using the replacement (an external copy) already, to help identify potential issues ahead of 3.16?

skirpichev: > I'm not familiar with this source so will leave for someone else to review if we should recommend it.

We use that in the CPython CI.

> What is our recommendation to users of Ubuntu 24.04?

Em?  Use https://deb.sury.org.

> And later versions of Ubuntu and Debian

Later versions come with the libmpdec-dev.

> should we be encouraging everyone to start using the replacement (an external copy) already

Yes, probably this should be rephrased.

skirpichev: I think this was updated.

PR #1600, opened 2025-07-05
Remove deleted modules from experts index

mattwang44:
Remove those modules deleted in 3.13 ([PEP 594](https://peps.python.org/pep-0594/)) from the expert index

### DIFF
--- a/core-developers/experts.rst
+++ b/core-developers/experts.rst
@@ -126,7 +126,6 @@ http
 idlelib               kbkaiser (inactive), terryjreedy*, serwy (inactive),
                       taleinat
 imaplib
-imghdr
 importlib             brettcannon
 inspect               1st1
 io                    benjaminp, stutzbach^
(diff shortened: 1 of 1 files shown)


StanFromIreland:
This whole file should be regenerated, there are also several missing entries, e.g. `zoneinfo` off the top of my head.

hugovk (approved):
I don't think we necessarily need to list all 300+ modules if there's no expert for them. We can add them as needed.

ezio-melotti (approved):

ezio-melotti:
> I don't think we necessarily need to list all 300+ modules if there's no expert for them. We can add them as needed.

IIRC that was done intentionally, both to be explicit about modules that don't have a specific maintainer and to encourage devs to add their names next to modules they are familiar with once they see they have no maintainers.

hugovk:
193 in the list when added in 2009, 34 with names to begin with: https://github.com/python/cpython/commit/1d74e137a9c08c5894b89ab51e7f12226b695533

Anyway, let's merge this.

PR #1611, opened 2025-07-19
Prevent `make venv` from reporting success when it actually failed

### LINKED ISSUE #1607: `make venv` succeeds even if resolution fails (e.g. an old Python version)
### Describe the bug

We should print a better error, we had a contributor at the sprints who couldn't get a venv set up due to having Python 3.10 installed -- we require 3.11.

cc @hugovk 

### Screenshots

_No response_

### Additional context

_No response_

nachocab:
Fixes #1607

----
📚 Documentation preview 📚: https://cpython-devguide--1611.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ venv:
 
 .PHONY: ensure-venv
 ensure-venv:
+	set -e;
 	@if [ ! -d $(VENVDIR) ] ; then \
 		echo "Creating venv in $(VENVDIR)"; \
 		if $(UV) --version >/dev/null 2>&1; then \
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,8 @@ venv:
 
 .PHONY: ensure-venv
 ensure-venv:
-	set -e;
 	@if [ ! -d $(VENVDIR) ] ; then \
+		set -e; \
 		echo "Creating venv in $(VENVDIR)"; \
 		if $(UV) --version >/dev/null 2>&1; then \
 			$(UV) venv --python=$(PYTHON) $(VENVDIR); \
--- a/documentation/start-documenting.rst
+++ b/documentation/start-documenting.rst
@@ -76,24 +76,22 @@ To build the documentation, follow the steps in one of the sections below.
 You can view the documentation after building the HTML
 by opening the file :file:`Doc/build/html/index.html` in a web browser.
 
-.. _doc-initial-requirements:
-
 Initial requirements
 --------------------
 
-
-Ensure your current working dir is the ``Doc`` subdirectory in your
-:ref:`CPython repository clone <checkout>`.
-
-.. code-block:: shell
-
-  cd Doc
-
-Ensure your Python version is at least 3.11 (Sphinx requirement).
+Ensure your current working directory is the top level ``Doc/`` directory
+inside your :ref:`CPython repository clone <checkout>`.
 
 .. code-block:: shell
 
-  python --version
+   cd Doc
+
+Ensure your Python version is at least 3.11.
+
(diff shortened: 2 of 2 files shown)


documentation/start-documenting.rst
```diff
@@ -76,12 +76,24 @@ To build the documentation, follow the steps in one of the sections below.
...
+Initial requirements
+--------------------
+
+
+Ensure your current working dir is the ``Doc`` subdirectory in your
+:ref:`CPython repository clone <checkout>`.
+
+.. code-block:: shell
+
+  cd Doc
```
StanFromIreland: This needs an explanation like there was before, otherwise it may lead people to `cd Doc` while already in the dir.

Makefile
```diff
@@ -57,6 +57,7 @@ venv:
 
 .PHONY: ensure-venv
 ensure-venv:
+	set -e;
```
brianschubert: Isn't this meaningless in a Makefile? Make runs each command in a separate shell, so setting a shell setting on one line has no effect on the subsequent lines. Make also already exits if a command returns a non-zero status, so the behavior this is trying to add should already exist

nachocab: Thanks for the comment Brian. My bad, I just moved it inside the if-block, where it is needed (if the `pip install` command fails, the `echo` command at the end will still run)

documentation/start-documenting.rst
nachocab: Thanks, Stan. I rephrased it a bit to be more clear (it still says "Ensure" like it was before)

StanFromIreland: I meant:

> switch to it with ``cd Doc`` if necessary.

Please add something similar, don't leave the codeblock without any explanation as to what is in it. There is a similar situation for the second one.

AA-Turner: @StanFromIreland can you elaborate, or suggest a diff? I'm mentoring @nachocab at the EP25 sprints. The "ensure" language seems reasonably clear.

A

StanFromIreland: From my view, the command with no explanation seems like a step that you should run, however it is conditional. I think it should be explained, similarly to what was there before ("switch to it with cd Doc if necessary"):

```
Ensure your current working directory is the top level ``Doc/`` directory
inside your :ref:`CPython repository clone <checkout>`. You can switch to
it with:

File: documentation/start-documenting.rst [hunk 3]
```diff
@@ -76,12 +76,22 @@ To build the documentation, follow the steps in one of the sections below.
...
+--------------------
+
+Ensure your current working directory is the top level ``Doc/`` directory
+inside your :ref:`CPython repository clone <checkout>`.
+
+.. code-block:: shell
+
+   cd Doc
+
+Ensure your Python version is at least 3.11.
```
StanFromIreland: Again for this case, there is no link to the following codeblock? I think a reader may ask: should I run it? should the output be identical? etc.

```suggestion
Ensure your Python version is at least 3.11. You can verify with the following command:

```

documentation/start-documenting.rst
```diff
@@ -76,12 +76,22 @@ To build the documentation, follow the steps in one of the sections below.
...
 
-   The following instructions all assume your current working dir is
-   the ``Doc`` subdirectory in your :ref:`CPython repository clone <checkout>`.
-   Make sure to switch to it with ``cd Doc`` if necessary.
+.. code-block:: shell
 
+   $ python --version
+   Python 3.13.5
```
StanFromIreland: Continuing on from previous note, as it would not let me select these lines.

```suggestion
   python --version
```

PR #1605, opened 2025-07-15
Add explanation to buildbots.rst about triggering buildbot on a PR

hoodmane:
----
📚 Documentation preview 📚: https://cpython-devguide--1605.org.readthedocs.build/

PR #1617, opened 2025-07-23
Shorter title: "Triggering on pull requests"

hugovk:
Little followup to https://github.com/python/devguide/pull/1605, we're already on the buildbots page, we can make the title a bit more concise. cc @hoodmane


hoodmane:
Thanks @hugovk !

Issue #1606, opened 2025-07-18
Feature:  Mention GraalPy in "Other interpreter implementations" sections


aisk:
### Describe the enhancement or feature you would like

There is an [Other interpreter implementations](https://devguide.python.org/#other-interpreter-implementations) section in the devguide that lists some alternate implementations like Jython.

But nowadays, Jython is more or less outdated (it still only supports 2.7, and 3.x is still far away). In the JVM world, there is another implementation called [GraalPy](https://github.com/oracle/graalpython), which is backed by big company (Oracle) and has Python 3 support, with more active development. Don't be misled by its name, it can be used with the normal JVM.
So I think it's worth mentioning in this section.

PR #1618, opened 2025-07-23
gh-1606: Mention GraalPy in other implementaion section

aisk:

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -217,14 +217,16 @@ CPython, they always have more things they would like to do than they have
 developers to work on them.  Some major examples that may be of interest are:
 
 * PyPy_: A Python interpreter focused on high speed (JIT-compiled) operation
-  on major platforms
+  on major platforms.
 * Jython_: A Python interpreter focused on good integration with the Java
-  Virtual Machine (JVM) environment
+  Virtual Machine (JVM) environment.
+* GraalPy: Another Python interpreter which has first-class support for
+  embedding in Java, built on GraalVM.
 * IronPython_: A Python interpreter focused on good integration with the
-  Common Language Runtime (CLR) provided by .NET and Mono
+  Common Language Runtime (CLR) provided by .NET and Mono.
 * Stackless_: A Python interpreter focused on providing lightweight
   microthreads while remaining largely compatible with CPython specific
-  extension modules
+  extension modules.
 * MicroPython_: A tiny Python interpreter with small subset of the Python
   standard library that is optimised to run on microcontrollers and in
   constrained environments.


index.rst
```diff
@@ -217,14 +217,16 @@ CPython, they always have more things they would like to do than they have
 developers to work on them.  Some major examples that may be of interest are:
 
 * PyPy_: A Python interpreter focused on high speed (JIT-compiled) operation
-  on major platforms
+  on major platforms.
 * Jython_: A Python interpreter focused on good integration with the Java
-  Virtual Machine (JVM) environment
+  Virtual Machine (JVM) environment.
+* GraalPy: Another Python interpreter which has first-class support for
```
brianschubert: Should we add a link to a project homepage? Maybe https://www.graalvm.org/python/?

index.rst
```diff
@@ -217,14 +217,16 @@ CPython, they always have more things they would like to do than they have
 developers to work on them.  Some major examples that may be of interest are:
 
 * PyPy_: A Python interpreter focused on high speed (JIT-compiled) operation
-  on major platforms
+  on major platforms.
 * Jython_: A Python interpreter focused on good integration with the Java
```
AA-Turner, line 225: Jython only supports version 2.7 of the language. Perhaps we should preferentially list GraalPy? cc @jeff5 @timfel

index.rst
aisk: Other alternate implementations don't have a link either, so I think it's better to keep them consistent now? If the link is preferred, maybe we can add them in a future PR.

AA-Turner: I think better to use this PR. The link Brian suggested seems fine.

A

index.rst
aisk: I agree. GraalPy can be used in a standard JVM (such as OpenJDK), so it's a better choice, as it's the only way to use Python 3 in the Java ecosystem.

index.rst
brianschubert: The other implementations actually do ;) The trailing underscore creates a hyperlink in rst. The link targets are defined at the bottom of the file: https://github.com/aisk/devguide/blob/aisk-patch-1/index.rst#L328-L332

aisk: Ah sorry, just notice this...

index.rst
```diff
@@ -217,14 +217,16 @@ CPython, they always have more things they would like to do than they have
 developers to work on them.  Some major examples that may be of interest are:
 
 * PyPy_: A Python interpreter focused on high speed (JIT-compiled) operation
-  on major platforms
+  on major platforms.
+* GraalPy_: Another Python interpreter which has first-class support for
```
StanFromIreland:
```suggestion
* GraalPy_: A Python interpreter which has first-class support for
```
„Another” is not needed now since they were reordered.

hugovk (approved):

AA-Turner (approved):

PR #1587, opened 2025-06-23
Update `translations/coordinating.rst`

StanFromIreland:
This page is aimed at coordinators, so I assume some git/gh knowledge etc.

Some things that need to be discussed:

- Use Discourse for announcements as it is public (more accessibly for translators), and has more readers nowadays?
- Add some general point for when translations are moved under the python org, because this is quite important it should be noted?
- Issue: https://github.com/python/devguide/issues/1586

closes https://github.com/python/devguide/issues/1586

PR #1619, opened 2025-07-24
docs: add label to translating so it can be referenced

marcorichetta:
Added a label to fix the link to the translation guide in https://devguide.python.org/

image

- [x] Tested it locally with `makehtml`

----
📚 Documentation preview 📚: https://cpython-devguide--1619.org.readthedocs.build/

### DIFF
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -1,3 +1,5 @@
+.. _translating:
+
 ===========
 Translating
 ===========


hugovk:
As https://github.com/python/devguide/pull/1587 is taking some time, let's just merge this.

StanFromIreland:
@hugovk the other PR fixes all of the issues, I recommend it instead.

hugovk:
Yep, but I can't reopen that because the branch was deleted, so let's at least do this one.

Thanks!

StanFromIreland:
Ah I did not realise that blocks it, as the diff is still visible. Also, if you have time I wouldn't mind the rest of your partial review:-)

PR #1622, opened 2025-08-03
Add Emma Smith to Core Devs

emmatyping:
Per the editorconfig, the CSV file has CR line endings, it seems that VS Code honored this and updated the line with Steve Holden's entry to include a CR instead of LF.

I viewed the generated webpage and it looked right for both my entry and Steve Holden's.

----
📚 Documentation preview 📚: https://cpython-devguide--1622.org.readthedocs.build/


StanFromIreland:
See Petr’s pr https://github.com/python/devguide/pull/1580 on how the list should be generated.

gpshead (approved):

PR #1624, opened 2025-08-06
Extend style guide "Use simple language" section

StanFromIreland:
The diff is quite short, and I think it will explain better than a comment here. A section on gender neutrality may also be a good addition, there is in fact already one in the translation style guide.

### DIFF
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -41,8 +41,9 @@ For more details about translations and their progress, see `the dashboard
      - Sanyam Khurana (:github-user:`CuriousLearner`)
      - :github:`GitHub <CuriousLearner/python-docs-hi-in>`
    * - Hungarian (hu)
-     -
+     - Tamás Bajusz (:github-user:`gbtami`)
      - :github:`GitHub <python/python-docs-hu>`,
+       `mailing list <https://mail.python.org/pipermail/python-hu>`__
    * - `Indonesian (id) <https://docs.python.org/id/>`__
      - | Irvan Putra (:github-user:`irvan-putra`),
        | Jeff Jacobson (:github-user:`jwjacobson`)


documentation/style-guide.rst
```diff
@@ -103,11 +103,16 @@ Unix
...
-Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
-such as "for example" or "that is."
+The documentation has a world-wide audience of varying English proficiency.
+To make the documentation more accessible and inclusive, the following should be
+avoided:
+
+- Esoteric phrasing.
+- Abbreviations, such as Latin "e.g." or "i.e.," or symbols, such as the
+  ampersand ('&'), where English words will do, such as
+  "for example", "that is" and "and," respectively.
```
AA-Turner, line 113: I don't agree with discouraging '&'. It's relatively common, and helps distinguish items in a list, e.g. "Tom & Jerry and Barbie & Ken have both featured in major animated works".

documentation/style-guide.rst
```diff
@@ -103,11 +103,16 @@ Unix
...
+- Idioms, as they are often confusing to non-native speakers and
+  are difficult to translate.
```
AA-Turner, line 115: I've said elsewhere, but part of the draw of Python is that it has historically been less 'clinical'/corporate/dry than other languages, and silly examples (*Monty Python*) & idioms in prose are part of that. I think there's a fine balance between avoiding overuse and confusion and a prohibition: I would be sad to see this type of thing go.

documentation/style-guide.rst
```diff
@@ -103,11 +103,16 @@ Unix
...
-Avoid esoteric phrasing where possible.  Our audience is world-wide and may not
-be native English speakers.
-
-Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
-such as "for example" or "that is."
+The documentation has a world-wide audience of varying English proficiency.
+To make the documentation more accessible and inclusive, the following should be
+avoided:
+
+- Esoteric phrasing.
```
ezio-melotti, line 110: Ironically "Esoteric phrasing" could arguably fit its own definition, even though these rules might not necessarily apply to the devguide itself. 🙃

documentation/style-guide.rst
ezio-melotti:
```suggestion
- Abbreviations, such as Latin "e.g." or "i.e." or symbols, such as the
  ampersand ("&"), where English words will do, such as
  "for example", "that is" and "and" respectively.
```

> I don't agree with discouraging '&'. It's relatively common

The same could be -- and was -- said for "e.g." and "i.e." too.
I wonder if this made more sense for printed media, since nowadays people can easily look those terms up and learn them instead.

PR #1628, opened 2025-08-11
Add a note to not set -O0 with --enable-optimizations

taegyunkim:
----
📚 Documentation preview 📚: https://cpython-devguide--1628.org.readthedocs.build/


hauntsaninja (commented):
There are a lot of things one should not do.

I don't know why this is important enough to justify this much of a callout. If it is actually important, I would vote we make configure complain about this combination

PR #1629, opened 2025-08-12
Document how to use varargs in Argument Clinic

skirpichev:
----
📚 Documentation preview 📚: https://cpython-devguide--1629.org.readthedocs.build/

### DIFF
--- a/development-tools/clinic.rst
+++ b/development-tools/clinic.rst
@@ -1495,23 +1495,33 @@ You can still use a self converter, a return converter, and specify
 a *type* argument to the object converter for :c:macro:`METH_O`.
 
 
-How to convert var-positional parameter functions
--------------------------------------------------
+How to convert ``*args`` parameters (starargs / var-positional)
+---------------------------------------------------------------
 
-To convert a var-positional parameter function, prepend the parameter name
-with ``*`` and use the the ``array`` converter.
+There are two converters suitable for ``*args``: *array* and *tuple*.
+
+Using the *array* converter will provide the implementation function with
+a C array *args* of type of :c:type:`PyObject * <PyObject>` and the number
+of items in the array as :c:type:`Py_ssize_t` *args_length*.
 For example::
 
    /*[clinic input]
    var_positional_sample
 
-       foo: int
+       spam: int
        *args: array
    [clinic start generated code]*/
 
-The implementation function will receive var-positional arguments as C array
-*args* of :c:type:`PyObject * <PyObject>`.  Alternatively, you could use
-``tuple`` converter to pass a regular :c:type:`PyTupleObject` as argument.
+Using the *tuple* converter will provide the implementation function with
+a standard :c:type:`PyTupleObject`.
+For example::
+
+   /*[clinic input]
+   var_positional_sample
+
+       spam: int
+       *args: tuple
+   [clinic start generated code]*/
 
 .. versionadded:: 3.11
 


development-tools/clinic.rst
```diff
@@ -1495,6 +1495,27 @@ You can still use a self converter, a return converter, and specify
...
+
+       foo: int
+       *args: object
+   [clinic start generated code]*/
+
+The implementation function will receive var-positional arguments
+as *args* array.
+
+.. versionadded:: 3.11
+.. versionchanged:: 3.14
```
vstinner: What changed in 3.14?

skirpichev: https://github.com/python/cpython/pull/126064

vstinner: Can you please add a sentence to explain this change?

skirpichev: Then, maybe just drop this entry?  IIUC, we only document current state of art here.

skirpichev:
```suggestion
```

skirpichev: Or this:
```suggestion
.. versionchanged:: 3.14
   Implementation was optimized to avoid temporary tuple creation. 
```

AA-Turner: I've applied the first suggestion, an internal optimisation doesn't affect how varargs are used

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,26 @@ You can still use a self converter, a return converter, and specify
 a *type* argument to the object converter for :c:macro:`METH_O`.
 
 
+How to convert var-positional parameter functions
+-------------------------------------------------
+
+To convert a var-positional parameter function,
+prepending the parameter name with ``*`` ,
+and the parameter should use the ``object`` converter::
```
AA-Turner:
```suggestion
To convert a var-positional parameter function, prepend the parameter name
with ``*`` and use the the ``object`` converter.
For example::
```

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,26 @@ You can still use a self converter, a return converter, and specify
...
+
+   /*[clinic input]
+   var_positional_sample
+
+       foo: int
+       *args: object
```
AA-Turner: Is this true? In `libclinic` I only see a `tuple` and `array` converter as subclasses of `VarPosCConverter`, and I can't see any uses of `*arg_name: object` in the tree, though there are plenty of uses of `*args : tuple` or `*args: array`.

skirpichev: Indeed, I forgot to rename it.

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,26 @@ You can still use a self converter, a return converter, and specify
 a *type* argument to the object converter for :c:macro:`METH_O`.
 
 
+How to convert var-positional parameter functions
+-------------------------------------------------
+
+To convert a var-positional parameter function, prepend the parameter name
+with ``*`` and use the the ``object`` converter.
```
skirpichev:
```suggestion
with ``*`` and use the the ``array`` converter.
```

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,26 @@ You can still use a self converter, a return converter, and specify
...
+For example::
+
+   /*[clinic input]
+   var_positional_sample
+
+       foo: int
+       *args: object
```
skirpichev:
```suggestion
       *args: array
```

AA-Turner (changes requested):
We should mention both the `tuple` and `array` converters, they do different things. Notably `array` results in ` PyObject * const *args` and `Py_ssize_t args_length`, whereas `tuple` just passes a regular `PyTupleObject`.

Also, I found the PR where `: object` was removed (https://github.com/python/cpython/pull/126560 by @serhiy-storchaka)

skirpichev:
> Also, I found the PR where : object was removed

It's not removed, just an alias for tuple.

development-tools/clinic.rst
```diff
@@ -1495,6 +1495,27 @@ You can still use a self converter, a return converter, and specify
...
+   /*[clinic input]
+   var_positional_sample
+
+       foo: int
+       *args: array
+   [clinic start generated code]*/
+
+The implementation function will receive var-positional arguments as C array
+*args* of :c:type:`PyObject * <PyObject>`.  Alternatively, you could use
+``tuple`` converter to pass a regular :c:type:`PyTupleObject` as argument.
```
AA-Turner: A suggested rephrasing to split `tuple` and `array`. I've also switched to saying `*args` instead of emphasising var-positional, the documentation uses `*args` or starargs, when discussing the feature in the language reference, it seems to be only internal to Clinic to say varpos.

```suggestion
How to convert ``*args`` parameters (starargs / var-positional)
---------------------------------------------------------------

There are two converters suitable for ``*args``: *array* and *tuple*.

Using the *array* converter will provide the implementation function with
a C array *args* of type of :c:type:`PyObject * <PyObject>` and the number
of items in the array as :c:type:`Py_ssize_t` *args_length*.
For example::

   /*[clinic input]
   var_positional_sample

       spam: int
       *args: array
   [clinic start generated code]*/

Using the *tuple* converter will provide the implementation function with
a standard :c:type:`PyTupleObject`.
For example::

   /*[clinic input]
   var_positional_sample

       spam: int
       *args: tuple
   [clinic start generated code]*/
```

skirpichev: IMO, it's too verbose.  But OK.

PR #1631, opened 2025-08-13
Group container instructions in Setup into a single section

brettcannon:
Having two top-level sections that both use the container seemed unnecessary.

Also update an internal link to make it easier to find the container definition file.

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -386,14 +386,14 @@ compiler just like building for :ref:`Unix <unix-compiling>` as well as:
 1. A C compiler that can target WebAssembly (for example, `WASI SDK`_)
 2. A WASI host/runtime (for example, Wasmtime_)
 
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -1124,17 +1124,26 @@ every rule.
...
+additional tools on your machine. All approaches use the container defined in
+the `cpython-devcontainers repo`_ in some way.
+
 .. _using-codespaces:
 
 Contribute using GitHub Codespaces
-==================================
+----------------------------------
 
 .. c_codespaces_start
```
AA-Turner, line 1141: What's this comment for?

AA-Turner (approved):

getting-started/setup-building.rst
brettcannon: It was already there, so 🤷. There's a similar comment above for `c_directory_structure_start` and `..._end`.

PR #1636, opened 2025-08-15
Add boolean to the style guide

hugovk:
Fixes https://github.com/python/docs-community/issues/129, which has been open for 11 months with 10 votes in favour of the Google guidance (Ned's comment + 9 👍), and no down votes, and no other suggestions.

----
📚 Documentation preview 📚: https://cpython-devguide--1636.org.readthedocs.build/documentation/style-guide/#specific-words

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -66,6 +66,10 @@ Specific words
 Some terms and words deserve special mention. These conventions should
 be used to ensure consistency throughout the documentation:
 
+boolean
+  Lowercase in most cases.
+  Uppercase for Boolean mathematics and Boolean logic.
+
 C API
   Python's `API <https://docs.python.org/3/c-api/>`_ used by C programmers
   to write extension modules. All caps and unhyphenated.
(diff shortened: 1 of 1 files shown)


documentation/style-guide.rst
```diff
@@ -66,6 +66,10 @@ Specific words
 Some terms and words deserve special mention. These conventions should
 be used to ensure consistency throughout the documentation:
 
+boolean
+  Lowercase in most cases.
+  Uppercase for Boolean mathematics and Boolean logic.
```
encukou: By my reading of the Google guidance, this should be:
```suggestion
  Lowercase in most cases.
  Uppercase for *Boolean mathematics* and *Boolean logic*.
  To refer to the Python or C data type, prefer using the exact, abbreviated
  name with appropriate markup (for example, ``:type:`bool```).
```

AA-Turner: To avoid saying "case" twice in one line
```suggestion
  Lowercase in most instances.
  Uppercase for *Boolean mathematics* and *Boolean logic*.
  To refer to the Python or C data type, prefer using the exact,
  abbreviated name with appropriate markup (for example, ``:type:`bool```).
```

willingc (approved):
I approve with adoption of @encukou / @AA-Turner's suggestion.

In general, I think using interchangeably is fine since there is no industry consensus of capitalization.

ezio-melotti (approved):
LGTM (once the suggested comments are applied).

PR #1675, opened 2025-10-29
Document that there are alternative dev containers

brettcannon:
Purposefully didn't go into too much detail as it's a bit of an advanced topic.

As well, fix a warning being triggered from `conf.py`.

----
📚 Documentation preview 📚: https://cpython-devguide--1675.org.readthedocs.build/


encukou:
Looks reasonable to me, thanks!

PR #1679, opened 2025-11-01
generative-ai: Add anti pattern example

corona10:
----
📚 Documentation preview 📚: https://cpython-devguide--1679.org.readthedocs.build/


getting-started/generative-ai.rst
```diff
@@ -24,3 +24,9 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Anti-patterns
+=============
+- While AI-assisted tools such as autocompletion can enhance productivity, they sometimes rewrite entire code blocks instead of making small, focused edits.
+  This can make it more difficult to review changes and to fully understand both the original intent of the code and the rationale behind the new modifications.
+  Please keep maintaining consistency with the original code helps preserve clarity, traceability, and meaningful review.
```
nedbat:
```suggestion
  Maintaining consistency with the original code helps preserve clarity, traceability, and meaningful review.
```
We also might want to mention (and link to?) the idea that we want to avoid churn.

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,9 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Anti-patterns
```
nedbat: Do we have a second anti-pattern to include in the list?  Other bad things: hallucinations, incorrectly changing tests instead of fixing code, ...

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,11 @@ Unacceptable uses
...
+=============
+- While AI-assisted tools such as autocompletion can enhance productivity, they sometimes rewrite entire code blocks instead of making small, focused edits.
+  This can make it more difficult to review changes and to fully understand both the original intent of the code and the rationale behind the new modifications.
+  Maintaining consistency with the original code helps preserve clarity, traceability, and meaningful reviews and also helps us avoid unnecessary code churn.
+- Sometimes AI assisted tools make failing unit tests pass by altering or bypassing the tests rather than addressing the underlying problem in the code.
+  Such changes do not represent a real fix and should be avoided.
```
Mariatta: I don't understand what should be avoided:
- avoid using AI-assisted tools?
- avoid rewriting code blocks?
- avoid using Ai-assisted tools that rewrite code blocks?
What is the recommended approach instead if we are telling people not to do a certain thing?

I would suggest starting the paragraph with something like: 
"We discourage the use of AI when ..... "
 
Then followed with the explanation of why.
Currently this section reads with the "why" first.

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,11 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Anti-patterns
+=============
+- While AI-assisted tools such as autocompletion can enhance productivity, they sometimes rewrite entire code blocks instead of making small, focused edits.
```
gpshead: I'd actually tighten this phrasing up a bit, we don't need to explain what shape AI tools can take or what they can do (anyone reading this knows that).

I find this list of anti-patterns to not be AI specific.  We already reject PRs coming from people who make larger than necessary changes that are difficult to review.  So I'd suggest wording it as a "positive desired thing, not negative example" phrasing, starting something like this:

- Make small focused edits, prefer not rewriting entire code blocks.
   This makes it easier to review changes and fully understand both the original intent of the code and the rationale behind the new modifications.
   ...

getting-started/generative-ai.rst
gpshead: I think this sentence was only intended to be tied to the false testing / false success bullet it is part of.  Rather than wording it as "should be avoided" the right thing for users to do is to request that authors have reviewed the work in detail for problems like this before proposing it as a PR.

_"Such changes do not represent a real fix. Authors must review the work done by AI tooling in detail to ensure it actually makes sense before proposing it as a PR."_

---

I could maybe suggest adding an AI use tip for this situation like _"iterate with your AI agent on the quality and utility of the change and tests to understand if the issue is actually fixed"_ but I'd leave that out for now as topic is maybe larger and constantly evolving, beyond the scope of this PR.

I'm trying to focus on a proactive positive outcome based wording rather than a negative like "avoid" or "discourage".

The reason being, the statements in here don't reflect all AI tooling. This devguide doc is about the use of AI, so we should assume those who read it are interested in or already using AI and are here to learn how to do it in an appropriate and productive manner.

Statements like "we discourage use of AI when ..." are likely to be outdated very quickly and discourage things that won't be true for long.  Thus my desire to focus this on what we do want and why it is important for project health and reviewer respect.

getting-started/generative-ai.rst
gpshead: i think you covered the incorrectly changing tests instead of fixing one. Good call. I like having that in the list.

I don't personally see that so often anymore with Opus 4.1 or Sonnet 4.5, but older or lower end models tend to reward hack their way into that state more often - it's a thing we'll likely encounter coming from new contributors for a while as they can't be assumed to be using the latest and greatest.

willingc (commented):
It would be interesting to add a link to: https://arxiv.org/pdf/2507.09089 and possibly https://www.fast.ai/posts/2025-10-30-build-to-last.html

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,11 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Anti-patterns
```
willingc: I would use a different title than antipatterns. I would like to move toward "Considerations or Keys for Success".

I would start by linking to the section about submitting a quality PR:
- consider whether the change is necessary
- small, focused changes
- tests that exercise the change
- style follows the existing code

AI coding tools will compose code. The contributor, not the AI tool, is responsible for the quality and validity of code submitted with a PR.

Take the time to consider the code changes suggested by an AI tool. Respect the reviewer's time by validating your PR before submission.

willingc:
I love the suggested opening @nedbat.

corona10:
A lot of opinions!, I will address reviews by this weekend!

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,16 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Considerations or Keys for Success
+==================================
```
hugovk: Use sentence case for headers:

https://devguide.python.org/documentation/style-guide/#capitalization
```suggestion
Considerations or keys for success
==================================

```

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,16 @@ Unacceptable uses
...
+- While AI assisted tools such as autocompletion can enhance productivity, they sometimes rewrite entire code blocks instead of making small, focused edits.
+  This can make it more difficult to review changes and to fully understand both the original intent of the code and the rationale behind the new modifications.
+  Maintaining consistency with the original code helps preserve clarity, traceability, and meaningful reviews and also helps us avoid unnecessary code churn.
+- Sometimes AI assisted tools make failing unit tests pass by altering or bypassing the tests rather than addressing the underlying problem in the code.
+  Such changes do not represent a real fix. Authors must review the work done by AI tooling in detail to ensure it actually makes sense before proposing it as a PR.
+- Please keep following principles for the quality of your contributions in mind whether you use generative AI or not:
+  - Consider whether the change is necessary.
+  - Make minimal, focused changes.
+  - Follow existing coding style and patterns.
+  - Write tests that exercise the change.
```
hugovk: The secondary bullet list isn't rendering:
https://cpython-devguide--1679.org.readthedocs.build/getting-started/generative-ai/#considerations-or-keys-for-success

This newline might fix it?

Also remove full stops to match the earlier list, and usually no need for "please" in technical docs.

```suggestion
- Keep the following principles for the quality of your contributions in mind whether you use generative AI or not:

  - Consider whether the change is necessary
  - Make minimal, focused changes
  - Follow existing coding style and patterns
  - Write tests that exercise the change
```

nedbat: "Keep the following ..."

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,16 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Considerations or Keys for Success
```
nedbat: I think a simpler heading works better:
```suggestion
Considerations for success
```

getting-started/generative-ai.rst
hugovk: Added to the suggestion.

getting-started/generative-ai.rst
```diff
@@ -24,3 +24,17 @@ Unacceptable uses
 Maintainers may close issues and PRs that are not useful or productive, including
 those that are fully generated by AI. If a contributor repeatedly opens unproductive
 issues or PRs, they may be blocked.
+
+Considerations for success
+==================================
```
willingc:
```suggestion
Considerations for success
==========================
```

getting-started/generative-ai.rst
```diff
@@ -4,10 +4,10 @@
 Generative AI
 =============
 
-Generative AI has evolved rapidly over the past decade and will continue in the future.
-Using generative AI and large language models (LLMs) can be helpful tools for contributors.
-Their overuse can also be problematic, such as generation of incorrect code, inaccurate documentation, and unneeded code churn.
-Discretion, good judgement, and critical thinking **must** be used when opening issues and pull requests.
+Generative AI has evolved rapidly and can be impressive. As with using any tool, the resulting contribution is
```
willingc:
```suggestion
Generative AI tools have evolved rapidly, and their suggested results can be helpful. As with using any tool, the resulting contribution is
```

willingc: Let's be explicit that Gen AI, agents, etc. are tools. I would prefer not to make the value judgement that the results are impressive.

willingc (approved):
A couple of additional suggestions which from my perspective make the wording acceptable to me.

willingc:
Thanks @corona10. I hope this will help give contributors more context.

PR #1681, opened 2025-11-05
Clarify development cycle

hugovk:
----
📚 Documentation preview 📚: https://cpython-devguide--1681.org.readthedocs.build/

### DIFF
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -68,7 +68,7 @@ Maintenance branches
 A branch for a previous feature release, currently being maintained for bug
 fixes, or for the next feature release in its
 :ref:`beta <beta>` or :ref:`release candidate <rc>` stages.
-There is usually either one or two maintenance branches at any given time.
+There are usually either one or two maintenance branches at any given time.
 After the final release of a new minor version (3.x.0), releases
 produced from a maintenance branch are called **bugfix** or **maintenance**
 releases; the terms are used interchangeably. These releases have a


developer-workflow/development-cycle.rst
```diff
@@ -68,8 +68,8 @@ Maintenance branches
 A branch for a previous feature release, currently being maintained for bug
 fixes, or for the next feature release in its
 :ref:`beta <beta>` or :ref:`release candidate <rc>` stages.
-There is usually either one or two maintenance branches at any given time for
-Python 3.x.  After the final release of a new minor version (3.x.0), releases
+There is usually either one or two maintenance branches at any given time.
+After the final release of a new minor version (3.x.0), releases
```
hugovk: It's always Python 3.x, we don't care about 2.x any more.

If there's a 4.x, and things change, we'll update it then.

developer-workflow/development-cycle.rst
```diff
@@ -204,8 +204,7 @@ Final
 ^^^^^
 
 When a final release is being cut, only the release manager (RM) can make
-changes to the branch.  After the final release is published, the full
-:ref:`development cycle <stages>` starts again for the next minor version.
```
hugovk, line 208: Incorrect.

After the final release is published in October, the dev cycle for the next one had already started back in May.

developer-workflow/development-cycle.rst
```diff
@@ -68,8 +68,8 @@ Maintenance branches
 A branch for a previous feature release, currently being maintained for bug
 fixes, or for the next feature release in its
 :ref:`beta <beta>` or :ref:`release candidate <rc>` stages.
-There is usually either one or two maintenance branches at any given time for
-Python 3.x.  After the final release of a new minor version (3.x.0), releases
+There is usually either one or two maintenance branches at any given time.
```
AA-Turner: There are ... branches, rather than 'is':

```suggestion
There are usually either one or two maintenance branches at any given time.
```

developer-workflow/development-cycle.rst
AA-Turner: Maybe add that this period starts after RC1? Or at least it has for the last couple of releases.

hugovk: The new dev cycle begins after beta 1, when `main` forks to a 3.x branch.

I think let's keep these explanations simpler, about the current cycle, rather than matching them up to previous/next cycles.

PR #1685, opened 2025-11-10
Replace local release-cycle.json with PEPs repo version

hugovk:
We now generate and publish `release-cycle.json` from the PEPs repo (https://github.com/python/peps/pull/4331) so can use that directly and don't need to duplicate this data.

----
📚 Documentation preview 📚: https://cpython-devguide--1685.org.readthedocs.build/

### DIFF
--- a/_tools/generate_release_cycle.py
+++ b/_tools/generate_release_cycle.py
@@ -41,7 +41,7 @@ def parse_version(ver: str) -> list[int]:
 
 
 @cache
-def get_versions() -> str:
+def get_versions() -> dict[str, dict[str, str | int]]:
     with urlopen("https://peps.python.org/api/release-cycle.json") as in_file:
         return json.loads(in_file.read().decode("utf-8"))
 


include/release-cycle.json
StanFromIreland, line 1: This will break scripts that depend on this file, e.g. your GitHub tools.

hugovk: Yeah, they also need updating to point to https://peps.python.org/api/release-cycle.json

Let's update/notify the ones we know about before merging.

conf.py
```diff
@@ -178,8 +179,8 @@
 
 # Dynamically expose the Python version associated with the "main" branch.
 # Exactly one entry in ``release-cycle.json`` should have ``"branch": "main"``.
-with open("include/release-cycle.json", encoding="UTF-8") as _f:
-    _cycle = json.load(_f)
+with urlopen("https://peps.python.org/api/release-cycle.json") as _f:
+    _cycle = json.loads(_f.read().decode("utf-8"))
```
AA-Turner, line 183: @hugovk can we split out this part of the PR & merge it?

A

hugovk: It's been a fortnight and most of the update PRs, and all the important ones, have been merged. Let's just merge the whole thing :)

hugovk:
It's been a fortnight and most of the update PRs, and all the important ones, have been merged. 

Ready for merge!

_tools/generate_release_cycle.py
```diff
@@ -7,6 +7,9 @@
 import csv
 import datetime as dt
 import json
+from functools import cache
+from pathlib import Path
+from urllib.request import urlopen
 
 import jinja2
```
AA-Turner, line 14:
```suggestion

import jinja2
import requests
```

_tools/generate_release_cycle.py
```diff
@@ -37,12 +40,17 @@ def parse_version(ver: str) -> list[int]:
     return [int(i) for i in ver["key"].split(".")]
 
 
+@cache
+def get_versions() -> str:
+    with urlopen("https://peps.python.org/api/release-cycle.json") as in_file:
+        return json.loads(in_file.read().decode("utf-8"))
```
AA-Turner, line 46:
```suggestion
    resp = requests.get("https://peps.python.org/api/release-cycle.json")
    resp.raise_for_status()
    return json.loads(resp.content)
```

conf.py
```diff
@@ -1,4 +1,5 @@
 import json
+from urllib.request import urlopen
```
AA-Turner, line 2:
```suggestion

import requests
```

conf.py
AA-Turner, line 183:
```suggestion
_resp = requests.get("https://peps.python.org/api/release-cycle.json")
_resp.raise_for_status()
_cycle = json.loads(resp.content)
```

_tools/generate_release_cycle.py
```diff
@@ -37,12 +40,17 @@ def parse_version(ver: str) -> list[int]:
     return [int(i) for i in ver["key"].split(".")]
 
 
+@cache
+def get_versions() -> str:
```
AA-Turner:
```suggestion
def get_versions() -> dict[str, dict[str, str | int]]:
```

_tools/generate_release_cycle.py
hugovk: For these one-shot requests, what's the benefit of adding a dependency instead of using the stdlib?

If we do add one, my vote is for `urllib3` over `requests`.

PR #1687, opened 2025-11-11
Make no-force-push advice linkable

hugovk:
Promote the note to something that can be linked to, so we can easily share it with contributors who force push their PRs.

Add an extra sentence: 

"When the PR is merged, everything will be squashed into a single commit."

----
📚 Documentation preview 📚: https://cpython-devguide--1687.org.readthedocs.build/


itamaro (approved):
Nit comments: 
1. Maybe say "If the PR is merged" instead of "When ..."
2. Consider qualifying the guidance with "for published PRs" (afaik it's ok to force push to draft PRs)

PR #1696, opened 2025-11-19
Run `pre-commit autoupdate`

StanFromIreland:
----
📚 Documentation preview 📚: https://cpython-devguide--1696.org.readthedocs.build/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.7
+    rev: v0.14.5
     hooks:
       - id: ruff
         name: Run Ruff (lint)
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
StanFromIreland, line 35: This isn't working since we aren't using [pre-commit.ci](https://pre-commit.ci/).

hugovk: Let's turn it on :)

@jacobcoffee Please could you enable it for this repo and https://github.com/python/python-docs-theme? I've sent a request for them both via the app.

StanFromIreland: Can we add python-docs-pl to that list please?

JacobCoffee: all should be added

StanFromIreland:
Thanks Jacob, closing in favour of the automatic PR then.

PR #1700, opened 2025-11-24
Translations: Add guide on merging back in translations

### LINKED ISSUE #1697: Feature: Moving docs and the impact on translations

Recently there have been a lot of talks about some quite large docs moves. Unfortunately for translations, the translated messages will not be moved automatically, but removed/lost (note that this excludes TX translations since TX memory will fill them back up in their new location). As such, it would be best if teams are notified such a move has occurred so that they can recover them and not use up their valuable volunteer time re-translating things.

There are a two things I would like to do, this list may grow as I work on them:

- [ ] Request that _Docs contributors_ notify translations teams after a big move occurs.
- [ ] Add a section for coordination teams on merging back in (i.e. recovering) old translations.

I will be working on this.

_No response_

### Additional context

_No response_

StanFromIreland:
For #1697

----
📚 Documentation preview 📚: https://cpython-devguide--1700.org.readthedocs.build/

### DIFF
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -371,7 +371,7 @@ Note that this is not necessary for Transifex-based translations, as Transifex's
 translation memory will automatically copy them over.
 
 The :pypi:`pomerge` tool merges translations between PO files by matching
-message, regardless of file paths. To use it, first install the package:
+messages, regardless of file paths. To use it, first install the package:
 
 .. code-block:: bash
 
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -361,5 +361,51 @@ If you would like to coordinate, open a pull request in the
 at the top of this page, and ping ``@python/editorial-board``.
 
 
+How do I merge translations back in after docs are moved?
+---------------------------------------------------------
+
+When docs are moved in the CPython repository, translated messages are not
+moved automatically and will be lost.
+They can be recovered by merging them into the new locations.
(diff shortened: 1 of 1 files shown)


documentation/translations/translating.rst
```diff
@@ -361,5 +361,51 @@ If you would like to coordinate, open a pull request in the
...
+message, regardless of file paths. To use it, first install the package:
+
+.. code-block:: bash
+
+    pip install pomerge
+
+Then, merge translations from a specific commit (replace :samp:`{COMMIT_HASH}`
+with the commit hash from before the files were moved):
+
+.. TODO: Provide Windows instructions.
```
StanFromIreland, line 383: I would greatly appreciate if someone could do this bit, since I don't have a clue:-)

zahraazizi5856-art: sasa

m-aciek (approved):
This looks good to me 👍

I think there are also offline translation memory tools that potentially people could use, so I would consider rephrasing the first paragraph for more general note that this is a workaround for lack of translation memory system.

documentation/translations/translating.rst
willingc: @StanFromIreland I'm not sure either. I like the overall changes of this PR. Nice work. Let's land as is and address at a later time. Thanks.

willingc (approved):

StanFromIreland:
Thanks Carol and Maciej :-)

PR #1708, opened 2025-11-29
Make release cycle charts reusable

hugovk:
Right now, the release cycle SVGs are monochrome when viewed as standalone images:

image

This is because all the colours are defined in the website's own `.css` file.

This PR moves the colours into the SVG files themselves, except for the theme-aware overrides using things like `--color-foreground-primary` from Furo, needed for dark/light mode support.

Also serve these images, so we can display them on https://www.python.org/downloads/ where our end users visit, and not just in the CPython developer's guide. Re: https://endoflife.date/recommendations#bad---python

Needed for PR: [pythondotorg#2837](https://github.com/python/pythondotorg/pull/2837)

----
📚 Documentation preview 📚: https://cpython-devguide--1708.org.readthedocs.build/

### DIFF
--- a/_static/devguide_overrides.css
+++ b/_static/devguide_overrides.css
@@ -6,34 +6,6 @@
     height: 110px;
 }
 
-/* Release cycle chart */
-.release-cycle-chart .release-cycle-year-line {
-    stroke: var(--color-foreground-primary);
-}
-
-.release-cycle-chart .release-cycle-year-text {
-    fill: var(--color-foreground-primary);
-}
-
-.release-cycle-chart .release-cycle-today-line {
-    stroke: var(--color-brand-primary);
-}
-
-.release-cycle-chart .release-cycle-row-shade {
-    fill: var(--color-background-item);
-}
-
-.release-cycle-chart .release-cycle-version-label {
-    fill: var(--color-foreground-primary);
-}
-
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -37,6 +37,10 @@
     "source_branch": "main",
 }
 html_static_path = ['_static']
+html_extra_path = [
+    'include/release-cycle.svg',
+    'include/release-cycle-all.svg',
+]
```
AA-Turner: Why is this needed? Semantically, I think we should prefer static path to extra path: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path

> A list of paths that contain extra files not directly related to the documentation

A

hugovk: > Why is this needed? 

So that they're served at https://devguide.python.org/release-cycle.svg


> Semantically, I think we should prefer static path to extra path: [sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path)
> 
> > A list of paths that contain extra files not directly related to the documentation
> 
> A

Okay, so have the generate script save them to _static, and have them served at https://devguide.python.org/_static/release-cycle.svg ?

I've pushed a commit for this.

hugovk:
Any further comments?

PR #1712, opened 2025-12-02
Remove empty internals doc pages

StanFromIreland:
Based on the discussion in the docs community meeting. I propose this, in short, I kept the only remaining content and moved it to an `internals.rst` file, removing the old "See the InternalDocs" pages.

----
📚 Documentation preview 📚: https://cpython-devguide--1712.org.readthedocs.build/internals/

### DIFF
--- a/internals.rst
+++ b/internals.rst
@@ -29,7 +29,7 @@ building your understanding of CPython internals and its evolution:
     "`A guide from parser to objects, observed using Eclipse`_", "Code walk from Parser, AST, Sym Table and Objects", Prashanth Raghu, 2.7.12
     "`CPython internals: A ten-hour codewalk through the Python interpreter source code`_", "Code walk from source code to generators", Philip Guo, 2.7.8
 
-.. _CPython's InternalDocs: https://github.com/python/cpython/tree/main/InternalDocs
+.. _CPython's InternalDocs: https://github.com/python/cpython/blob/main/InternalDocs/README.md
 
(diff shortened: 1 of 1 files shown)


StanFromIreland:
> Isn't *this* a bit of the restructuring?

That was my thinking too, we'll end up removing these pages at some point, why not do that now? "Baby steps" as some say.

> move “Source code layout” to CPython's `InternalDocs`

Sounds like a good idea, I'll open a PR.

> The URLs will still work because they are now redirects in conf.py, no?

Yes, I don't want to leave dead links, so I made them redirect to the single page. We were referencing one of the empty pages in the CPython docs till a month or so ago, so I assume there are quite a few more in the wild.

internals.rst
```diff
@@ -72,6 +29,7 @@ building your understanding of CPython internals and its evolution:
     "`A guide from parser to objects, observed using Eclipse`_", "Code walk from Parser, AST, Sym Table and Objects", Prashanth Raghu, 2.7.12
     "`CPython internals: A ten-hour codewalk through the Python interpreter source code`_", "Code walk from source code to generators", Philip Guo, 2.7.8
 
+.. _CPython's InternalDocs: https://github.com/python/cpython/tree/main/InternalDocs
```
encukou: Let's link directly to the README.
```suggestion
.. _CPython's InternalDocs: https://github.com/python/cpython/blob/main/InternalDocs/README.md
```

StanFromIreland:
Thanks Petr!

PR #1713, opened 2025-12-09
Replace deprecated workflow with RtD app

hugovk:
For https://github.com/python/core-workflow/issues/587.

This workflow is deprecated, and uses the `pull_request_target` target which can be insecure. 

Let's test replacing it with the Read the Docs app. If it goes well, we can replace it in the other repos.

@python/organization-owners I've sent a request to install the RtD app for just this repo, please could you enable it?

----
📚 Documentation preview 📚: https://cpython-devguide--1713.org.readthedocs.build/


JacobCoffee:
@hugovk done

StanFromIreland (approved):
LGTM

PR #1722, opened 2026-01-06
Add free-threaded to style guide (#1353)

### LINKED ISSUE #1353: Disallow "No-GIL" from the glossary of specific words
The Steering Council, on accepting [PEP 703 (Making the Global Interpreter Lock Optional in CPython)](https://discuss.python.org/t/pep-703-making-the-global-interpreter-lock-optional-in-cpython-acceptance/37075), said:

> We want to avoid negatives in terms and flags and such, so we won’t get into double-negative terrain (like we do when we talk about ‘non no-GIL’). We’d like a positive, clear term to talk about the no-GIL build, and we’re suggesting ‘free-threaded’. (Relatedly, that’s why the build mode/ABI letter is ‘t’ and not ‘n’; that change was already made.)

"No-GIL" comes up regularly; shall we add this to the [style guide](https://devguide.python.org/documentation/style-guide/) so we have something to point to beyond the lengthy acceptance message?

arunkumargururaj07-star:
Closes #1353
Added 'free-threaded' to the style guide as the preferred term for PEP 703 to avoid double-negatives.

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -77,12 +77,12 @@ C API
   to write extension modules. All caps and unhyphenated.
 
 CPU
-    Central processing unit. No need to spell out.
+   Central processing unit. No need to spell out.
 
 free-threaded
-    The preferred term for the build mode that makes the global interpreter
-    lock (GIL) optional (per :pep:`703`). Avoid using "No-GIL" to prevent
-    double negatives (for example, "non-no-GIL").
+   The preferred term for the build mode that makes the global interpreter
+   lock (GIL) optional (per :pep:`703`). Avoid using "No-GIL" to avoid
+   double negatives (for example, "non-no-GIL").
 
 POSIX
    The name assigned to a particular group of standards. This is always
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -82,7 +82,10 @@ CPU
 POSIX
    The name assigned to a particular group of standards. This is always
    uppercase.
-
+free-threaded
+    The preferred term for the build mode that makes the Global Interpreter
+    Lock (GIL) optional (per PEP 703). Avoid using "No-GIL" to prevent
+    double-negatives (e.g., "non-no-GIL").
 Python
    The name of our favorite programming language is always capitalized.
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -83,9 +83,9 @@ POSIX
    The name assigned to a particular group of standards. This is always
    uppercase.
 free-threaded
-    The preferred term for the build mode that makes the Global Interpreter
-    Lock (GIL) optional (per PEP 703). Avoid using "No-GIL" to prevent
-    double-negatives (e.g., "non-no-GIL").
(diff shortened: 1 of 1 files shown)


documentation/style-guide.rst
```diff
@@ -82,7 +82,10 @@ CPU
 POSIX
    The name assigned to a particular group of standards. This is always
    uppercase.
-
+free-threaded
+    The preferred term for the build mode that makes the Global Interpreter
+    Lock (GIL) optional (per PEP 703). Avoid using "No-GIL" to prevent
+    double-negatives (e.g., "non-no-GIL").
```
hugovk:
```suggestion
    The preferred term for the build mode that makes the global interpreter
    lock (GIL) optional (per :pep:`703`). Avoid using "no-GIL" to prevent
    double negatives (for example, "non-no-GIL").
```

Also list alphabetically and include whitespaces between entries.

documentation/style-guide.rst
```diff
@@ -1,4 +1,4 @@
-.. _style-guide:
+﻿.. _style-guide:
```
StanFromIreland: Why the unrelated change?

documentation/style-guide.rst
```diff
@@ -77,7 +77,12 @@ C API
   to write extension modules. All caps and unhyphenated.
 
 CPU
-   Central processing unit. No need to spell out.
+    Central processing unit. No need to spell out.
```
StanFromIreland: Can you revert the unrelated changes?

documentation/style-guide.rst
```diff
@@ -77,7 +77,12 @@ C API
...
+
+free-threaded
+    The preferred term for the build mode that makes the global interpreter
+    lock (GIL) optional (per :pep:`703`). Avoid using "No-GIL" to prevent
+    double negatives (for example, "non-no-GIL").
```
StanFromIreland: The indentation is incorrect.

documentation/style-guide.rst
```diff
@@ -77,7 +77,12 @@ C API
   to write extension modules. All caps and unhyphenated.
 
 CPU
-   Central processing unit. No need to spell out.
+    Central processing unit. No need to spell out.
+
+free-threaded
+    The preferred term for the build mode that makes the global interpreter
+    lock (GIL) optional (per :pep:`703`). Avoid using "No-GIL" to prevent
```
```suggestion
    lock (GIL) optional (per :pep:`703`). Avoid using "No-GIL" to avoid
```
A little suggestion, "prevent" seems a little odd here.

arunkumargururaj07-star: @StanFromIreland I've applied all your feedback. The PR has been updated with the fixes.

documentation/style-guide.rst
StanFromIreland: This is still changed.

willingc:
```suggestion
```

willingc (commented):
Thanks @arunkumargururaj07-star for the addition.

willingc (approved):

PR #1761, opened 2026-03-08
Overhaul buildbot worker setup documentation

stratakis:
As I've recently had to setup again the aarch64 buildbots, I've taken a shot at re-reading and verifying the current docs. I've rewriten the Linux buildbot worker setup instructions with per-distro nested tabs, all of which I've verified in containers and VMs.

Also added recommended buildbot.tac settings, FreeBSD rc.d instructions, and pip/manual setup for distros without available buildbot packages.

Plus some minor MacOS fixups.

### DIFF
--- a/testing/new-buildbot-worker.rst
+++ b/testing/new-buildbot-worker.rst
@@ -96,9 +96,10 @@ the steps below as appropriate if you choose that path.
          dnf install epel-release epel-next-release
          dnf install buildbot-worker
 
-      The RPM creates a ``buildbot-worker`` system user (nologin shell, home
-      at ``/var/lib/buildbot/worker``), installs a templated systemd unit
-      ``buildbot-worker@.service``, and creates ``/var/lib/buildbot/worker/``.
+      The RPM creates a ``buildbot-worker`` system user, installs a
+      templated systemd unit ``buildbot-worker@.service``, and creates
+      ``/var/lib/buildbot/worker/`` as the base directory for worker
+      instances.
 
       .. tip::
 
@@ -133,7 +134,8 @@ the steps below as appropriate if you choose that path.
 
       The package creates a ``buildbot`` system user, installs a templated
       systemd unit ``buildbot-worker@.service``, and creates
-      ``/var/lib/buildbot/workers/``.
+      ``/var/lib/buildbot/workers/`` as the base directory for worker
+      instances.
 
       .. tip::
 
@@ -211,10 +213,10 @@ the steps below as appropriate if you choose that path.
 
          pkg install devel/py-buildbot-worker
 
-      The port creates a ``buildbot`` system user (nologin shell, home at
-      ``/var/db/buildbot``) and installs an ``rc.d`` script at
-      ``/usr/local/etc/rc.d/buildbot-worker`` with profile support.
-      The default base directory is ``/var/db/buildbot/workers``.
+      The package creates a ``buildbot`` system user, installs an
+      ``rc.d`` service with profile support, and creates
+      ``/var/db/buildbot/workers/`` as the base directory for worker
+      instances.
 
       Create the worker (replace ``WORKERNAME`` and ``WORKERPASSWD`` with
       the credentials provided to you from your buildmaster-config issue)::
--- a/testing/new-buildbot-worker.rst
+++ b/testing/new-buildbot-worker.rst
@@ -138,37 +138,6 @@ the steps below as appropriate if you choose that path.
 
          systemctl enable --now buildbot-worker@WORKERNAME.service
 
-   .. tab:: FreeBSD
-
-      ::
-
-         pkg install devel/py-buildbot-worker
-
-      The port creates a ``buildbot`` system user (nologin shell, home at
-      ``/var/db/buildbot``) and installs an ``rc.d`` script at
-      ``/usr/local/etc/rc.d/buildbot-worker`` with profile support.
-      The default base directory is ``/var/db/buildbot/workers``.
-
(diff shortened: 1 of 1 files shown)


testing/new-buildbot-worker.rst
```diff
@@ -60,28 +60,156 @@ Setting up the buildbot worker
...
+             buildbot-api.python.org:9020 WORKERNAME WORKERPASSWD
+
+      Edit ``info/admin``, ``info/host``, and ``buildbot.tac`` in the worker
+      directory (see below for recommended settings).
+
+      Enable and start the service::
+
+         systemctl enable --now buildbot-worker@WORKERNAME.service
+
+   .. tab:: FreeBSD
```
encukou: FreeBSD is not Linux; can it get its own tab?

testing/new-buildbot-worker.rst
```diff
@@ -129,89 +257,111 @@ if you choose that path.
...
-     If you installed the buildbot-worker through
-     your package manager it would be::
-
-       [Unit]
-       Description=Buildbot Worker
-       Wants=network-online.target
-       After=network-online.target local-fs.target
+   If you installed via a distro package (Fedora, RHEL, CentOS, Debian,
+   Ubuntu, or FreeBSD), the service was already enabled in the installation
+   step above.
```
encukou: If we're adding nested tabs above, they might work here as well: *Distro package*, *systemd*, *SysV init*, *Cronjob* :)

testing/new-buildbot-worker.rst
encukou: How about this?
```suggestion
Many tests won't run in this configuration, since they require
substantially more memory, but these resources should be sufficient.
Builders with minimal settings might need more maintenance: they check
that Python's resource-hungry tests are tagged and skipped correctly.
```

testing/new-buildbot-worker.rst
stratakis: Sure. In my head I kinda get the idea that I should add either a BSD tab, or a Unix tab and add FreeBSD, with the possibility of extending to other flavors in the future, either by me or others. Or just a FreeBSD tab for now on its own. Thoughts?

testing/new-buildbot-worker.rst
stratakis: Hmm right, that could work.

testing/new-buildbot-worker.rst
encukou: I'd say FreeBSD, to avoid tab groups with a single tab. If we add more BSDs/Unix-likes later, it can be split up.

encukou:
Everything's squashed.
Don't force push (so we don't need to re-review everything). And don't worry about rebasing before a push -- neatly split commits are only nice-to-have.

testing/new-buildbot-worker.rst
stratakis: I believe have Unix there would be better just so next contributors don't have to shift around tabs, they can just create a new one. And I might add myself other BSDs now that I got a bit the hang of testing them.

encukou (approved):
Looks good, thanks!
I'll merge later this week if there are no objections.

encukou:
> Would that be of interest?

I'd definitely merge them, on the chance that they help someone.

I take the instructions as hints -- they'll inevitably go out of date and they'll inevitably not apply exactly to all environments (and the environments we want tests for are the weird ones :)
If someone goes through the setup process, it's very valuable to update the instructions. Going through the process just to update them might not be worth your time.

testing/new-buildbot-worker.rst
```diff
@@ -129,89 +320,126 @@ if you choose that path.
...
+      Install these as ``/etc/init.d/buildbot-worker`` and
+      ``/etc/default/buildbot-worker`` respectively, then configure the
+      worker instances in the default file.
 
-   * Alternatively you can create a cronjob. Add the following line to ``/etc/crontab``::
+   .. tab:: Cronjob
 
-         @reboot buildbot-worker restart /path/to/buildarea
+      If neither systemd nor a SysV init script is practical, you can use
+      a cronjob.  Add the following line to ``/etc/crontab``::
```
hugovk:
```suggestion
   .. tab:: cron job

      If neither systemd nor a SysV init script is practical, you can use
      a cron job.  Add the following line to ``/etc/crontab``::
```

testing/new-buildbot-worker.rst
```diff
@@ -129,89 +320,126 @@ if you choose that path.
...
+         service buildbot-worker status
+         service buildbot-worker restart
+
+      On OpenBSD::
+
+         rcctl check buildbot_worker
+         rcctl restart buildbot_worker
+
+      If you installed via pip, you will need to write an ``rc.d`` script
+      or use the cronjob approach described in the Linux tab.
```
hugovk:
```suggestion
      or use the cron job approach described in the Linux tab.
```

hugovk (commented):
This section: https://cpython-devguide--1761.org.readthedocs.build/testing/new-buildbot-worker/index.html#conventional-always-on-machines

Has a Unix tab with FreeBSD and OpenBSD subtabs:

Image

---

But this section: https://cpython-devguide--1761.org.readthedocs.build/testing/new-buildbot-worker/index.html#service-management

Has a Unix tab with a single rc.d tab, but then separate FreeBSD and OpenBSD instructions:

Image

Shall we get rid of the rc.d tab, because there's no other sibling tabs?

And instead have FreeBSD and OpenBSD subtabs?

hugovk (approved):
Thank you!

vstinner:
Thanks @stratakis for enhanced documentation!

PR #1635, opened 2025-08-14
Prune entries from the Experts Index

AA-Turner:
We suggest and recommend using the Experts Index for contributors to find a relevant committer/core developer to review a PR or comment on an issue/proposal. Listing names of inactive committers, those who have left the team, or those that never moved from BPO to GitHub can create a frustrating contributor experience.

I've split this change into 5 commits:

* Remove all entries marked as "inactive"
* Remove all those marked "inactive" by the voters script (i.e. no commits in the last 2 years and no explict 'stay active' response)
* Remove all those who are marked as no longer a core developer (with a `left = "..."` entry)
* Remove all former BPO usernames (`^` entries)
* Remove all those with no commits in the last 5 years, regardless of 'stay active' voter status

We can of course re-add people at any time, this is more of a housekeeping activity than anything else.

A

(I've requested review from the currently listed devguide experts: @merwok @ezio-melotti @willingc @Mariatta @hugovk)

PR #1766, opened 2026-03-13
Remove modules with no expert

hugovk:
As suggested in https://github.com/python/devguide/pull/1635#discussion_r2836065763.

### DIFF
--- a/core-team/experts.rst
+++ b/core-team/experts.rst
@@ -205,7 +205,7 @@ Miscellaneous
 =============
 
 ==================  ==========================================================
-Interest Area       Maintainers
+Interest area       Maintainers
 ==================  ==========================================================
 algorithms          rhettinger*, serhiy-storchaka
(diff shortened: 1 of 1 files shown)


core-team/experts.rst
```diff
@@ -20,9 +20,9 @@ following is added to an issue or pull request, they will be notified
 automatically.  The :cpy-file:`.github/CODEOWNERS` file is also used to indicate
 maintainers that will be automatically added as reviewers to pull requests.
 
+Names followed by a '*' may be assigned issues involving the module or topic.
 Unless a name is followed by a '*', you should never assign an issue to
-that person.  Names followed by a '*' may be assigned issues involving the
-module or topic.
+that person.
```
hugovk, line 25: I flipped this to put the positive first, so we don't lead with double negative of "unless"/"never" which is less clear. Can also move to separate PR if preferred.

willingc (approved):
Thanks @hugovk. Much cleaner.

PR #1767, opened 2026-03-13
Add PSRT emeritus table

StanFromIreland:
We recently moved retired experts to a separate table ([f2d8b2e64e40e](https://github.com/python/devguide/commit/f2d8b2e64e40e4b38a0a3ed867c4fc2c387cf443)), this PR proposes we do the same for PSRT members. (Seth will take care of populating the table.)

PR #1768, opened 2026-03-18
Fill PSRT emeritus list

sethmlarson:
Follow-up from https://github.com/python/devguide/pull/1767 by @StanFromIreland. Many accounts didn't make the GitHub migration, so don't have a GitHub username that I was able to verify.


StanFromIreland (approved):

hugovk (approved):
Thanks!

PR #1774, opened 2026-03-31
Add Stan to experts list

StanFromIreland:


hugovk (approved):
Thanks!

i18n is making software ready for translation. If you like, feel free to add a translation or l10n interest area.

StanFromIreland:
I considered that, but we already have an l10n (technically also T9N) section: https://devguide.python.org/core-team/experts/#documentation-translations

StanFromIreland:
Thanks!

PR #1775, opened 2026-03-31
Remove platforms/areas with no experts

StanFromIreland:
This expands https://github.com/python/devguide/pull/1766 to platforms and miscellaneous interest areas, noticed when writing the other PR.


willingc (approved):
Thanks @StanFromIreland.

PR #1795, opened 2026-04-22
Minor improvements to the Emscripten instructions

hoodmane:


getting-started/setup-building.rst
```diff
@@ -522,39 +522,47 @@ The simplest way to build Emscripten is to run:
...
+  exit with an error if it is not; run ``install-emscripten`` to populate the
+  cache.
+* ``CROSS_BUILD_DIR`` (or the ``--cross-build-dir`` flag) defines the location
+  of the ``cross-build`` directory that will be used for builds. This can be
+  useful if you need to maintain builds of multiple versions of Python
+  side by side.
 
 It is possible (but not necessary) to enable ``ccache`` for Emscripten builds
-by setting the ``EM_COMPILER_WRAPPER`` environment, but this step will only
-take effect if it is done **after** ``emsdk_env.sh`` is sourced (otherwise, the
```
hoodmane, line 556: This would be difficult to do now since our script automatically sources `emsdk_env.sh`. Thankfully it is not true that it has to be set after sourcing `emsdk_env.sh`. Ccache must be at least version 4.3

getting-started/setup-building.rst
```diff
@@ -522,39 +522,49 @@ The simplest way to build Emscripten is to run:
...
+3. Download the code for all the binary dependencies of Python (such as
+   ``libffi`` and ``mpdecimal``), and compile them for Emscripten; and
+4. Build a copy of Python that can run on Emscripten (the "host" python).
+
+The built binary dependencies are cached inside the Emscripten cache directory.
+Once built for a given Emscripten version, they will not be rebuilt on
+subsequent runs unless there is a change in the version or build script for the
+dependency.
+
+It is assumed that nvm_ is installed in `${HOME}/.nvm`.
```
hoodmane: Ideally, we should search for nvm first in `NVM_DIR` then in `XDG_CONFIG_HOME` and then in `HOME` since that's where nvm likes to install itself. But we don't currently do that.

hoodmane: Also, we should add a `HOST_RUNNER` environment variable so people can set `export HOST_RUNNER=node` if they don't want nvm to be used. Currently you can do that by passing `--host-runner=node` so I guess I can document that much.

willingc (approved):
Nice updates @hoodmane.

PR #1796, opened 2026-04-28
PSRT: Add link to report security issues

hugovk:
https://devguide.python.org/developer-workflow/psrt/ shows up first (for me) in Google for "PSRT".

This page is mostly process info for us.

Let's put a clear link to the page with info about how to report issues.

### DIFF
--- a/developer-workflow/psrt.rst
+++ b/developer-workflow/psrt.rst
@@ -4,6 +4,8 @@ Python Security Response Team (PSRT)
 The Python Security Response Team (PSRT) is responsible for handling
 vulnerability reports for CPython and pip.
 
+To report security issues: https://www.python.org/dev/security/
+
 Members
 -------
 


developer-workflow/psrt.rst
```diff
@@ -4,6 +4,8 @@ Python Security Response Team (PSRT)
 The Python Security Response Team (PSRT) is responsible for handling
 vulnerability reports for CPython and pip.
 
+To report security issues: https://www.python.org/dev/security/
```
StanFromIreland, line 7: WDYT about making it a nice [`sphinx-design` button](https://sphinx-design.readthedocs.io/en/furo-theme/badges_buttons.html) with text like "Report a security issue"?

hugovk: I don't think we need an extra Sphinx extension just for this.

sethmlarson (approved):
Thanks Hugo, this is a good idea.

PR #1797, opened 2026-04-30
Fix subpixel gap between release cycle mask rects

qb20nh:
Fix subpixel gap by stretching white part by 1 pixel into black area, and moving gradient mask to top in z order

Before:
www python org_downloads_ (2)

After:
www python org_downloads_ (1) (1)


qb20nh:
Will the image at https://www.python.org/downloads/ be updated by this too?

PR #1809, opened 2026-05-23
Minor grammar fixes

nedbat:

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -508,7 +508,7 @@ used in ``python.sh``:
 Emscripten
 ----------
 
-Emscripten_ is a complete open source compiler toolchain. It compiles C/C++ code
+Emscripten_ is a complete open-source compiler toolchain. It compiles C/C++ code
 into WebAssembly_/JavaScript executables, for use in JavaScript runtimes,
 including browsers and Node.js.
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -83,7 +83,7 @@ You will only need to execute these steps once per machine:
 
    (You can use both SSH-based or HTTPS-based URLs.)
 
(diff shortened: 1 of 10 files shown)


testing/oss-fuzz.rst
```diff
@@ -2,7 +2,7 @@ OSS-Fuzz for CPython
 ====================
 
 CPython uses `OSS-Fuzz <https://google.github.io/oss-fuzz/>`__, Google's
-continuous fuzzing service for open-source projects, to find bugs and
+continuous fuzzing service for open source projects, to find bugs and
```
hugovk: I normally write "open source" as a noun without a hyphen, but with an active when it acts as an adjective before a noun: "open-source project".

* https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source
* https://en.wikipedia.org/wiki/Open_source
* https://en.wikipedia.org/wiki/Open-source_software

getting-started/setup-building.rst
```diff
@@ -508,7 +508,7 @@ used in ``python.sh``:
 Emscripten
 ----------
 
-Emscripten_ is a complete open-source compiler toolchain. It compiles C/C++ code
+Emscripten_ is a complete open source compiler toolchain. It compiles C/C++ code
```
hugovk: Same here.

testing/oss-fuzz.rst
nedbat: yes, i was looking at these instances. If we go in that direction, then there are a bunch of places to add the hyphen.

nedbat: Interesting: https://en.wikipedia.org/wiki/Open_source seems to get it wrong in a few places.

nedbat:
Better?

willingc (approved):
Thanks @nedbat

PR #1810, opened 2026-05-24
PSRT: Add template for rejecting duplicate reports

StanFromIreland:
This would have saved me some time this morning, and it's becoming an exceedingly common thing to have to deal with.

CC @python/psrt


willingc (approved):
Thanks @StanFromIreland. A nice addition.

sethmlarson (approved):

PR #1811, opened 2026-05-25
CI: declare workflow-level `permissions: {}`

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at workflow level. No GitHub API writes from the workflow.

Post-CVE-2025-30066 hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on: [pull_request, push, workflow_dispatch]
 env:
   FORCE_COLOR: 1
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Check build, markup, and links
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,9 @@ on: [pull_request, push, workflow_dispatch]
 env:
   FORCE_COLOR: 1
 
+permissions:
+  contents: read
```
hugovk: Or even less:
```suggestion
permissions: {}
```

.github/workflows/lint.yml
```diff
@@ -2,6 +2,9 @@ name: Lint
 
 on: [push, pull_request, workflow_dispatch]
 
+permissions:
+  contents: read
```
hugovk:
```suggestion
permissions: {}
```

arpitjain099:
Done @hugovk - both files now use `permissions: {}` per your suggestion. Pushed.

hugovk:
Thanks!

PR #1812, opened 2026-05-26
Add zizmor to pre-commit and fix findings

hugovk:
Follow on from https://github.com/python/devguide/pull/1811.

Add zizmor to pre-commit and fix findings.

Update other pre-commit hooks.

Add handy env vars: 
  `FORCE_COLOR: 1`
  `RUFF_OUTPUT_FORMAT: github`

Replace pre-commit with prek to remove Node.js 20 deprecation warning and unpinned transitive dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,11 @@ updates:
     actions:
       patterns:
         - "*"
+  cooldown:
+    # https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
+    # Cooldowns protect against supply chain attacks by avoiding the
+    # highest-risk window immediately after new releases.
+    default-days: 14
 
 - package-ecosystem: pip
   directory: "/"
@@ -21,3 +26,5 @@ updates:
     pip:
       patterns:
         - "*"
+  cooldown:
+    default-days: 14
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,14 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-python@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -10,6 +10,11 @@ updates:
     actions:
       patterns:
         - "*"
+  cooldown:
+    # https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
+    # Cooldowns protect against supply chain attacks by avoiding the
+    # highest-risk window immediately after new releases.
```
StanFromIreland, line 16: I don't like the [false sense of security](https://mastodon.social/@stanfromireland/116580665468633547) this gives us, this comment makes this seem fail-safe.

hugovk: It's a copy-paste from https://github.com/python/cpython/pull/141866 which references https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns.

Nothing is fail-safe in security, there are many things we can do to protect against these attacks, and this is a good one. They layer on top of each other.

StanFromIreland: Indeed, I find the comment additionally misleading. I suggest either:

- Removing it.
- Adding a link to the Dependabot issue highlighting the flaw in their cooldown system.

hugovk: Let's ask @gpshead who wrote the comment and @woodruffw who wrote the blog post.

gpshead: Dependency cooldowns are not a false sense of security. most supply chain attacks are found and fixed and yanked from the upstream package distribution ecosystems within a few days.  I don't think we should block this PR on wanting the comment better, but we should apply any edit making it better to all places we've added this.

dependabot defaults to not having a cooldown - https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown- - this is simply how we enable the good thing.  _(it'd be great if they changed their default in light of the modern world, but that's not our problem to solve)_

gpshead (approved):

.github/dependabot.yml
StanFromIreland: I’d like to clarify that I’m not against cooldowns (in fact, I support the idea). However, Dependabots implementation is poor, and any threat actor can easily circumvent it (https://github.com/dependabot/dependabot-core/issues/13078).

ezio-melotti (approved):

PR #1818, opened 2026-05-31
Fix Discord link in Tamil translation section

terminaljoint:
Updated the Discord link for Tamil translation.


StanFromIreland:
Can you please confirm this link has the expiry set to never?

terminaljoint:
> Can you please confirm this link has the expiry set to never?

Yeah, it's expiry is set to never, last time we have mistakenly enabled temporary membership, in this new link, we disabled it., and has no limited number for users to join and be in, so this updated link is a safe bet...

StanFromIreland (approved):

PR #1819, opened 2026-06-02
Revise the "Backporting merged changes" guide

StanFromIreland:
I find the current instructions a little brief, so I propose expanding them a little more like so. Generally focusing on `cherry-picker`, I don't think we also need to document the complete instructions for doing it with `git` here. If this is merged, I plan to add a link to the section to the template failed-backport response, e.g.:

image


getting-started/git-boot-camp.rst
```diff
@@ -557,60 +557,80 @@ dismissing your previous review that requested changes.
...
 does **NOT** stop it.
 
+
 Backporting merged changes
 --------------------------
 
-A pull request may need to be backported into one of the maintenance branches
-after it has been accepted and merged into ``main``.  It is usually indicated
-by the label ``needs backport to X.Y`` on the pull request itself.
+After a pull request has been merged into ``main``, it may need to be backported
```
JelleZijlstra, line 566: Do we have a section anywhere about deciding when to backport? If so, it would be good to link it here.

getting-started/git-boot-camp.rst
```diff
@@ -557,60 +557,80 @@ dismissing your previous review that requested changes.
...
+to one or more maintenance branches. This is indicated by the
+:samp:`needs backport to {X.Y}` labels on the pull request.
 
-Use the utility script
-`cherry_picker.py <https://github.com/python/cherry-picker>`__
-to backport the commit.
+``miss-islington`` will automatically attempt to create backport PRs for the
```
JelleZijlstra, line 571:
```suggestion
The Miss Islington bot will automatically attempt to create backport PRs for the
```

StanFromIreland: I've opted to keep this as-is for consistency with the other instances throughout the devguide.

StanFromIreland:
Thanks for the review @JelleZijlstra (can I get another one ;-), I also noticed I had mistaken a duplicate section as this section. I've now merged the two sections, I moved it to the more general page as I don't think we need to limit backporting to commiters.

JelleZijlstra (approved):

encukou:
I think there's some merit to having two sections (though not with the existing text): one to explain what needs to be done, and another explaining how to use the automation to do that. If the tool doesn't work (or you want to improve the tool, or you just “desire a different approach” to quote the bootcamp page's intro), it's good to have a description of the lower-level process.

StanFromIreland:
If anyone's interested, I've written a patch to add the link to Miss Islington's comments: https://github.com/python/miss-islington/pull/738

StanFromIreland:
> I think there's some merit to having two sections (though not with the existing text): one to explain what needs to be done, and another explaining how to use the automation to do that. If the tool doesn't work (or you want to improve the tool, or you just “desire a different approach” to quote the bootcamp page's intro), it's good to have a description of the lower-level process.

I do agree, but currently neither of the sections provide that. I would prefer to do this first (tidy up the existing content), as it will helpful to the contributors who do use the tool. We can add another section in a follow up (I would suggest you open an issue)?

encukou:
OK, let's do that! This PR is an improvement :)

faisal202 (approved):

PR #1825, opened 2026-06-04
Add note on unexpected exceptions to security policy

StanFromIreland:
CC @python/psrt 

This is something we've gotten quite a few times, e.g. [GHSA-2frx-2h99-jv56](https://github.com/python/cpython/security/advisories/GHSA-2frx-2h99-jv56) or [GHSA-32pj-fh79-3c9p](https://github.com/python/cpython/security/advisories/GHSA-32pj-fh79-3c9p). It's only a security issue if the unexpected exception is attacker-triggerable and causes actual availability harm (not merely an unhandled Python exception).

### DIFF
--- a/security/policy.rst
+++ b/security/policy.rst
@@ -46,7 +46,8 @@ Availability vulnerabilities must also demonstrate an "upward" change in posture
 for the attacker, rather than a "lateral" one.
 Unexpected Python exceptions are not vulnerabilities by themselves unless they
 satisfy the availability criteria above.
-This is to avoid handling performance improvements as security vulnerabilities.
+This is to avoid handling performance and correctness improvements as security
+vulnerabilities.
 
 Vulnerabilities in dependencies of Python (such as zlib, Tcl/Tk, or OpenSSL)
 are not vulnerabilities in Python unless Python's use of the dependency
--- a/security/policy.rst
+++ b/security/policy.rst
@@ -44,10 +44,9 @@ dead-locks, and resource exhaustion) must be
 triggerable with data inputs that are reasonably sized for the use case.
 Availability vulnerabilities must also demonstrate an "upward" change in posture
 for the attacker, rather than a "lateral" one.
-Unexpected Python exceptions are not vulnerabilities by themselves unless they
-satisfy the availability criteria above.
-This is to avoid handling performance and correctness improvements as security
-vulnerabilities.
+This is to avoid handling performance improvements as security vulnerabilities.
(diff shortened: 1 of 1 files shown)


security/policy.rst
```diff
@@ -44,7 +44,10 @@ dead-locks, and resource exhaustion) must be
 triggerable with data inputs that are reasonably sized for the use case.
 Availability vulnerabilities must also demonstrate an "upward" change in posture
 for the attacker, rather than a "lateral" one.
-This is to avoid handling performance improvements as security vulnerabilities.
+Unexpected Python exceptions are not vulnerabilities by themselves unless they
```
sethmlarson: I think this is a great point, exceptions could easily be framed as an escalation of "privilege" (executing program -> crashed program). For this reason I don't think "unless they satisfy the criteria above" is enough, we need to be explicit that exceptions are normal behavior of the program and need to be handled by user code when processing arbitrary input. How's this?

> Exceptions are an expected part of control flow when processing arbitrary inputs, therefore crashes resulting from unhandled exceptions are not security vulnerabilities.

I'd also suggest having this sentence last in this paragraph and reverting to "avoid handling performance improvements as security vulnerabilities". Exceptions usually are about crashes impacting availability instead of performance.

sethmlarson (approved):

StanFromIreland:
Thanks Seth!

security/policy.rst
```diff
@@ -44,10 +44,9 @@ dead-locks, and resource exhaustion) must be
 triggerable with data inputs that are reasonably sized for the use case.
 Availability vulnerabilities must also demonstrate an "upward" change in posture
 for the attacker, rather than a "lateral" one.
-Unexpected Python exceptions are not vulnerabilities by themselves unless they
-satisfy the availability criteria above.
-This is to avoid handling performance and correctness improvements as security
-vulnerabilities.
+This is to avoid handling performance improvements as security vulnerabilities.
+Exceptions are an expected part of control flow when processing inputs,
+therefore crashes resulting from unhandled exceptions are not security vulnerabilities.
```
encukou, line 49: We usually use the term “crash” for segfaults rather than exceptions (see `Lib/test/crashers` for example). Should this be just:

```suggestion
therefore unhandled exceptions are not security vulnerabilities.
```

encukou: PR: https://github.com/python/devguide/pull/1849

StanFromIreland: Oh right, sorry I had this saved but plan to give it to someone looking to contribute.

PR #1826, opened 2026-06-05
Clarify that performance improvements are not backported

gaborbernat:
A contributor recently proposed backporting a performance improvement to the maintenance branches, and the question surfaced that the devguide does not say whether optimizations should be backported. Core developers confirmed the long-standing practice in the linked discussion: optimizations are not backported, because an optimization can introduce a regression and a regression in a bugfix release is costly to track down and undo.

This adds that rule to the maintenance-branch section of the development cycle, next to the existing "bug fixes, test improvements, and documentation" guidance, and spells out the narrow exception a few core developers noted: a change that is a clear win, carries no regression risk (including for code that subclasses or monkey-patches internals), and matters enough to applications to justify it, discussed first.

Discussion: https://discuss.python.org/t/performance-improvements-and-backports/107625

### DIFF
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -75,18 +75,13 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes, test improvements, and edits to the documentation.
+bug fixes, test improvements, and edits to the documentation. Performance
+improvements are not among them: an optimization can introduce a regression,
+and a regression in a maintenance release is painful to track down and undo.
 Also, a general rule for maintenance branches is that compatibility
 must not be broken at any point between sibling micro releases (3.12.1, 3.12.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
-
-Performance improvements are not backported. An optimization can introduce a
-regression, and a regression shipped in a maintenance release is painful to
-track down and undo, so optimizations remain on the ``main`` branch. A backport
-is a rare exception, reserved for a clear win that carries no regression risk
-(including for code that subclasses or monkey-patches internals) and matters
-enough to applications to justify it, and it **must** be discussed first.
 
 Backporting changes reduces the risk of future conflicts.
 For documentation, it increases the visibility of improvements,
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -81,6 +81,13 @@ must not be broken at any point between sibling micro releases (3.12.1, 3.12.2,
 etc.).  For both rules, only rare exceptions are accepted and **must** be
 discussed first.
 
+Performance improvements are not backported. An optimization can introduce a
+regression, and a regression shipped in a maintenance release is painful to
+track down and undo, so optimizations remain on the ``main`` branch. A backport
+is a rare exception, reserved for a clear win that carries no regression risk
+(including for code that subclasses or monkey-patches internals) and matters
+enough to applications to justify it, and it **must** be discussed first.
+
 Backporting changes reduces the risk of future conflicts.
 For documentation, it increases the visibility of improvements,
 since most readers access the `stable documentation <https://docs.python.org/3/>`__
--- a/developer-workflow/development-cycle.rst
+++ b/developer-workflow/development-cycle.rst
@@ -74,14 +74,15 @@ produced from a maintenance branch are called **bugfix** or **maintenance**
 releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
-The only changes allowed to occur in a maintenance branch without debate are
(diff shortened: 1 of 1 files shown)


developer-workflow/development-cycle.rst
skirpichev: IMO, this sentence is very clear.  Why we need to add a new paragraph?

gaborbernat: I think worth calling out performance improvements, as the discussion topic clearly proves that wasn't clear enough for everyone, though perhaps we can put it into this sentence so it's not a new one.

developer-workflow/development-cycle.rst
```diff
@@ -75,7 +75,9 @@ releases; the terms are used interchangeably. These releases have a
 **micro version** number greater than zero.
 
 The only changes allowed to occur in a maintenance branch without debate are
-bug fixes, test improvements, and edits to the documentation.
+bug fixes, test improvements, and edits to the documentation. Performance
```
StanFromIreland: I don't really get why "performance improvements" are special here, they are features, and we already explicitly state (albeit [elsewhere](https://devguide.python.org/core-team/committing/#assessing-a-pull-request)):

> The only branch that receives new features is `main`, the in-development branch.

In this event, I think everything generally went according to the docs: can't backport features "without debate," however exceptions can be made, provided you open a discussion first.

gaborbernat: If we all agree I can withraw this PR, but I think there's benefit of having this here, as would have saved that discussion thread and initial confusion by someone else.

JelleZijlstra: I don't think it's obvious that performance improvements are "features", so it's worth clarifying.

vstinner (approved):
LGTM

developer-workflow/development-cycle.rst
StanFromIreland: This does not seem to be the place for explaining what features are.

gaborbernat: Perhaps, we should cross link the "feature" explanation location then here? So people can click through and see how we define a feature.

gaborbernat: > I don't really get why "performance improvements" are special here, they are features, and we already explicitly state (albeit [elsewhere](https://devguide.python.org/core-team/committing/#assessing-a-pull-request)):

@StanFromIreland by the way I believe you wanted to link to https://devguide.python.org/developer-workflow/development-cycle/#in-development-main-branch not the committing part, right?

StanFromIreland: > Perhaps, we should cross link the "feature" explanation location then here? So people can click through and see how we define a feature.

Yes, I think we should clarify what features are and link here, overall I think that will be more helpful.

StanFromIreland: > @StanFromIreland by the way I believe you wanted to link to https://devguide.python.org/developer-workflow/development-cycle/#in-development-main-branch not the committing part, right?

I quoted the section in committing.

gaborbernat: Cross-linking is awkward because that section lists everything main takes, bug fixes included, and bug fixes are the one type that backports without debate. Pointing there to explain what needs discussion sends people to a list where one of the four entries is the opposite case.

I pushed a different fix: the maintenance section now splits changes into low-risk (bug fixes, tests, docs, backport freely) and higher-risk (new features, semantic changes, performance improvements, need a discussion and a strong case first). It names performance improvements directly, so the original confusion can't recur, and it never has to define what a feature is.

StanFromIreland: I think this is going the wrong way, and complicating this paragraph. It's goal is to explain what we do and don't backport, not list all the kinds of features. IMO, the note in this paragraph should be something like:

> New features should not be backported.

We note features in various places, they shouldn't all have to explain what exactly they are.

gaborbernat: I might be missing something but where do we define that performance improvement is a feature? I know you referenced earlier,  but pointed to the opening PR section. The [devguide.python.org/developer-workflow/development-cycle#in-development-main-branch](https://devguide.python.org/developer-workflow/development-cycle/#in-development-main-branch) specifically calls it out as a separate thing, not as a feature.

> We note features in various places, they shouldn't all have to explain what exactly they are.

Explicit better than implicit, I think is better to repeat things (or at least cross link things) rather than let people assume from their knowledge from other locations.

StanFromIreland: > cross link things) rather than let people assume from their knowledge from other locations.

Yes, that's my idea from above:

> I think we should clarify what features are and link here

gpshead (approved):

gpshead (approved):
I like this wording. far less controversial to some than my own statement about performance backports over on discuss.  At the end of the day the release managers are part of the discussion required before backporting non-bugfix or non-low-risk.

gpshead:
_(merged as it had a couple of approvals, further improvements can be iterated upon)_

PR #1827, opened 2026-06-05
Simplify the landing for new contributors

StanFromIreland:
We essentially had two introductory sections, with the first paragraph linking to one directly below, which also reiterated the first introduction. So, I propose removing the intermediate "Contributing" title and re-flowing the introduction.
Additionally, I also tried to make it more encouraging. The introduction may feel a little overwhelming for new contributors, since we immediately present a large table with lots of links. As such, I think the video helps set a more welcoming tone before readers move on to the great big devguide :-)

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -1,4 +1,5 @@
 .. _devguide-main:
+.. _contributing:
 
 ========================
 Python Developer's Guide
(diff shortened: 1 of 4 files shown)


nedbat (approved):

index.rst
```diff
@@ -63,15 +56,23 @@ Guide for contributing to Python:
...
+so you may need to backtrack for missing concepts or terminology.
 
-A number of individuals from the Python community have contributed to a series
+For broader open source contribution advice, a number of individuals from the
+Python community have contributed to a series
 of excellent guides at `Open Source Guides <https://opensource.guide/>`__.
 For example,
 `How to Contribute to Open Source <https://opensource.guide/how-to-contribute/>`__.
 
+.. admonition:: You don't have to be a compiler engineer to work on Python -- Savannah Bailey
```
hugovk: I know YouTube says Bailey, but @savannahostrowski would you prefer Ostrowski? 
```suggestion
.. admonition:: You don't have to be a compiler engineer to work on Python -- Savannah Ostrowski
```

index.rst
```diff
@@ -63,15 +56,23 @@ Guide for contributing to Python:
...
 can stop where you feel comfortable and begin contributing immediately without
 reading and understanding everything.  If you do choose to skip
-around within the guide, be aware that sections build on each other,
-so you may find it necessary to backtrack to fill in
-missing concepts and terminology.
+around within the guide, be aware that some sections build on each other,
+so you may need to backtrack for missing concepts or terminology.
 
-A number of individuals from the Python community have contributed to a series
+For broader open source contribution advice, a number of individuals from the
```
hugovk:
```suggestion
For broader open-source contribution advice, a number of individuals from the
```

index.rst
savannahostrowski: Let's go with Ostrowski. 

(I haven't gotten around to changing because I don't want to deal with the government right now 😅).

savannahostrowski (approved):

PR #1828, opened 2026-06-07
Introduce "type names" section to the style guide

### LINKED ISSUE #1821: Feature: Expand the style-guide to cover how type names should be written in prose

In order to smooth out stylistic debates about how docs are spelled, the style-guide page should provide guidance about how type-names should be written.

There are various cases in the stdlib in which we refer to a type vs a more general idea (for example: `tuple` vs "tuples").
There are also cases in which documentation wants to refer to a type as an established idea in the relevant domain rather than the type (for example: "thread-local storage" vs `threading.local`).
Finally, there are cases where the name of the idea is related to the type name, as in "deque" for `collections.deque` or "Context Variable" for `contextvars.ContextVar`.

The style-guide should provide rules for which style to use when writing a type name in prose, so that we avoid and can heal pages like the current `contextvars` docs, which use `ContextVar`, "Context Variable", and "context variable".

I propose the following rule as an addition to the style guide below the [specific words section](https://devguide.python.org/documentation/style-guide/#specific-words):

> Type names
> ----------
>
> When writing the names of types in prose, prefer to write the name of the type exactly as it appears in source, styled as a class reference or an unlinked class. For example, refer to ``dict`` as ``:class:`dict`‌`` or ``:class:`!dict`‌``.
>
> Links should be used according to the `guidance on links <links>`.
>
> Some type names are commonly understood ideas or nouns outside of Python. For example, "tuples" are a general programming concept, as distinct from the ``tuple`` type.  When referring to general ideas, do not style the relevant word as a type.
>
> Many types have descriptive names which do not exactly match their type name.  For example, "context variables" describes ``contextvars.ContextVar``, and "partial function" may be used to describe an application of ``functools.partial``. Use these names only when they serve to clarify the text better than the type name itself would, and put them in lowercase.

Make CPython doc issues and edits without establishing style guide guidance, trying to make things consistent organically over time.
(I think this is unlikely to succeed, but it is possible.)

### Additional context

In [a recent CPython docs PR](https://github.com/python/cpython/pull/150325), we found that we could not write text that was consistent across different modules because those modules had different styles for referring to relevant classes by name.

Some docs use class names verbatim, as they appear in source, and code-quoted. For example, the new `frozendict` type is generally written exactly as it is named, styled as a class or link/ref: "`frozendict`".

Some docs use class names as though they are ordinary nouns. For example, the [`deque` docs](https://docs.python.org/3/library/collections.html#collections.deque) refer to "deques", without code quoting or linking.

A small category of docs refer to the class name as though it is a proper noun. At time of writing, I can only find `Context Variables` as an example of this.

sirosen:
Resolves #1821

This enhancement to the style guide was motivated by a desire to resolve some inconsistent phrasing in the docs, which caused some disagreement in a CPython docs PR.

The new section describes guidance for how type names should be documented. Class references or unlinked class styling are preferred, and other styles are described for use only when they are superior to the class name.

---

The issue has no comments, but I introduced the idea in the docs community where it got some positive feedback.
The discussion there helped to shape the issue, which has some basic :+1: s from a couple of core devs.

### DIFF
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -58,6 +58,35 @@ of these entities are not assigned any special markup, but the preferred
 spellings are given in :ref:`specific words` to aid authors in maintaining the
 consistency of presentation in the Python documentation.
 
+
+Use simple language
+===================
+
+Avoid esoteric phrasing where possible.  Our audience is world-wide and may not
+be native English speakers.
+
+Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
+such as "for example" or "that is."
+
+
+Charged terminology to avoid
+============================
+
+Avoid terminology that may be considered insensitive or exclusionary.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Avoid
+     - Instead
+   * - whitelist
+     - allowlist
+   * - blacklist
+     - blocklist, denylist
+   * - master/slave
+     - main, parent/child, server/client, primary/secondary
+
+
 .. _specific words:
 
 Specific words
@@ -135,34 +164,6 @@ name.  For example, "context variables" describes ``contextvars.ContextVar``,
 and "partial function" may be used to describe an application of
 ``functools.partial``.  Use these names only when they serve to clarify the text
 better than the type name itself would, and put them in lowercase.
-
-
-Use simple language
-===================
-
-Avoid esoteric phrasing where possible.  Our audience is world-wide and may not
-be native English speakers.
-
-Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
-such as "for example" or "that is."
-
-
-Charged terminology to avoid
-============================
-
-Avoid terminology that may be considered insensitive or exclusionary.
-
-.. list-table::
-   :header-rows: 1
-
-   * - Avoid
-     - Instead
-   * - whitelist
-     - allowlist
-   * - blacklist
-     - blocklist, denylist
-   * - master/slave
-     - main, parent/child, server/client, primary/secondary
 
 
 .. index:: diataxis
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -119,7 +119,7 @@ Unix
 Type names
 ==========
 
-When writing the names of types in prose, prefer to write the name of the type
+When writing the names of types in prose, write the name of the type
 exactly as it appears in source, styled as a class reference or an unlinked
 class.  For example, refer to dict as ``:class:`dict`‌`` or ``:class:`!dict`‌``.
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -116,6 +116,27 @@ Unix
    1970s.
 
 
+Type names
+==========
+
+When writing the names of types in prose, prefer to write the name of the type
+exactly as it appears in source, styled as a class reference or an unlinked
+class.  For example, refer to dict as ``:class:`dict`‌`` or ``:class:`!dict`‌``.
+
+Links should be used according to the :ref:`guidance on links <links>`.
+
+Some type names are commonly understood ideas or nouns outside of Python.
+For example, "tuples" are a general programming concept, as distinct from the
+``tuple`` type.  When referring to general ideas, do not style the relevant word
+as a type.
+
+Many types have descriptive names which do not exactly match their type
+name.  For example, "context variables" describes ``contextvars.ContextVar``,
+and "partial function" may be used to describe an application of
+``functools.partial``.  Use these names only when they serve to clarify the text
+better than the type name itself would, and put them in lowercase.
+
+
 Use simple language
 ===================
 
--- a/documentation/style-guide.rst
+++ b/documentation/style-guide.rst
@@ -152,7 +152,7 @@ When writing the names of types in prose, write the name of the type
 exactly as it appears in source, styled as a class reference or an unlinked
 class.  For example, refer to dict as ``:class:`dict`‌`` or ``:class:`!dict`‌``.
 
-Links should be used according to the :ref:`guidance on links <links>`.
+Links should be used according to the :ref:`guidance on links <style-guide-links>`.
 
 Some type names are commonly understood ideas or nouns outside of Python.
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

documentation/style-guide.rst
```diff
@@ -116,6 +116,27 @@ Unix
    1970s.
 
 
+Type names
+==========
+
+When writing the names of types in prose, prefer to write the name of the type
```
```suggestion
When writing the names of types in prose, write the name of the type
```

documentation/style-guide.rst
```diff
@@ -116,6 +116,27 @@ Unix
...
+Some type names are commonly understood ideas or nouns outside of Python.
+For example, "tuples" are a general programming concept, as distinct from the
+``tuple`` type.  When referring to general ideas, do not style the relevant word
+as a type.
+
+Many types have descriptive names which do not exactly match their type
+name.  For example, "context variables" describes ``contextvars.ContextVar``,
+and "partial function" may be used to describe an application of
+``functools.partial``.  Use these names only when they serve to clarify the text
+better than the type name itself would, and put them in lowercase.
```
hugovk: I suggest moving this entry after the simple language one, and perhaps also after the charged terminology one as well.

sirosen: Should the "Specific words" section above it also move? I thought it made sense to put these together, and I still do. What would you say to moving both of these?

I could do it here (pragmatic and easy, but my least favorite way), do a precursor PR to move "Specific words" down, or do a follow-up PR.

hugovk: Moving "Specific words" sounds okay to me, no preference whether in this PR or another.

sirosen: I'll do it here.

I'm used to smaller projects with relatively much more review bandwidth per PR. I'm adjusting a bit for cpython and related projects. :slightly_smiling_face:

sirosen: 3c5e2c7fea8809a89a5470c9fa6b0eefb9a24a5e pushes them down, and as "proof" of my work making no other changes, the git diff selects the text which I did not cut-and-paste.

I did introduce one extra newline, which seemed like it was missing for a section-break.

hugovk (approved):
Thanks!

documentation/style-guide.rst
```diff
@@ -116,32 +145,25 @@ Unix
...
-Avoid esoteric phrasing where possible.  Our audience is world-wide and may not
-be native English speakers.
-
-Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
-such as "for example" or "that is."
+When writing the names of types in prose, write the name of the type
+exactly as it appears in source, styled as a class reference or an unlinked
+class.  For example, refer to dict as ``:class:`dict`‌`` or ``:class:`!dict`‌``.
 
+Links should be used according to the :ref:`guidance on links <links>`.
```
```
/home/runner/work/devguide/devguide/documentation/style-guide.rst:155: WARNING: undefined label: 'links' [ref.ref]
```

sirosen: Shouldn't that link to the section title? I feel certain that this worked for me.

I just added an explicit anchor, on the assumption that I got confused and forgot how sphinx handles section-names.

StanFromIreland: > Shouldn't that link to the section title?

You have a `:ref:`, not a regular link <code>\`guidance on links \<links\>\`_</code> (I don't remember if we still need the `#`) is what you're thinking about

I'm +1 for an explicit reference however, I think it's far more future proof. But can we rename it to something a little more useful, like `links-style-guide`.

sirosen: Ah, so that we can use it from outside of the style-guide! Yes, I like that, but mildly prefer `style-guide-links` (namespace first).

I've added a co-authored commit. :slightly_smiling_face:

documentation/style-guide.rst
```diff
@@ -116,32 +145,25 @@ Unix
...
-     - allowlist
-   * - blacklist
-     - blocklist, denylist
-   * - master/slave
-     - main, parent/child, server/client, primary/secondary
+Many types have descriptive names which do not exactly match their type
+name.  For example, "context variables" describes ``contextvars.ContextVar``,
+and "partial function" may be used to describe an application of
+``functools.partial``.  Use these names only when they serve to clarify the text
+better than the type name itself would, and put them in lowercase.
```
nedbat: In line with my [comment on the issue](https://github.com/python/devguide/issues/1821#issuecomment-4822446152), could we add:

> Python classes can also be discussed not as linked classes, but as concepts or with descriptive names. For example, a long discussion of dictionaries won't use `dict` everywhere. Sometimes "dict" or "dictionary" will feel more natural. The author will need to exercise judgement when to use class links and when to use descriptive names.

sirosen: I agree with this proposed text, but feel that it has a degree of overlap with the last paragraph.

I'm going to try to integrate the two but intend to append this if I can't figure out a good combination.

encukou: The style guide as a whole could use a PEP-8 style note:

> know when to be inconsistent – sometimes style guide recommendations just aren’t applicable. When in doubt, use your best judgment. Look at other examples and decide what looks best. And don’t hesitate to ask!

---

Specifically for class names in long discussion, perhaps this could mirror the guidance on links. The first occurrence should use `:class:`; subsequent ones can use plain words if it's clear that it's still talking about the same thing.

sirosen: I've made an edit to weave these suggestions into the text, which required a little tuning of the first paragraph in addition to a larger change to the last.

The part which is most different from Ned's suggestion is this:
> Once it is clear that the text refers to a specific type, use the naming which suits the context: in the case of ``dict``, any of "dict", "dictionary", or "``:class:`dict```" may be best.

I mean for that to carry the message about "the author will need to exercise judgement" as well as the idea that, like links, you should first clarify the exact type and then exercise more freedom in how you write.

I'm very concerned with brevity for these docs, so I was trying to pack the ideas in very tightly, in the smallest reasonable number of words.

nedbat (approved):

encukou (approved):

StanFromIreland (approved):
LGTM

PR #1832, opened 2026-06-12
Update Łukasz's and Petr's roles

StanFromIreland:


hugovk:
Should we add @encukou to https://github.com/orgs/python/teams/organization-owners or https://github.com/orgs/python/teams/cpython-repository-admins to continue dev-in-res tasks?

encukou:
Yes.
I've officially [asked the SC](https://github.com/python/steering-council/issues/354) to confirm.

PR #1834, opened 2026-06-12
Use `sphinx-doc/github-problem-matcher` in CI

StanFromIreland:
This would save me [doing it manually](https://github.com/python/devguide/pull/1828#discussion_r3404460702).


hugovk:
What does this do? Can you show a demo?

StanFromIreland:
The README has some examples: https://github.com/sphinx-doc/github-problem-matcher#example

PR #1838, opened 2026-06-13
Move the "quick reference" to a dedicated page

### LINKED ISSUE #1837: Feature: move the "quick reference" from the devguide root to a dedicated page under Getting Started

As the root page has evolved, it has become more clearly an index page, with some high level content but not much detailed documentation, leaving that for other pages. Technical guidance at the level of "run this command" no longer belongs at the root of the docs, except perhaps in summary fashion.

I think we should add a new page to the Getting Started docs, but *after* "Setup and building" in doc order. The intention being that this is explicitly *not* a "quickstart guide", and treating it as such should be subtly discouraged.

Initially, I intended to propose we remove the section. The content it covers *is* fully captured by other pages.
However, the "quick reference" is, itself, unusual in that it spans the concerns of cloning, building, testing, running `blurb`, and opening a PR. For a new contributor, who may have done only 1-2 PRs already, having an easy reference doc which fits on one page, especially as reminders for those final steps, is useful.

I also weighed the possibility of subtly changing the purpose of the doc, from "Quick Reference" to "Quickstart". There are two reasons I rejected the idea. First, the target audience for this doc is not, in my opinion, brand new contributors. It's for newer or infrequent contributors who benefit from having an easy reference to things they've already learned. The second reason is that the typical implication of a "quickstart" guide is that "you follow these steps and you'll be ready", which seems like the wrong message to send to new contributors.

### Additional context

This was inspired out of a discord conversation ([ref](https://discord.com/channels/935215565872693329/935219976481161216/1515465671733940338)). @nedbat noted that [the "quick reference" section](https://devguide.python.org/#quick-reference) in the index doc is not well-placed. I agree, and think there are a couple of options for what to do.

At first, I thought it was _just_ a git reference doc, and intended to propose removal. Looking closely, I think there is a useful thing here worth preserving. I had never -- as a new contributor -- looked at this bit of doc, and there are subtle details in it which are good reminders! e.g.,
```bash
./python -m test -j3
```
reminds me that I can pass `-j <N>` to the testsuite.

sirosen:
A new page at `Getting Started > Quick Reference` is added to contain the
(moved) content which was in the Quick Reference section of the index
document for the site.

The content is left almost entirely untouched, but the introductory text
is slightly altered to de-emphasize use of the doc for new contributors.

Intentionally, as described in #1837, the page is placed in Getting
Started but after the setup guide.

---

closes #1837


getting-started/quick-reference.rst
```diff
@@ -0,0 +1,97 @@
+.. _quick-reference:
+
+===============
+Quick Reference
+===============
+
+Here are the basic steps needed to get set up and open a pull request.
+
+This is meant as a checklist and cheat-sheet, not a comprehensive guide.
+For complete instructions please see the :ref:`setup guide <setup>`.
```
StanFromIreland: I don't think this makes sense now, we redirect people *back* to the page they've just read?

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,97 @@
...
+
+         ./configure --config-cache --with-pydebug && make -j8
+
+   .. tab:: Windows
+
+      .. code-block:: dosbatch
+
+         PCbuild\build.bat -e -d
+
+   See also :ref:`more detailed instructions <compiling>`,
```
StanFromIreland: All of the above steps should have been completed by the time they reach this page, so we don't quite need them.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,97 @@
...
+         :file:`./python.exe` in order to avoid filename conflicts with
+         the ``Python`` directory.
+
+   .. tab:: Windows
+
+      .. code-block:: dosbatch
+
+         .\python.bat -m test -j3
+
+5. Create a new branch where your work for the issue will go, for example::
```
StanFromIreland: This section is duplicated by our other ["Quick guide"](https://devguide.python.org/getting-started/pull-request-lifecycle/#quick-guide).

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,97 @@
+.. _quick-reference:
+
+===============
+Quick Reference
```
hugovk: Use sentence case for headers:

https://devguide.python.org/documentation/style-guide/#capitalization


```suggestion
Quick reference
```

getting-started/quick-reference.rst
nedbat: This doesn't go back to the home page. It goes to a detailed page about git etc.

StanFromIreland: I was not referring to the home page. As @sirosen noted in the PR description:

> the page is placed in Getting Started but after the setup guide.

We're sending them back to the page just before.

nedbat: I see, sorry. So we should move this up in the contents.

StanFromIreland: I think we can rename it to "Overview," and make it an index for the overall quite long "Getting started" section. Additionally, we could merge it with the other [Quick Guide](https://devguide.python.org/getting-started/pull-request-lifecycle/#quick-guide), and add a link to the [Where to get help](https://devguide.python.org/getting-started/getting-help/) at the end. Some of the steps could also use additional links to the more detailed sections.

getting-started/index.rst
```diff
@@ -8,6 +8,7 @@ Getting started
    :maxdepth: 5
 
    setup-building
+   quick-reference
```
willingc, line 10: I wonder if it would be better to move this before setup-building. I think we might want to rethink the name "Quick reference". The original text was created to be a checklist/cheatsheet for getting started quickly especially at sprints.

I don't object to making a separate doc for the content. I would leave a clear link/callout in the index page at the same place in the text so that folks who have the page bookmarked for this content know where it moved.

getting-started/quick-reference.rst
willingc: May make sense to be more specific on the title of this doc. Quick reference feels too broad. Something that evokes the original purpose checklist/cheatsheet would make more sense.

getting-started/quick-reference.rst
willingc: I would love to see this page be minimalist visually and cheatsheet-like. I think @StanFromIreland is on the right track though "Overview" feels too broad.

getting-started/quick-reference.rst
willingc: I think all the "see also" items could go later in this document. Perhaps in a "Dive deeper" section.

getting-started/quick-reference.rst
willingc: Steps 5-7 could be their own section or their own page. Keep 1-4 as build from source and run tests.

index.rst
```diff
@@ -74,103 +74,6 @@ For example,
          :width: 100%
 
 
-.. _quick-reference:
-
-Quick reference
----------------
-
-Here are the basic steps needed to get set up and open a pull request.
```
willingc, line 82: When moving these items to a new page, let's keep the Quick reference title and have an admonition of where the content has moved and why (for approx. 6 months).

willingc (commented):
@sirosen Thanks for following up on the issue that you filed. I think you have some good points. The devguide has evolved quite a bit over the past decade and simplifying the index page makes good sense.

getting-started/quick-reference.rst
sirosen: We could put it at or near the end of Getting Started, under the name "Workflow cheatsheet".

It would read logically that way, and that name is harmonious with the element of this doc that I'm trying to preserve.

getting-started/quick-reference.rst
sirosen: > Additionally, we could merge it with the other [Quick Guide](https://devguide.python.org/getting-started/pull-request-lifecycle/#quick-guide), and add a link to the [Where to get help](https://devguide.python.org/getting-started/getting-help/) at the end.

Oh no! I didn't realize that there was another, other-other quick reference. 😂

My initial reaction is that I think these should be combined, still in the form of a new page.

I need to do some careful cross-comparison of the content and see if such a change is feasible in practice, without this PR growing too large in scope.

index.rst
StanFromIreland: We can add a little deferred Javascript redirect for the anchor, e.g.:

```js
function redirectAnchor() {
    const anchor = window.location.hash.slice(1);
    if (!anchor || document.getElementById(anchor)) {
        return;
    }

    if (anchor !== "anchor") {
        return;
    }

    window.location.replace(new URL("redirect", window.location.href).href);
}

redirectAnchor();
```

sirosen: I like the idea, but have two reasons I don't want to pursue it at the moment.

1.  Reusability. We probably don't want to do a bespoke bit of JS each time we do a redirect.

    I'd like us to pack this idea up to be reusable. I think the redirector you recently proposed for CPython docs was similar -- but I did see that not everyone agreed on that.

2.  Scope creep. This will already end up being a larger PR than I intended.

    When I circle back on this, I'll be moving more pieces around and actually changing some content to address feedback. I worry that adding in new code for the site may make this harder to merge, so I'd like to stick strictly to content edits.

(I can definitely take feedback if there's broad agreement that everyone really wants us to add `redirectAnchor()` right here! But I'm hesitant to embrace the idea without more signal.)

sirosen:
I've just pushed a few changes that attempt to address all of the outstanding review comments.

1. It seemed to me that I had to choose: (A) double down on this as a "quick start" or "quick reference" and move it to the start of Getting Started (B) adjust phrasing to make this a "cheat sheet" and move it to the end.
I looked at both, thinking I wanted (B), but (A) was much easier to achieve. I kept the name the same, "quick reference", although I somewhat agree with Carol's thought that the name is a bit vague and imperfect -- I just didn't come up with a better one.

2. I also broke it down into sections, rather than keeping it as a list. It's still very short in content, but it looks bulkier with the section headings. I think that with sections the page makes more sense, as it crosses several topic areas.

3. Each section links deeply to the matching place in Getting Started (or testing docs, for the testing item).

4. I eliminated the PR "quick guide" and made sure that the content from there about opening PRs is carried over. A very small amount of content from the start of that section is lost (I noted this in Discord chat; potentially I could file an issue to restore it?).

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+Fork and clone the repo
+-----------------------
+
+Fork `the CPython repository <https://github.com/python/cpython>`__
+to your GitHub account and :ref:`get the source code <checkout>` using::
+
+   git clone https://github.com/<your_username>/cpython
+   cd cpython
+
+We recommend also setting up ``pre-commit``::
```
StanFromIreland: @encukou, are you happy to do this now that we've hash-pinned all the third party hooks?

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+
+   .. code-block:: shell
+
+      ./configure --config-cache --with-pydebug && make -j $(nproc)
+
+.. tab:: macOS
+
+   .. code-block:: shell
+
+      ./configure --config-cache --with-pydebug && make -j8
```
StanFromIreland: Why are we limiting to `8`?

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+`blurb-it <https://blurb-it.herokuapp.com/>`__,
+or the :pypi:`blurb` tool and its ``blurb add`` command.
+
+.. tip::
+
+   You can read more about ``blurb`` in its
+   `repository <https://github.com/python/blurb>`__.
+
+For more information on writing news entries,
+see :ref:`"Updating NEWS and What's New in Python" <news-entry>`.
```
StanFromIreland: This should be higher, we don't want unnecessary news entries (it explains when you should(n't) add one).

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+
+   .. code-block:: shell
+
+      ./python -m test -j3
+
+.. tab:: macOS
+
+   .. code-block:: shell
+
+      ./python.exe -m test -j8
```
StanFromIreland: Similarly here, why the difference? We can pass `-j0` here.

getting-started/quick-reference.rst
hugovk: Some number is better than none, easy to adjust once you know a number goes there, and arbitrary selection based on market hardware: https://github.com/python/devguide/pull/1545

StanFromIreland: If `$(nproc)` can fail then why not leave it up to `make` and pass `-j` (I'm not sure if that works over on macOS, I presume it would be the same as on Linux), it should use all available then.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
+.. _quick-reference:
+
+===============
+Quick reference
+===============
+
+Here are the basic steps needed to get set up and open a pull request.
+
+This is meant as a checklist and cheat-sheet, not a comprehensive guide.
+For complete instructions please see the :ref:`setup guide <setup>` and the
```
hugovk: We don't need "please" in technical docs
```suggestion
For complete instructions see the :ref:`setup guide <setup>` and the
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+Make sure the
+:ref:`continuous integration checks on your Pull Request are green <keeping-ci-green>` (successful).
+
+Read and respond to reviewer comments on your pull request.
+
+See also, GitHub's documentation on `commenting on Pull Requests`_.
+
+.. note::
+
+   In order to keep the commit history intact, please avoid squashing or amending
```
hugovk:
```suggestion
   In order to keep the commit history intact, avoid squashing or amending
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Here are the basic steps needed to get set up and open a pull request.
+
+This is meant as a checklist and cheat-sheet, not a comprehensive guide.
+For complete instructions please see the :ref:`setup guide <setup>` and the
+:ref:`pull request guide <pullrequest>`.
+
+
+Setup Git
+=========
```
hugovk:
```suggestion
Set up Git
==========
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+   You can read more about ``blurb`` in its
+   `repository <https://github.com/python/blurb>`__.
+
+For more information on writing news entries,
+see :ref:`"Updating NEWS and What's New in Python" <news-entry>`.
+
+Create pull requests
+--------------------
+
+Create pull bequests on GitHub from your branches, on your fork, and make sure
```
hugovk:
```suggestion
Create pull requests on GitHub from your branches, on your fork, and make sure
```

getting-started/quick-reference.rst
hugovk: It depends on the `make` command. With GNU `make`:

```
       -j [jobs], --jobs[=jobs]
            Specifies the number of jobs (commands) to run simultaneously.  If there is
            more than one -j option, the last one is effective.  If the -j option is
            given without an argument, make will not limit the number of jobs that can
            run simultaneously.
```

"Don't limit" isn't the same as "use all available".

getting-started/quick-reference.rst
encukou: What's the policy on updating those hash-pins?

(FWIW, you don't need to make me happy. Just Seth ;)

hugovk: None yet.

I'd like to automate with either:

* https://pre-commit.ci to auto-update them weekly, monthly or quarterly (my pick is quarterly). This can also autofix formatting in PRs, often a tedious thing to ask contributors to do (we can also configure to disable that, my preference is to leave enabled).

* Dependabot: we're already using this for bumping GitHub Actions and pip dependencies. Interval can be daily, weekly, monthly, quarterly, semiannually, yearly, or a custom cron. It won't autofix PRs. We can also configure cooldowns.

And we can manually update when we need a new feature or bugfix.

sirosen: I take it that there was some discussion about ensuring these are pinned for safety. 😁 

---

I use pre-commit.ci for this in most of my projects, and the autofixing is quite nice.

_If_ the autofixing isn't desirable, I'd tend towards Dependabot, since the infrastructure is provided by GitHub. (I think this is nicer in having fewer providers and putting less load on a smaller platform.)

getting-started/quick-reference.rst
sirosen: I see that Hugo 👍'ed the comment below about switching to `-j0` for the test command.
I'm unclear, should I apply a change here as well?

I'm not familiar with the semantics of `-j0` (I'm guessing that it means unlimited job slots?); I've always either used no argument or used a positive integer.

I have no strong opinion about this detail. `-j 8` reads very reasonably to me. `-j 0` feels a little more obscure, but I don't hate it.

getting-started/quick-reference.rst
willingc: Let's be consistent with the number. We could add a note that for most recent macs 8 would be fine.

getting-started/quick-reference.rst
willingc: Perhaps move this to line 124. Then, start a new paragraph for `blurb-it`.

index.rst
willingc: OK, let's do this. Let's remove and if folks get confused, we can always add this later. Thanks @sirosen.

willingc (approved):
Thanks @sirosen. I appreciate the effort.

getting-started/quick-reference.rst
sirosen: I'd like to keep a link to the detailed reference at the end of this subsection if we can manage it. Keeping the ordering of the content consistent section to section makes the doc more navigable.
(If we can't maintain this ordering, that's okay, but I want to try.)

I can definitely edit the perhaps-too-pithy "many changes deserve a NEWS entry" up above.
Here's what's written in the linked doc:

> Most changes made to the codebase deserve an entry in [Misc/NEWS.d](https://github.com/python/cpython/blob/main/Misc/NEWS.d), except for the following:
> 
> - documentation changes
> - test changes
>  - strictly internal changes with no user-visible effects
> - changes that already have a NEWS entry
> - reverts that have not yet been included in any formal release (including alpha and beta releases)

If I reproduce this whole list, the quick-reference begins to wander into being a not-so-quick-reference.

Would this update be satisfactory?
```diff
-Many changes deserve a NEWS entry which documents what changed.
+Most user-visible changes, other than documentation changes,
+deserve a NEWS entry which documents what changed.
```

sirosen: Concurrent comments!
I'm considering that Carol's suggestion maybe works for what I want, in that we maintain "Short Version, Deep Link, Short Version, Deep Link, ..." alternating structure.

So the possibility with that approach is something like...

> Many changes deserve a NEWS entry which documents what changed.
> For more information on how and when to write news entries,
> see :ref:`"Updating NEWS and What's New in Python" <news-entry>`.
>
> To add a NEWS entry, add a new file in the ``Misc/NEWS.d/`` directory.
> The news entry can be created by using
> `blurb-it <https://blurb-it.herokuapp.com/>`__,
> or the :pypi:`blurb` tool and its ``blurb add`` command.
> 
> .. tip::
> 
>    You can read more about ``blurb`` in its
>    `repository <https://github.com/python/blurb>`__.
>
> For more information about how to create news entries, see
> :ref:`"How to add a NEWS entry" <crossref-I-need-to-add>`.

index.rst
sirosen: FWIW, I did include a "breadcrumb" per your other suggestion. It's now way at the bottom of this diff, but it's short-and-sweet:

```
.. note::

   The quick reference documentation has been moved to serve as a cheat-sheet and overview
   in :ref:`Getting started <getting-started>`.

Go to :ref:`the new quick reference <quick-reference>`.
```

getting-started/quick-reference.rst
encukou: > @encukou, are you happy to do this now

No, this doesn't spark joy here. It don't want to recommend this setup. But I won't block it, if it's not mandatory.


From `pre-commit.ci` problem statement:

> Developers spend a fair chunk of time during their development flow on fixing relatively trivial problems in their code.

I see this differently. Developers spend a fair chunk of time *looking at code while their fingers are engaged*.
Auto-fixing style issues often gives code the superficial appearance of quality -- it *looks* like someone spent time polishing it, but without the deeper benefits.
If we want to avoid the tedium, let us simply *tolerate* the occasional lack of trailing comma or over-long line.

That said, I agree that there *are* automatic checks that are useful -- unused `import`, ReST single-backquotes, or security scans, but they seem to be in the minority -- and I don't think they're worth installing locally before you first build CPython.

getting-started/quick-reference.rst
encukou: Unlike `python -m test`, `make` does not understand `-j0`.

The Mac/BSD equivalent for `make -j$(nproc)` is `make -j $(sysctl -n hw.logicalcpu)`. Not something you'd want to type in, but perhaps people copy/paste from here?

willingc: Let's use @encukou's suggestion `make -j $(sysctl -n hw.logicalcpu)` or `make -j 8`. My M2 on Tahoe returns 8. I slightly prefer Petr's suggestion since it will use the resources available on the user's system.

getting-started/quick-reference.rst
willingc: I like this suggestion @sirosen. Thoughts @StanFromIreland?

willingc (approved):

getting-started/quick-reference.rst
StanFromIreland:
```suggestion
      ./configure --config-cache --with-pydebug && make -j$(sysctl -n hw.logicalcpu)
```
I don't have a Mac or Mac-knowledge to test, but I think using all available is better then picking numbers we'll have to update in the future.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+
+
+Run the tests
+=============
+
+.. tab:: Unix
+
+   .. code-block:: shell
+
+      ./python -m test -j3
```
StanFromIreland:
```suggestion
      ./python -m test -j0
```

getting-started/quick-reference.rst
StanFromIreland:
```suggestion
      ./python.exe -m test -j0
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+   .. note::
+      :ref:`Most <mac-python.exe>` macOS systems use
+      :file:`./python.exe` in order to avoid filename conflicts with
+      the ``Python`` directory.
+
+.. tab:: Windows
+
+   .. code-block:: dosbatch
+
+      .\python.bat -m test -j3
```
StanFromIreland:
```suggestion
      .\python.bat -m test -j0
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+.. tab:: macOS
+
+   .. code-block:: shell
+
+      ./python.exe -m test -j8
+
+   .. note::
+      :ref:`Most <mac-python.exe>` macOS systems use
+      :file:`./python.exe` in order to avoid filename conflicts with
+      the ``Python`` directory.
```
StanFromIreland, line 81:
```suggestion
```
Seeing as this is a quick reference, we can skip such things where it's "on macOS by default."

getting-started/quick-reference.rst
StanFromIreland: That seems good, but we can drop the `tip` (and move it to the other section if it's not already there), "quick references" should ideally be succinct :-)

Additionally, I'd reword the second paragraph a little, like so:

```rst
A news entry can be created locally with the :pypi:`blurb` tool
and its ``blurb add`` command or online after a pull request has
been opened with blurb-it <https://blurb-it.herokuapp.com/>__.
```

We've generally discouraged manual creation as it complicates things for new contributors. Users who want to do so can see the instructions in the [How to add a news entry](https://cpython-devguide--1838.org.readthedocs.build/getting-started/pull-request-lifecycle/#how-to-add-a-news-entry) section.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,177 @@
...
+   Agreement (CLA) as described in the :ref:`Licensing <cla>` section of
+   this guide.
+
+.. _creating Pull Requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
+
+Work on your pull request
+-------------------------
+
+Make sure the
+:ref:`continuous integration checks on your Pull Request are green <keeping-ci-green>` (successful).
```
StanFromIreland:
```suggestion
Make sure the :ref:`continuous integration checks on your Pull
Request are green <keeping-ci-green>` (successful).
```
(FYI you can wrap inside a `:ref:`)

getting-started/quick-reference.rst
willingc: Let's use the longer suggestion from Petr here too.

getting-started/quick-reference.rst
willingc: @StanFromIreland I would leave this note as is. macOS systems do not typically use `.exe` files so being explicit here reduces confusion.

getting-started/quick-reference.rst
willingc: I have similar feelings as @encukou about autofix (though I do have it enabled on CI in other projects. In general, I would have CI run the pre-commit run. Locally, I think it should be optional, not recommended (a subtle difference).

Folks, I think this is the only item that needs a decision after cleanup of a few commands where there is consensus. Let's try to converge so that we can merge. Thanks!

nedbat: Perhaps I've missed something, but it seems that the original page didn't mention pre-commit, and this is an addition. Since the goal of this PR is to move content, and this point is becoming difficult, shouldn't we let the status quo win and not add pre-commit now?

sirosen: Yes, this was something I added, in the course of walking through the original and cross-checking it against the full guide.

I hadn't expected it to be contentious, as [it's documented as "recommended" in the more complete docs](https://devguide.python.org/getting-started/setup-building/#install-pre-commit-as-a-git-hook). I would be okay with changing that to "optional" here.

That said, I prefer dropping it. Doing so makes the guide shorter, and this is meant to be a "quick reference". I'll plan to do that when I come back to apply (final? :crossed_fingers: ) changes, probably in a few hours time.

willingc: Let's drop it. We can always add it in a future PR.

getting-started/quick-reference.rst
sirosen: I'm going to apply Stan's suggestion, since this is an example of the `python -m test` usage, and we're therefore updating all of them to `-j0`.
But I'm also applying the suggestion above (Petr's idea) for the `make -j` invocation.

The end result is that for `make` on Linux and macOS, we'll have parallel versions using `nproc` and `sysctl`, but for `python -m test` we'll use `-j0` on all platforms.

sirosen:
I've just made what I think are the final necessary changes for this scope of work.

There's been a lot of feedback! As a result, it's possible I've missed something somewhere. Just let me know if there's more to do!

I intentionally skipped the suggestion to remove the macOS note admonition about `python.exe`. I'd like to leave that as-is at least for now.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
+.. _quick-reference:
+
+===============
+Quick reference
+===============
```
hugovk, line 5:
```suggestion
===============
Quick reference
===============

.. include:: /include/activate-tab.rst
```
To autoselect the tab based on your OS, like  https://devguide.python.org/getting-started/setup-building/#install-dependencies

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
...
+
+See also, GitHub's documentation on `commenting on Pull Requests`_.
+
+.. note::
+
+   In order to keep the commit history intact, avoid squashing or amending
+   history and then force-pushing to the PR.
+   Reviewers often want to look at individual commits.
+
+   CPython uses squash merges, so PRs will end up as single commits when merged.
```
hugovk: Let's keep this with a linkable header, so we can link to it after contributors force push.

See https://github.com/python/devguide/pull/1687.

```suggestion
Don't force-push
----------------

In order to keep the commit history intact, avoid squashing or amending
history and then force-pushing to the PR.
Reviewers often want to look at individual commits.

CPython uses squash merges, so PRs will end up as single commits when merged.
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Work on your pull request
+-------------------------
+
+Make sure the :ref:`continuous integration checks on your Pull
+Request are green <keeping-ci-green>` (successful).
+
+Read and respond to reviewer comments on your pull request.
+
+See also, GitHub's documentation on `commenting on Pull Requests`_.
```
hugovk:
```suggestion
See also, GitHub's documentation on `commenting on pull requests`_.
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
...
+   Agreement (CLA) as described in the :ref:`Licensing <cla>` section of
+   this guide.
+
+.. _creating Pull Requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
+
+Work on your pull request
+-------------------------
+
+Make sure the :ref:`continuous integration checks on your Pull
+Request are green <keeping-ci-green>` (successful).
```
hugovk:
```suggestion
Make sure the :ref:`continuous integration checks on your pull
request are green <keeping-ci-green>` (successful).
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Create pull requests on GitHub from your branches, on your fork, and make sure
+to put the relevant issue number in ``gh-NNNNNN``` format in the pull request title.
+For example:
+
+.. code-block:: text
+
+   gh-12345: Fix some bug in spam module
+
+See also, GitHub's documentation on `creating Pull Requests`_.
```
hugovk:
```suggestion
See also, GitHub's documentation on `creating pull requests`_.
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,168 @@
...
+
+For more detailed guidance, follow the :ref:`step-by-step pull request guide <pullrequest-steps>`.
+
+.. note::
+
+   First time contributors will need to sign the Contributor Licensing
+   Agreement (CLA) as described in the :ref:`Licensing <cla>` section of
+   this guide.
+
+.. _creating Pull Requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
```
hugovk:
```suggestion
.. _creating pull requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
```

hugovk (commented):
Shall we also add "Quick reference" to the table at  https://devguide.python.org ?

getting-started/quick-reference.rst
sirosen: I didn't know we had that. Cool!

getting-started/quick-reference.rst
sirosen: I wasn't sure if "Pull Request" is a proper noun. I'll apply all of these to lowercase it.

getting-started/quick-reference.rst
sirosen: 👍 I'll also make a small amendment to move up the ref below, so that it stays in the section where it is used.

sirosen:
> Shall we also add "Quick reference" to the table at https://devguide.python.org ?

Are you okay with us circling back on that after this PR? I think that table needs some more adjustment beyond this addition.

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,171 @@
...
+-------------------------
+
+Make sure the :ref:`continuous integration checks on your pull
+request are green <keeping-ci-green>` (successful).
+
+Read and respond to reviewer comments on your pull request.
+
+See also, GitHub's documentation on `commenting on pull requests`_.
+
+.. _commenting on Pull Requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request
```
hugovk:
```suggestion
.. _commenting on pull requests: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request
```

hugovk (approved):
> > Shall we also add "Quick reference" to the table at [devguide.python.org](https://devguide.python.org) ?
> 
> Are you okay with us circling back on that after this PR? I think that table needs some more adjustment beyond this addition.

Yes, absolutely. Thanks for this PR!

getting-started/pull-request-lifecycle.rst
```diff
@@ -290,13 +240,20 @@ in the "What's New in Python" entry.
...
 -----------------------
 
 ``NEWS`` entries go into the ``Misc/NEWS.d`` directory as individual files. The
 ``NEWS`` entry can be created by using `blurb-it <https://blurb-it.herokuapp.com/>`_,
 or the :pypi:`blurb` tool and its ``blurb add`` command.
 
+.. tip::
+
+   You can read more about ``blurb`` in its
+   `repository <https://github.com/python/blurb>`__.
```
StanFromIreland:
```suggestion
   :github:`repository <python/blurb>`.
```

getting-started/quick-reference.rst
```diff
@@ -0,0 +1,171 @@
...
+been opened with `blurb-it <https://blurb-it.herokuapp.com/>`__.
+
+For more information about how to create news entries, see
+:ref:`"How to add a NEWS entry" <news-entry-howto>`.
+
+Create pull requests
+--------------------
+
+Create pull requests on GitHub from your branches, on your fork, and make sure
+to put the relevant issue number in ``gh-NNNNNN``` format in the pull request title.
```
StanFromIreland:
```suggestion
to put the relevant issue number in ``gh-NNNNNN`` format in the pull request title.
```

getting-started/pull-request-lifecycle.rst
```diff
@@ -16,58 +16,8 @@ that you create a branch in Git, make your changes, push those changes
...
-   are green <keeping-ci-green>` (successful)
-
-#. Review and address `comments on your Pull Request`_
-
-#. When your changes are merged, you can :ref:`delete the PR branch
-   <deleting_branches>`
-
-#. Celebrate contributing to CPython! :)
-
-Don't force-push
```
StanFromIreland, line 56: Why drop this section? We have linked to it in *many* comments at this point, this will break a large amount of references.

StanFromIreland (commented):
Last few comments, in general looking good.

getting-started/pull-request-lifecycle.rst
sirosen: Hm. We have a copied, near-identical version of it in the new guide, but that won't help with links to this page...

What about reintroducing it below, before or after "Resolving merge conflicts"?

I've been trying to remove duplicate docs, but this seems like something that should stay on this page too.

sirosen: I went ahead and pushed this -- after my prior comment I started to feel confident that this is the right improvement.

willingc: Yes, we should definitely keep the "force-push" section. I see that you have made the change as requested. Thanks.

willingc:
OK, it looks as if all the outstanding items have been addressed. Any additional items can be handled in follow-up PRs. Thank you @sirosen for this helpful improvement. :sunny: Thanks to all of the reviewers for your constructive feedback and time. I going to merge this now.

PR #1840, opened 2026-06-18
Feature: Add tabs to build dependencies for Linux

### LINKED ISSUE #1790: Feature: Add tabs to build dependencies for Linux

https://devguide.python.org/getting-started/setup-building/#build-dependencies has parts for `dnf`-based distros and `apt`-based distros. Those could be put under appropriate tabs since they don't overlap. See the macOS tab where it's broken down to Homebrew and MacPorts as an example.

_No response_

### Additional context

_No response_

pranavchoudhary-tech:
Resolves #1790

Hi @brettcannon! 

As requested in the issue, I've reorganized the Linux build dependencies section to use separate `dnf` and `apt` tabs. This groups the dependencies neatly so users only see the commands relevant to their specific Linux distribution, matching the style of the macOS tabs below it.

Let me know if you need any adjustments!

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -768,64 +768,67 @@ some of CPython's modules (for example, ``zlib``).
    distribution, but the appropriate commands for some popular distributions
    are below.
 
-   On **Fedora**, **RHEL**, **CentOS** and other ``dnf``-based systems::
+   .. tab:: dnf
 
-      $ sudo dnf install git pkg-config
-      $ sudo dnf install dnf-plugins-core  # install this to use 'dnf builddep'
-      $ sudo dnf builddep python3
+      On **Fedora**, **RHEL**, **CentOS** and other ``dnf``-based systems::
 
-   Some optional development dependencies are not included in the above.
-   To install some additional dependencies for optional build and test components::
+         $ sudo dnf install git pkg-config
+         $ sudo dnf install dnf-plugins-core  # install this to use 'dnf builddep'
+         $ sudo dnf builddep python3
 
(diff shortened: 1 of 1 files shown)


brettcannon (approved):

getting-started/setup-building.rst
```diff
@@ -768,64 +768,67 @@ some of CPython's modules (for example, ``zlib``).
    distribution, but the appropriate commands for some popular distributions
    are below.
```
StanFromIreland, line 769: This needs to be updated now, we no longer show distros, we list package managers. Can you please open a PR for this?

encukou:
I'm a bit late here, but: do Ubuntu users know they need the `apt` tab? It doesn't look obvious to me.

brettcannon:
> I'm a bit late here, but: do Ubuntu users know they need the `apt` tab? It doesn't look obvious to me.

It says it under the tab, but trying to list every Debian or RH derivative never ends. Plus if you don't recognize `apt` there's going to be a bit more of a learning curve around a terminal before you try to compile Python from scratch.

It's also a 50/50 click if you have to look at both to figure it out.

encukou:
> trying to list every Debian or RH derivative never ends

Yeah, but Ubuntu is the well-known one and Debian is the "root"; I don't think we need more.
Same with Fedora/RHEL/CentOS -- I'd currently assume that users of any of the the derivatives would know this is their section them. (And CentOS is probably superfluous here.)

Anyway, merged in #1841 before I read this. Sorry for that!
Let's revisit when more tabs are added & we start running out of space?

PR #1841, opened 2026-06-18
Linux build dependencies: use distro names, not package managers

pranavchoudhary-tech:
Follow-up to PR #1840 to address the review comment from @StanFromIreland about changing 'distributions' to 'package managers' in the Linux tab introduction text.

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -765,7 +765,7 @@ some of CPython's modules (for example, ``zlib``).
    For Unix-based systems, we try to use system libraries whenever available.
    This means optional components will only build if the relevant system headers
    are available. The best way to obtain the appropriate headers will vary by
-   distribution, but the appropriate commands for some popular distributions
+   distribution, but the appropriate commands for some popular package managers
    are below.
 
    .. tab:: dnf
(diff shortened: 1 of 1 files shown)


getting-started/setup-building.rst
```diff
@@ -765,7 +765,7 @@ some of CPython's modules (for example, ``zlib``).
    For Unix-based systems, we try to use system libraries whenever available.
    This means optional components will only build if the relevant system headers
    are available. The best way to obtain the appropriate headers will vary by
-   distribution, but the appropriate commands for some popular distributions
+   distribution, but the appropriate commands for some popular package managers
```
StanFromIreland: I think we should reword it a little, we current awkwardly repeat "appropriate"

> The best way to obtain these headers varies by distribution, but commands for some popular package managers are given below.

StanFromIreland (approved):

encukou:
This doesn't look correct. Package names do depend on the distro, rather than the package manager software.

I guess it's OK to use `dnf` instead of “Fedora, RHEL, CentOS and so on” in the tab title, but to me it doesn't make sense in running text.

StanFromIreland:
I think longer tab title are fine, too. We can switch to distro names?

pranavchoudhary-tech:
Done! I've updated the tab titles to the distro names and changed the text back to refer to distributions.

encukou (approved):
Looks good!

I'll merge next week if there's no more discussion.

pranavchoudhary-tech:
@StanFromIreland @encukou Thank you both for your guidance and for merging this! I really enjoyed working on it. I'd love to keep contributing—are there any other specific issues or areas in the devguide that you'd recommend I tackle next?

PR #1844, opened 2026-06-26
Add a note to the style guide about abbreviations

### LINKED ISSUE #1824: Recommend avoiding `abbr` role

We should avoid using the [Sphinx `abbr` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-abbr) because it is inaccessible on mobile, screen readers, Braille display, and other touch devices.

If you want to use it, spell out in plain text.

For example, instead of:

```rst
:abbr:`BMP (Basic Multilingual Plane)`
```
Use:
```
Basic Multilingual Plane (BMP)
```

See https://github.com/python/cpython/pull/150673#pullrequestreview-4428977059 for more.

_No response_

### Additional context

_No response_

sirosen:
Hat tip to @hugovk for this! Not only the issue (#1824), but also [his comments preceding it](https://github.com/python/cpython/pull/150673#pullrequestreview-4428977059) made this easy to write.

This as a draft because it conflicts with #1828, which is already approved and should ideally merge first.
Once that's done, the exact positioning of this content in the page may change.

Open question: should this be converted to be a subsection of "Use simple language" or "Specific words"?

-----

resolves #1824

After the section on specific terms, but before "simple language", add a section that explains that documentation should spell out acronyms, preferring the `"<full-spelling> (<acronym>)"` format.


StanFromIreland:
Hi Stephen, this issue was originally specific to `:abbr:`, but I like your expanded approach. I would however suggest a little move.

- We have a [section for Sphinx roles](https://cpython-devguide--1844.org.readthedocs.build/documentation/markup/#roles), which is where I think the role-specific guidance should go.
- Spelling out acronyms can be moved to "Use simple language" (which, as I think about it now, would be better if it was named "Use accessible language," but that's for another day ;-).

sirosen:
Oh, I hadn't even thought to put it in the role usage docs! :facepalm: 
That seems obvious now.

I want to keep some explanation of the rationale. I'll move things as you suggest, and I think the "some assistive technology" note will end up with the roles.

---

Edit done! LMK if it needs more tuning.

sirosen:
> "Use simple language" (which, as I think about it now, would be better if it was named "Use accessible language," but that's for another day ;-).

Maybe... I like "Use simple language". It has two virtues:
1. It is a simpler word for the idea of "accessibility", more familiar to more readers -- to a large degree they are equivalent here
2. In the measure that the two words are not equivalent, "simple" is good general advice for writing well

I could probably be convinced, but we can discuss outside of this PR. :grin:

StanFromIreland:
As I said, that's for another day ;-) I also have to catch up on the Discord discussion.

Thanks for the changes, it's what I had in mind.

documentation/style-guide.rst
```diff
@@ -68,6 +68,10 @@ be native English speakers.
 Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
 such as "for example" or "that is."
 
+The first time an acronym is used on a page, spell it out.
+Prefer to write out the full term and follow it with the acronym in parentheses.
+For example, write "JavaScript Object Notation (JSON)".
```
hugovk: There are probably exceptions that we don't want to write out, for example, "HTML" can usually be left as is.

Do we need to mention stuff like this?

StanFromIreland: Maybe start the note with, "in general" to allow for exceptions.

sirosen: Are we able to describe the primary case for breaking the rule?
For example, what about adding this sentence to the end?

> When an acronym is very well known or expanding it could add confusion, as with "HTML" or "IP Address", exercise your judgement.

I wouldn't mind "In general, ..." if the above seems no good.

sirosen: I looked at it both ways and added the `In general, ...` phrasing. My version was too verbose.

hugovk (approved):

documentation/style-guide.rst
```diff
@@ -68,6 +68,10 @@ be native English speakers.
 Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
 such as "for example" or "that is."
 
+In general, the first time an acronym is used on a page, spell it out.
+Prefer to write out the full term and follow it with the acronym in parentheses.
+For example, write "JavaScript Object Notation (JSON)".
```
```suggestion
For example, write "Basic Multilingual Plane (BMP)".
```
I think 'JSON' is common enough that we don't always need to provide the full term. I suggest we use BMP instead (the abbreviation that led to the issue).

sirosen: I'm slightly torn on this because I think "BMP" is pretty obscure, but "JSON" is better known. To me, that gives "JSON" better explanatory power. But I'm okay with the suggestion -- we can always tweak it again later.

documentation/style-guide.rst
```diff
@@ -68,6 +68,10 @@ be native English speakers.
 Don't use Latin abbreviations like "e.g." or "i.e." where English words will do,
 such as "for example" or "that is."
 
+In general, the first time an acronym is used on a page, spell it out.
```
JelleZijlstra: Ironically nothing in the devguide defines the "HTML" abbreviation which you use in the other part of this PR. And I think that's a good thing: in computing there are some terms that are more widely understood as abbreviations than as the full form, and adding the full form is more likely to distract than to help readers. HTML is one of those terms. Other examples off the top of my head are UTF-8, PNG, SVG, SQL.

sirosen: Yeah, sometimes an acronym becomes a term or word in its own right.
The guidance should be read as flexible in this respect.

We added the "In general" to try to succinctly cover this. Do you feel it needs more explanation? We can try another arrangement if the current text looks too strong.

JelleZijlstra: The current text will likely lead people to comment that you need to say "Hyper-Text Markup Language". I'd suggest wording like "Commonly understood acronyms such as HTML or UTF-8 do not need to be expanded."

sirosen: I tried a few different "spellings" of the idea, and ended up with almost exactly this suggested text.

(Sorry, I broke the typical rules and did an amend, as I forgot to put Jelle in as a co-author at first!)

StanFromIreland: > as I forgot to put Jelle in as a co-author at first!)

In that case, in the future please comment for whoever is merging to add it in (we squash anyway), to avoid amends.

hugovk: Or add an empty commit with the co-author, or don't worry about it :)

documentation/markup.rst
StanFromIreland:
```suggestion
In the CPython documentation, there are a few common cases
```
We're now at three :-)

documentation/style-guide.rst
```diff
@@ -68,6 +68,11 @@ be native English speakers.
...
```
StanFromIreland: Can we also add "abbreviation" here please.

StanFromIreland (commented):
A couple final nits, otherwise looks great.

Additionally, can you fix these, please:

```
$ gg ":abbr:"
developer-workflow/development-cycle.rst:to the :abbr:`VCS (version control system)`.
developer-workflow/stdlib.rst:tracker and :abbr:`VCS (version control system)`.
developer-workflow/stdlib.rst:by following the :abbr:`PEP (Python Enhancement Proposal)` process.
development-tools/clinic/index.rst:The Argument Clinic :abbr:`CLI (Command-Line Interface)` is typically used to
getting-started/pull-request-lifecycle.rst::abbr:`VCS (version control system)` to be released
getting-started/setup-building.rst:     Contains the :abbr:`PEG (Parser Expression Grammar)` grammar file for
testing/coverage.rst:to the language and :abbr:`stdlib (standard library)` are accompanied by
```

With those, LGTM :-)

documentation/markup.rst
StanFromIreland: Hmm, that suggestion seems broken, for clarity: "couple" -> "few"

documentation/style-guide.rst
sirosen: I'll use an empty commit if it comes up again! That way, it's automatically included in the squash and nobody has to do anything extra.

I strive for perfection, including in my citations. Clearly I need to take the no-force-pushes rule into account more strongly! :sweat_smile:

documentation/markup.rst
sirosen: I think I'll do a normal commit with this + the `:abbr:` fixes + the fix below, so I'll try not to think about what that weird rendering in GitHub means.

StanFromIreland (approved):
Looks good, thanks.

PR #1847, opened 2026-06-29
Security Policy: Add note on referencing provisional CVE IDs

StanFromIreland:
This issue has come up a few times now, e.g. in [python/cpython#GHSA-8mmf-429p-8g58](https://github.com/python/cpython/security/advisories/GHSA-8mmf-429p-8g58).

### DIFF
--- a/security/policy.rst
+++ b/security/policy.rst
@@ -138,6 +138,9 @@ Here's what to expect for how a vulnerability report will be handled:
   may open a public issue.
 * If the PSRT determines the report is a vulnerability, the PSRT will
   accept the report and a CVE ID will be assigned by the PSF CNA.
+
+  Do not publicly reference the assigned CVE ID before its record is published,
+  as the report and ID remain provisional and may still be changed.
 * Once a public pull request containing a fix is merged to CPython,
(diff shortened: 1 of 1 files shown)


security/policy.rst
```diff
@@ -138,6 +138,8 @@ Here's what to expect for how a vulnerability report will be handled:
   may open a public issue.
 * If the PSRT determines the report is a vulnerability, the PSRT will
   accept the report and a CVE ID will be assigned by the PSF CNA.
+  Do not publicly reference the assigned CVE ID before its record is published,
```
hugovk, line 141: Could also be its own bullet point:

```suggestion
* Do not publicly reference the assigned CVE ID before its record is published,
```

hugovk (approved):

sethmlarson (approved):

security/policy.rst
StanFromIreland: I'll keep it together, the other bullet points are steps in the process, so this doesn’t quite fit.

StanFromIreland:
Thanks for the reviews!

PR #1849, opened 2026-06-30
Security policy: exceptions are not crashes

encukou:
This is mostly word nitpicking, but: by "crash" we generally mean a segfault or similar; see `Lib/test/crashers` for example.
Unhandled exceptions are not crashes, and are not a security concern.

However, since segfaults usually can be weaponized, a “crash resulting from unhandled exception” *would* be a PSRT issue.


sethmlarson (approved):

StanFromIreland (approved):
LGTM, however, can we keep it under availability vulnerabilities?

security/policy.rst
```diff
@@ -45,8 +45,10 @@ triggerable with data inputs that are reasonably sized for the use case.
 Availability vulnerabilities must also demonstrate an "upward" change in posture
 for the attacker, rather than a "lateral" one.
 This is to avoid handling performance improvements as security vulnerabilities.
-Exceptions are an expected part of control flow when processing inputs,
-therefore crashes resulting from unhandled exceptions are not security vulnerabilities.
+
```
encukou, line 48: @StanFromIreland do you mean keeping it in the previous paragraph?
```suggestion
```
We can do that, but I think it reads better when it's separate.

StanFromIreland: I'm +1 for keeping it together, it goes naturally with "Vulnerabilities that affect availability" but I'll leave it up to you.

Cyprius59 (approved):

Cyprius59 (approved):

Cyprius59 (approved):

PR #1868, opened 2026-07-22
Expand "various new privileges" in the new core team member checklist

encukou:
A note for myself, and future deputies:

image

I can't think of more than GitHub & Discord.
@ambv, do you happen to have a list?

### DIFF
--- a/core-team/join-team.rst
+++ b/core-team/join-team.rst
@@ -80,7 +80,8 @@ are granted through these steps:
         * - Docs Discord
           - "Core Team"
           - ---
-          - ``@encukou``
+          - | ``@encukou`` (Petr)
+            | ``@justbinary`` (Stan)
 
      (Some of these are unofficial spaces managed as a courtesy.
      Adjust the list to taste.)
--- a/core-team/join-team.rst
+++ b/core-team/join-team.rst
@@ -53,8 +53,23 @@ are granted through these steps:
 
 #. Once the candidate has provided the pertinent details, the approver will:
 
-   - Enable the various new privileges.
-   - Remove the new committer from the triage team, if applicable.
+   - Enable new permissions for the new committer, and remove redundant ones
+     if applicable:
+
+     .. list-table::
+        :header-rows: 1
+        :align: left
+
+        * - Service
+          - Add to
+          - Remove from
+        * - GitHub
+          - :gh-python-team:`python-core`
+          - :gh-python-team:`python-triage`
+        * - Discord
+          - "Core Developer"
+          - "Triager"
+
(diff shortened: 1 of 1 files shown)


core-team/join-team.rst
```diff
@@ -53,8 +53,23 @@ are granted through these steps:
...
+
+        * - Service
+          - Add to
+          - Remove from
+        * - GitHub
+          - :gh-python-team:`python-core`
+          - :gh-python-team:`python-triage`
+        * - Discord
+          - "Core Developer"
+          - "Triager"
```
StanFromIreland, line 78: DPO role?

encukou: Thanks for the catch!

core-team/join-team.rst
StanFromIreland: I think it's more than a "few" now, I suggest updating this.

encukou:
I also realized there's more places that would welcome a ping when a new core team member appears. Let's invite them.

core-team/join-team.rst
```diff
@@ -53,16 +53,45 @@ are granted through these steps:
...
+          - ---
+          - `admins <https://discuss.python.org/g/admins>`__
+        * - Core Discord
+          - "Core Developer"
+          - "Triager"
+          - ``@Server Admin``
+        * - Docs Discord
+          - "Core Team"
+          - ---
+          - ``@encukou``
```
StanFromIreland: You can also add me as a backup, if you'd like, I'm happy to take care of it. I just checked the last few team members, and I see Pieter is already added, but Chris and Tomas were missed (which I've since remediated :-).

core-team/join-team.rst
```diff
@@ -53,16 +53,45 @@ are granted through these steps:
...
    - Add their details to `🔒 python/voters <https://github.com/python/voters>`__.
    - Once the python/voters update is merged, regenerate the public team membership
      list at :ref:`developers`.
      See "Public list of members" in the ``voters`` README.
-   - Post an announcement in the `Committers Discourse category
-     <https://discuss.python.org/c/committers/5>`__.  The past few announcements
-     were in the form of a separate post on the already open topic with
-     the poll.
+   - Post an announcement to the Discourse topic with the promotional poll.
+     The topic should be in the `Promotion tag in the Committers category`_.
```
StanFromIreland, line 94: I don't think this is necessary here, since the original promotion post is already up?

encukou: Having a link here is very useful :)

StanFromIreland (approved):

encukou:
Thank you for the review!

PR #1874, opened 2026-08-08
Python Security Policy: Add note on sandboxes

StanFromIreland:
CC @python/psrt 

Prompted by a recent report (and several similar ones before it), this clarifies that sandboxing untrusted Python code is not a security boundary we support. Escapes from third-party sandboxes should be reported to their developers instead.

### DIFF
--- a/security/policy.rst
+++ b/security/policy.rst
@@ -33,6 +33,10 @@ and :func:`exec` are documented to execute arbitrary Python code that is
 supplied as data. The :mod:`ctypes` module is documented to enable modifying
 arbitrary locations in memory.
 
+CPython does not support sandboxing untrusted Python code as a security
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

gpshead (approved):

woodruffw (approved):

woodruffw:
(Speaking from PSRT perspective: seems reasonable to me, IMO it'd be fair to generalize even further to "Python does not make security representations about arbitrary code, any claims about the security of arbitrary code should be evaluated by the party making those claims.")

Eclips4 (approved):
FWIW, there have been were a few attempts to make Python sandboxable, and none of them were successful succesfull. As Victor Stinner once said, "run python in a sandbox, not the opposite!"

picnixz (approved):

PR #1882, opened 2026-08-15
Add a little section on the motivation behind translation

StanFromIreland:


willingc (approved):
I love this @StanFromIreland. Thoughtfully written and a lovely addition. I suggest you share in the core dispatch newsletter.

This would be great content to add to the PSF website as well @JacobCoffee.

documentation/translations/translating.rst
```diff
@@ -169,6 +167,39 @@ Roughly, translation works as follows:
   docs.python.org.
 
 
+Motivation
+==========
+
+The Python documentation is written in English, but Python is taught, learned,
+and used around the world, in many languages. English is the native language
+of only about 5% of the world's population, and fewer than a fifth more speak
+it as an additional language [#f1]_. For many people, learning Python also
```
hugovk, line 176: Rather than bare footnotes, let's use direct links that save our readers an extra click.

StanFromIreland: I wrote https://github.com/python/devguide/pull/1883, but I have some concerns about this approach.

m-aciek (commented):
In my opinion motivation section could/should go without a section heading first in Translating chapter, Overview could/should go as the second, and current intro text as third with _Project status_ section heading.

This would be more natural and engaging to read for an average devguide reader.

I like the added text very much!

PR #1883, opened 2026-08-16
Replace footnotes with direct links

StanFromIreland:
This was suggested by Hugo in https://github.com/python/devguide/pull/1882#discussion_r3790450136. However, I'm not convinced. While I agree it saves a click for those who want to access the reference, I think that it distracts those who don't.

### DIFF
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -173,16 +173,18 @@ Motivation
 ==========
 
 The Python documentation is written in English, but Python is taught, learned,
-and used around the world, in many languages. English is the native language
-of only about 5% of the world's population, and fewer than a fifth more speak
-it as an additional language [#f1]_. For many people, learning Python also
-means overcoming a language barrier. They must learn not only Python, but also
-enough English to be able to use the language's documentation.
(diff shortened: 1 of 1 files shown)


documentation/translations/translating.rst
```diff
@@ -173,16 +173,18 @@ Motivation
...
-it as an additional language [#f1]_. For many people, learning Python also
-means overcoming a language barrier. They must learn not only Python, but also
-enough English to be able to use the language's documentation.
-
-The United Nations Open Source Principles call on projects both to provide
-documentation and to foster inclusive participation [#f2]_. Providing
-documentation only in English limits who can benefit from it and who can
-participate in the wider Python community.
+and used around the world, in many languages. English is the `native language
+of only about 5% of the world's population
```
StanFromIreland, line 177: This is the one I'm particularly concerned about, if we don't make the entire sentence a link it doesn’t quite make sense. But, that makes it quite distracting. WDYT @hugovk ?

hugovk: We could link "English", or "native language", or "only about 5%".

I suggest "native language", as that is what the page is about.

PR #1885, opened 2026-08-17
Document deb822 sources for Ubuntu 24.04 and later

### LINKED ISSUE #1530: Bug: Dependency installation on Ubuntu 24.04
### Describe the bug

There has been a change in how sources for apt are configured in Ubuntu 24.04. We add directions to cover this case in this doc: https://devguide.python.org/getting-started/setup-building/#install-dependencies

### Screenshots

_No response_

### Additional context

_No response_

opensource-joe:
Closes #1530.

## The problem, reproduced

The Debian/Ubuntu tab tells you to enable source packages by editing `/etc/apt/sources.list`, either by adding a `deb-src` line or by uncommenting the ones already there.

On Ubuntu 24.04 that file is a four line stub, and its own comments explain why:

```
# Ubuntu sources have moved to the /etc/apt/sources.list.d/ubuntu.sources
# file, which uses the deb822 format. Use deb822-formatted .sources files
# to manage package sources in the /etc/apt/sources.list.d/ directory.
```

There is nothing to uncomment. A contributor who follows the guide as written reaches the next step and gets:

```
$ sudo apt-get build-dep python3
Reading package lists...
E: You must put some 'deb-src' URIs in your sources.list
```

Which is an unhelpful place to land on your first CPython build.

## Verification

On **Ubuntu 24.04.4 LTS**, tested in both directions:

| | |
|---|---|
| `/etc/apt/sources.list` as shipped | 4 lines, all comments, **zero** `deb-src` lines to uncomment |
| `apt-get build-dep -s python3` before the change | `E: You must put some 'deb-src' URIs in your sources.list` |
| after adding `deb-src` to `Types` in `ubuntu.sources` and `apt-get update` | resolves the full dependency set, through to `python3-sphinx` |
| host restored afterwards | `diff` against the backup is identical, and `build-dep` fails again as before |

Ubuntu's own generated `ubuntu.sources` documents the fix at line 21:

```
## Types: Append deb-src to enable the fetching of source package.
```

## What changed

The 22.04 one-line instructions are **kept, not replaced**, since releases still using that format need them. The section now branches on which format the release uses, and names the file each one actually lives in.

I have not built the docs locally to check the rendering of the nested literal blocks inside the tab, so that is worth a glance in review.

### DIFF
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -794,7 +794,6 @@ some of CPython's modules (for example, ``zlib``).
 
       On **Ubuntu 24.04 and later**, and on other releases using the deb822
       format, the sources are in ``/etc/apt/sources.list.d/ubuntu.sources``.
-      ``/etc/apt/sources.list`` is left as a stub with nothing to uncomment.
       Add ``deb-src`` to the ``Types`` field of the entries you want sources
       for::
 
--- a/getting-started/setup-building.rst
+++ b/getting-started/setup-building.rst
@@ -790,13 +790,32 @@ some of CPython's modules (for example, ``zlib``).
       dependencies for the Python you're working on by using the ``apt`` command.
 
       First, make sure you have enabled the source packages in the sources list.
-      You can do this by adding the location of the source packages, including
-      URL, distribution name and component name, to ``/etc/apt/sources.list``.
-      Take Ubuntu 22.04 LTS (Jammy Jellyfish) for example::
+      Where those live depends on your release.
+
+      On **Ubuntu 24.04 and later**, and on other releases using the deb822
+      format, the sources are in ``/etc/apt/sources.list.d/ubuntu.sources``.
+      ``/etc/apt/sources.list`` is left as a stub with nothing to uncomment.
+      Add ``deb-src`` to the ``Types`` field of the entries you want sources
+      for::
+
+         $ sudo nano /etc/apt/sources.list.d/ubuntu.sources
+
+      changing::
+
+         Types: deb
+
+      to::
+
+         Types: deb deb-src
+
+      On **Ubuntu 22.04 and other releases using the one-line format**, add the
+      location of the source packages, including URL, distribution name and
+      component name, to ``/etc/apt/sources.list``. Taking Ubuntu 22.04 LTS
+      (Jammy Jellyfish) as the example::
 
          $ deb-src http://archive.ubuntu.com/ubuntu/ jammy main
 
-      Alternatively, uncomment lines with ``deb-src`` using an editor, for
+      Alternatively, uncomment the lines with ``deb-src`` using an editor, for
       example::
 
          $ sudo nano /etc/apt/sources.list


StanFromIreland:
Why are you closing and re-opening? You also have to sign the CLA before we can review.

opensource-joe:
Asking for a hand with the `CLA Signing` check, because I have run out of things to test from my side.

**I have signed the CLA.** `cla.python.org` shows a completed signature, not an unsigned form. The check has stayed red since.

Everything I can verify myself lines up, so there is no address mismatch to correct:

| | |
|---|---|
| Address the bot is asking for | `joecastleengineering@gmail.com` |
| Commit **author** on `a3c27a1` | `joecastleengineering@gmail.com` |
| Commit **committer** on `a3c27a1` | `joecastleengineering@gmail.com` |
| My GitHub **primary** email | `joecastleengineering@gmail.com` |

Both identity fields on the commit also resolve to this account rather than sitting unlinked.

The bot has re-evaluated four times today and returned `failure` each time, at `05:30:08Z`, `05:40:05Z`, `05:42:10Z` and `16:36:36Z`. The last of those is roughly eleven hours after signing, which I think rules out the CLA records simply not having propagated yet. These were genuine re-runs rather than a cached result: each one re-stamped every status on the commit, and `pre-commit.ci` and Read the Docs visibly returned to `pending` and re-ran alongside it. The bot's own comment has never changed from its original text.

**One thing I should own up to:** those re-evaluations came from me closing and reopening this PR, since `python-cla-bot` does not appear to re-check on its own and a signature recorded after the PR was opened will not move the status without a new PR event. That is what the close and reopen pairs in the timeline are. Apologies for the noise; I have stopped, and I will not do it again on this PR.

So my question is whether there is something further on my end that I have missed, or whether this needs someone who can see the CLA records. Happy to try anything suggested. I am equally happy for this to sit until someone has time, since it is a small documentation change and nothing is urgent about it.

For what it is worth the content side is green: `pre-commit.ci` passes and the Read the Docs build succeeds, with the rendered result at the preview link above.

hugovk:
I expect this is due to the recent GitHub outage, which had apparently just been resolved:

https://www.githubstatus.com/incidents/zkxwbgr0cnmx

So let's try again.

@opensource-joe Please also read the AI policy:

https://devguide.python.org/getting-started/ai-tools/

Specifically:

> We expect PR authors and those filing issues to be able to explain their proposed changes in their own words.

Which also applies to PR and issue comments, so please avoid pasting AI output directly.

opensource-joe:
@StanFromIreland and @hugovk, I went ahead and signed the CLA, possibly for a second time, although it only showed as one time for joecastleengineering@gmail.com, so it may have been under a different email. My other email, josephcastle@gmail.com, is tied to this account, and jcastle-zz, which is a closed GH account (long story while I was a GH Star). Apologies for opening and closing the PR so many times; here is what I think happened: On the closing and reopening, a prior session retriggered the check four times, since the bot doesn't re-poll on its own. It has stopped and will not happen again. I read up on the AI policy (https://devguide.python.org/getting-started/ai-tools/) and appreciate the link. I am happy to keep contributing, and if you have specific suggestions, issues, PRs, etc., and want to tag me, please do. Cheers!

zware:
Note that the CLA bot *does* re-poll on its own, as evidenced by the fact that it edited its comment 1 hour ago after the CLA was signed (again?) without any further prompting.  The closing and reopening was never necessary.

getting-started/setup-building.rst
```diff
@@ -790,13 +790,32 @@ some of CPython's modules (for example, ``zlib``).
...
 
       First, make sure you have enabled the source packages in the sources list.
-      You can do this by adding the location of the source packages, including
-      URL, distribution name and component name, to ``/etc/apt/sources.list``.
-      Take Ubuntu 22.04 LTS (Jammy Jellyfish) for example::
+      Where those live depends on your release.
+
+      On **Ubuntu 24.04 and later**, and on other releases using the deb822
+      format, the sources are in ``/etc/apt/sources.list.d/ubuntu.sources``.
+      ``/etc/apt/sources.list`` is left as a stub with nothing to uncomment.
```
encukou: There's no need to reference the historical location here.
```suggestion
```
Otherwise this looks good!

PR #1886, opened 2026-08-19
Argument Clinic: document deletion of an attribute and @deleter

serhiy-storchaka:
The setter generated by Argument Clinic is called with `NULL` to delete the attribute, and by default it rejects the deletion with an `AttributeError`. The `@deleter` directive, applied to the `@setter`, lets the implementation handle the deletion. Neither was documented.

The example of the generated code is also updated -- it was written before the accessors started to take `PyObject *self`.

### DIFF
--- a/development-tools/clinic/howto.rst
+++ b/development-tools/clinic/howto.rst
@@ -1419,24 +1419,32 @@ The generated glue code looks like this:
 .. code-block:: c
 
     static PyObject *
-    _io_TextIOWrapper__CHUNK_SIZE_get(textio *self, void *Py_UNUSED(context))
+    _io_TextIOWrapper__CHUNK_SIZE_get(PyObject *self, void *Py_UNUSED(context))
     {
         PyObject *return_value = NULL;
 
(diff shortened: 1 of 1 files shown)


development-tools/clinic/howto.rst
```diff
@@ -1446,6 +1454,45 @@ The generated glue code looks like this:
...
+   declaring the same accessor twice is an error.
+
+The setter slot of :c:type:`PyGetSetDef` is used both for setting and for
+deleting the attribute: the setter is called with ``NULL`` as the value to
+delete it.
+As shown above, the generated setter rejects the deletion with an
+:exc:`AttributeError` before calling the "impl" function.
+
+If the attribute can be deleted, apply the ``@deleter`` directive to the
+``@setter``.
```
encukou: Consider hinting that order is important (``@deleter`` after ``@setter`` doesn't work):
```suggestion
If the attribute can be deleted, add the ``@deleter`` directive after
``@setter``.
```

JelleZijlstra (approved):

PR #1887, opened 2026-08-20
Argument Clinic: document converters for getters and setters

serhiy-storchaka:
Documents the changes in python/cpython#156066, and should be merged after it.

The value of a `@setter` can be declared as a parameter with a converter, and a `@getter` can define a return converter.
The accessors of an attribute are now identified by its Python name, so they can use different C basenames and can be defined several times in mutually exclusive preprocessor conditional blocks.

`_ssl._SSLContext.check_hostname` is used as the main example instead of `_io.TextIOWrapper._CHUNK_SIZE`, because it uses converters, and `function.__type_params__` is used as the example of `@deleter`.

### DIFF
--- a/development-tools/clinic/howto.rst
+++ b/development-tools/clinic/howto.rst
@@ -1397,7 +1397,7 @@ that facilitate :py:class:`property`-like access for a class.
 You can use the ``@getter`` and ``@setter`` directives to generate
 "impl" functions using Argument Clinic.
 
-This example --- taken from :cpy-file:`Modules/_io/textio.c` ---
+This example --- taken from :cpy-file:`Modules/_ssl.c` ---
 shows the use of ``@getter`` and ``@setter`` in combination with
 the :ref:`@critical_section <clinic-howto-critical-sections>` directive
 (which achieves thread safety without causing deadlocks between threads)::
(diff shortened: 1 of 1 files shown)


development-tools/clinic/howto.rst
```diff
@@ -1465,50 +1508,28 @@ As shown above, the generated setter rejects the deletion with an
...
-   >>> import sys, _io
-   >>> a = _io.TextIOWrapper(sys.stdout)
-   >>> a._CHUNK_SIZE
-   8192
-   >>> a._CHUNK_SIZE = 30
-   >>> a._CHUNK_SIZE
-   30
+The value is only converted if the attribute is not deleted, so it must have
+a default value, which the "impl" function receives for the deletion.
+The implicitly declared value gets the default ``NULL``.
```
encukou: IMO, this is too confusing. The connection between *default* and *placeholder for `del`* is not obvious.
An `if (value=NULL)` inside the function covers all the use cases and is much clearer. Can we disallow arbirtary defaults (or enforce using `NULL`) here?

serhiy-storchaka: If we use converter such as `int`, we cannot use `if (value=NULL)` inside the function. For now, there are no such examples in the CPython code, but we can add them in future.

If deletion is equivalent to setting to None or other value, we can use it as the default value and remove a special case from function. For now there are no examples which do not store NULL internally.

So practically all current examples use NULL as default value. But adding additional restriction on the default value will only complicate the code and unnecessary restrict the future use.

PR #1893, opened 2026-08-26
linklint is now sphinx_linklint

nedbat:


documentation/style-guide.rst
```diff
@@ -237,7 +237,7 @@ the second mention is more appropriate for a link.  Some units are long enough
 to have a few repeated links.  Use judgement to decide when a link will help
 the reader.
 
-We use the :pypi:`linklint Sphinx extension <linklint>` to suppress two kinds
+We use the :pypi:`linklint Sphinx extension <sphinx_linklint>` to suppress two kinds
```
```suggestion
We use the :pypi:`linklint Sphinx extension <sphinx-linklint>` to suppress two kinds
```

StanFromIreland (approved):
One little nit, otherwise LGTM.

documentation/style-guide.rst
nedbat: pypi finds it either way, but I've switched it to the canonical name

PR #1894, opened 2026-08-26
Reorder initial sections on Translating page

m-aciek:
* _Motivation_ section goes without a section heading first in Translating chapter,
* _Overview_ goes as the second,
* and current intro text as third with _Project status_ section heading.

In my opinion this is more natural and engaging to read for an average devguide reader.

### DIFF
--- a/documentation/translations/translating.rst
+++ b/documentation/translations/translating.rst
@@ -6,6 +6,73 @@ Translating
 
 .. highlight::  rest
 
+The Python documentation is written in English, but Python is taught, learned,
+and used around the world, in many languages. English is the native language
+of only about 5% of the world's population, and fewer than a fifth more speak
+it as an additional language [#f1]_. For many people, learning Python also
+means overcoming a language barrier. They must learn not only Python, but also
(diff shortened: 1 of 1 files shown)


StanFromIreland:
I'll have to think about this a little. Having the motivation separate was useful as it was clear what it was, and it could be linked to. Now it looks a bit like an introduction. We also hide the table, which for most people looking to translate, us probably the most important thing on the page (although I admit it was a little awkwardly placed).

m-aciek:
This table used to be a place to look for information about in-progress translations. Now that we have a dashboard, well discoverable, under own subdomain, it took (or should took) over this role. The names of coordinators and links to decisions are not critical, are more of a metadata.

* Dashboard: [10k visitors](https://analytics.python.org/translations.python.org?f=is,page,/&period=91d) in last 90 days.
* This page: [850 visitors](
https://analytics.python.org/devguide.python.org?f=is,page,/documentation/translations/translating/&period=91d&keybindHint=N) in the same period

Issue #1895, opened 2026-08-30
Devguide recommends a very slow testing configuration


ncoghlan:
### Describe the enhancement or feature you would like

At the PyCon AU sprints, I found that running the test suite in parallel with `make test` under `--with-pydebug` on my machine took ~3.5 minutes, while running with the default configuration took ~1 minute.

Running the test suite sequentially with `python -m test` took ~18 minutes on a default build, while `--with-pydebug` took ~27 minutes.

One of the sprint attendees (with a much slower system in general) had their sequential test run take ~45 minutes under `--with-pydebug`, dropping to around ~25 minutes when using the default build configuration, and then down to ~5 minutes by running with `make test` rather than `python -m test`.

https://github.com/python/devguide/issues/1171 covered updating https://devguide.python.org/getting-started/quick-reference/#quick-reference to recommend parallel test execution, but https://devguide.python.org/testing/run-write-tests/#running buries the parallel execution command well down the page.

There are multiple intersecting issues here:

* between the intro, the table of section references, and the video embed, the "Quick reference" link has dropped "below the fold" on the landing page (at least on my laptop screen)
* some of the instructions were originally written for C-focused development and then never revisited for pure Python contributors
* we're inconsistent in our recommendations for how to run the test suite (and we don't emphasise how slow it is to run them sequentially, so you almost never want to actually do that if you can avoid it)

One change I think would be worth making is to recommend an unqualified `make` as the default build configuration (especially when working pure Python changes without touching any of the C code), and mention `make --with-pydebug` as an option that is slower in execution, but provides a much better debugging experience when working on C level changes (with reference to https://docs.python.org/3/using/configure.html#python-debug-build)

Another would be to move the quick reference section heading and linking paragraph up above the section link table.

hugovk:
> [#1171](https://github.com/python/devguide/issues/1171) covered updating [devguide.python.org/getting-started/quick-reference#quick-reference](https://devguide.python.org/getting-started/quick-reference/#quick-reference) to recommend parallel test execution, but [devguide.python.org/testing/run-write-tests#running](https://devguide.python.org/testing/run-write-tests/#running) buries the parallel execution command well down the page.

Most people have multicore, so parallel testing is a good default.

A couple of ideas:

* Add `-j0` to all the examples, except change the current "If you have a multi-core or multi-CPU machine, you can enable parallel testing using several Python processes so as to speed up things:" into how to run single process instead (`--single-process`).

[...]

PR #1896, opened 2026-08-30
GH-1895: Move quick reference link higher on landing page

ncoghlan:
Partially addresses GH-1895


StanFromIreland:
This partially reverts some of the work that @sirosen and I did when restructuring the landing page. I think we had a slightly different goal in mind at the time. We wanted to make the landing page friendlier and for it to provide a more complete introduction, rather than having people jump straight into the quick reference (which was originally at the very top of the page).

For contributors at sprints, jumping straight into the code may work well because they generally have a mentor around to help straighten out any issues. But for people trying to contribute on their own at home, I worry that the immediate jump into code made getting started seem quite difficult (I also got some feedback along these lines from some new contributors). I think it would be good to preserve some of that gentler introduction here.

ncoghlan:
Ah, that makes sense - I did wonder about the origin of the current sequencing.

I've put this back to draft for now - I'd still like to move the quick reference link before the big table, but it doesn't need to be the very first thing in the section (and it definitely isn't an urgent change).

PR #1905, opened 2026-09-12
Add guidelines for bulk reporting of issues

StanFromIreland:
This came up recently when a user opened a large number of issues (>45) in an hour, as the result of LLM analysis. This wasn't ideal, and we asked them to report as a single issue, in the same style @devdanzin has been doing (and has been working well for us). As such, I propose we document this so that we can link to it if this happens again.

### DIFF
--- a/security/policy.rst
+++ b/security/policy.rst
@@ -1,3 +1,5 @@
+.. _security-policy:
+
 ===============
 Security policy
 ===============
--- a/triage/issue-tracker.rst
+++ b/triage/issue-tracker.rst
@@ -29,6 +29,8 @@ If you would like to file an issue about this devguide, please do so in the
 :github:`devguide repository <python/devguide>` instead.
 
 
+.. _checking-if-a-bug-already-exists:
+
 Checking if a bug already exists
 --------------------------------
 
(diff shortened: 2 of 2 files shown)


hugovk:
I'm not sure if we should be encouraging any form of bulk reporting, even via single issues. 

It's still *a lot of work* for us to churn through, bulk findings are likely found via tools and not from actual real-life use, and we probably don't really need to fix them all?

I agree @devdanzin has been finding useful stuff, although he got in touch first to ask how (or whether) to report. But I think this is an exception rather than something to encourage as policy.

ZeroIntensity:
I think an issue-per-bug is fine as long as it's not all at once. Daniel's stuff has been posted to the issue tracker over multiple years, so it hasn't bothered anybody.

That said, I'm not sure we need to be encouraging this at all. It's good to have one or two trusted people doing some fuzzing, but crashes/bugs found by fuzzers should typically be very low priority. I've expressed this concern to other contributors before; CPython has a *lot* of existing open issues affecting real applications, and it's not a good use of our time to entertain LLMs making single-line fixes to artificial bugs across the codebase. (I think we have far too many people doing this already -- try and find a single `type-crash` issue from the past 6 months without some AI-generated PR attached!)

StanFromIreland:
I didn't intend this as encouragement, but rather as an acknowledgement of the current situation. People _are_ fuzzing CPython and reporting large numbers of issues as a result, whether we like it or not. We've seen this several times recently, both on the issue tracker and in security reports.

My motivation was therefore primarily to make sure that, when people do send us all of their findings, they do so in the way we prefer.

However, if this is being interpreted as encouragement, I'll withdraw this PR.

PR #1908, opened 2026-09-17
Rename builtin functions article name

m-aciek:
Update after Python docs, builtins has been moved to a dedicated chapter.


hugovk (approved):
Thanks, this is following the move in https://github.com/python/cpython/pull/156682.

PR #1909, opened 2026-09-18
PPC: add the packaging-council repository

henryiii:
This is the repository for the PPC (Python Packaging Council) we are preparing.

Currently in draft until this repo has something in it and is made public.

Should the typing-council repo be mentioned here too?


JelleZijlstra:
This is a "for example" list, I don't think it needs to include all repos.

hugovk:
I agree it doesn't need adding to the example list.

Let's instead add it to this governance table:

https://devguide.python.org/developer-workflow/development-cycle/#governance

And see the policy for adding new repos to the GitHub org:

https://devguide.python.org/developer-workflow/development-cycle/#organization-repository-policy

pradyunsg:
> And see the policy for adding new repos to the GitHub org:
>
> https://devguide.python.org/developer-workflow/development-cycle/#organization-repository-policy

I'm reading this as an ask to follow the process quoted below.

> Before adding a new repository to the organization, open a discussion to seek consensus in the Committers Discourse category. Once people are satisfied with that, ask the Python steering council to grant permission.

I think we'd be happy to (it'd be a bit surprising if anyone says no!) but just want to confirm that I'm not misreading. 😅

hugovk:
Yes, I suggest to follow the process, and expect it to be a quick yes. 

I think the SC came up with this policy because:

> Note that several repositories remain in the organization for historic reasons, and would probably not be appropriate to add today.

And we'd like to avoid repeating that. So let's follow it for the (hopefully!) clear cases. (Also it gives visibility to the process, which has been skipped recently in less clear cases.)

pradyunsg:
https://discuss.python.org/t/adding-a-python-packaging-council-and-possibly-moving-packaging-python-org/109135

developer-workflow/development-cycle.rst
```diff
@@ -428,6 +428,9 @@ and the repository where you can open an issue to ask for a decision.
    * - Typing Council
      - :pep:`729`
      - :github:`python/typing-council`
+   * - Packaging Council
```
hugovk, line 428: The list is alphabetical (after SC) not numerical or chronological, please can you move it between DEB + TC?

henryiii: Sure, done.

PR #1911, opened 2026-09-21
Update PSRT inactivity policy

StanFromIreland:
Changes based on the recent PSRT decisions that:
- Member activity is reviewed biannually.
- Former RMs and PSRT admins are included in the inactivity check.

CC @ambv please note that I've removed your RM manager note, in concordance with https://github.com/python/peps/commit/03bdeb3d912818163ce27e1b5f776d1af1de60d4.


security/psrt.rst
```diff
@@ -32,14 +32,17 @@ The nomination succeeds if the nomination receives at least
...
+Every six months the Steering Council will receive a report of inactive PSRT
+members with the recommendation to remove the inactive users from the PSRT.
 “Inactive” is defined as a member who hasn’t coordinated or commented on a
-vulnerability report in the past year since the last report was generated.
+vulnerability report in the past six months since the last report was generated.
 The Steering Council may remove members of the PSRT with a simple vote.
 
 Members of the PSRT who are a Release Manager or Steering Council member may
 remain in the PSRT regardless of inactivity in vulnerability reports.
+When a Release Manager's term ends, their PSRT membership becomes a regular
```
warsaw: Do we need to define the end of an RM's term, or is it obvious?  I'd define it as after the official EOL of the last release they are an RM for.

warsaw (approved):
Other than one question, LGTM!

security/psrt.rst
hugovk: I think it's obvious, but we can be explicit.

That definition makes sense: they no longer need to make any security releases for their versions, so there's no need for them to be present unless they want to be active.

---

Aside: The "may" in "Members of the PSRT who are a Release Manager or Steering Council member may remain in the PSRT regardless of inactivity in vulnerability reports" suggests membership is optional for RMs (and SC). I was told I had to join, and [membership is now required by PEP 101](https://peps.python.org/pep-0101/#things-you-ll-need).

hugovk (approved):

security/psrt.rst
StanFromIreland: > I'd define it as after the official EOL of the last release they are an RM for.

This however, is technically not what we've been doing. We list Steve and Ned as RMs, however as per [PEP 101](https://peps.python.org/pep-0101/#how-to-make-a-release) they're the {W,M}Es. To continue this practice, I suggest something like "as specified in PEP 101" since we're including the slightly broader release team?

warsaw: Yes, RM members should be mandatory, and optional for SC members.  I'd list release experts (WE, ME) as optional but allowed even if they aren't active.  But yeah, we can cross reference PEP 101 if you think that's better.

sethmlarson (approved):
Thanks Stan for the PR and all for the reviews.

security/psrt.rst
StanFromIreland: I've pushed a clarification.

